PR #4, opened 2021-04-06
Set up an initial actions file

greatbacon:
Adds a `build` file for github actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE
           npm i
-          npm i -g @feltcoop/gro
+          npm i @feltcoop/gro
 
       - name: Build the project
         run: gro build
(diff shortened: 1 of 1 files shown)


ryanatkn (approved):
Looks good!

ryanatkn:
We also have `gro check` that is an alias for:

- `npm test`
- typechecking with TypeScript
- format checking with Prettier
- codegen is fresh (if the project uses this feature)

greatbacon:
> We also have `gro check` that is an alias for:
> 
> * `npm test`
> * typechecking with TypeScript
> * format checking with Prettier
> * codegen is fresh (if the project uses this feature)
> 
> example: https://github.com/feltcoop/gro/blob/main/.github/workflows/nodejs.yml#L29

Good to know! I ran into a snag trying to install gro, so I'm gonna do stuff with that in a follow up PR

ryanatkn:
Was just thinking of that, see this example: 

https://github.com/feltcoop/felt-mockup/blob/main/.github/workflows/nodejs.yml#L44

ryanatkn:
There's also `npm ci` but I'm ignorant here. I've looked at other projects and I see both `npm i` and `npm ci` used, so maybe it doesn't matter? But the docs seem to indicate it'll be faster. https://docs.npmjs.com/cli/v7/commands/npm-ci

Issue #6, opened 2021-04-07
Use gro for a deploy system


greatbacon:
In order to get the shortest path to production deployments, we want to just use a gro command line (`gro deploy`) to manage deployments to our single instance. More complicated infra will come as we begin to expand.

ryanatkn:
For reference here's Gro's internal `src/deploy.task.ts` for static site deployment using git worktrees and pushing to a "deploy" branch for GitHub pages. It may have some rough edges; I'm about to make a PR using it with Felt, which is a static site project using `@sveltejs/adapter-static`, unlike the server deployment which will be like some variation of `@sveltejs/adapter-node`. (we currently have a hacky implementation, which is fine because this part of SvelteKit isn't yet stable)

PR #9, opened 2021-04-12
Add deploy command

greatbacon:
This PR is for issue [#6]

This PR adds a first pass at a simple deploy command that targets a single semi-private Linode instance.

In order to deploy you will need an ssh key added to the account (& the instance)

This work includes a hack to bundle the package.json files, & currently does not actually start the server (will change that once `gro` is updated)

I think at the end of all this it still won't work, because then I'll have to set us up a DB lol


.gitignore
```diff
@@ -10,3 +10,4 @@ node_modules
 *.ignore.*
 ignore/
 *.pem
+deploy.tar
```
ryanatkn, line 13: How about we change this to `/deploy` and put it after the `/dist` above? Gro needed something like this for its static deployment. Instead I put in a temporary hack that hides what it's doing, but that goes against its goals of transparency; the `/deploy` directory is the perfect place for this stuff.

src/deploy.task.ts
```diff
@@ -0,0 +1,27 @@
+import type {Task} from '@feltcoop/gro/dist/task/task.js';
+import {spawnProcess} from '@feltcoop/gro/dist/utils/process.js';
+import {DIST_DIR_NAME, paths} from '@feltcoop/gro/dist/paths.js';
+
+
+export const task: Task = {
+    description: 'deploy felt server to prod',
+    run: async ({invokeTask}) => {        
+        //TODO - add more dynamic naming
+        await spawnProcess('tar',['-cvf','deploy.tar', DIST_DIR_NAME, 'package.json', 'package-lock.json']);
```
ryanatkn, line 10: I see this isn't formatted and will cause CI to fail - run `gro format`. I'm wondering if we should add a `.vscode` directory to each project and include the `formatOnSave` option - SvelteKit does this and I used to: https://github.com/sveltejs/kit/blob/master/.vscode/settings.json

greatbacon: When is `gro format` supposed to be run? I just assumed it was being run as part of the `dev` or `build` commands?

ryanatkn: It's wasted work normally, the editor should do this work. We only ever currently automatically run `gro format` when we do `gro format --check` in `gro check` and users can compose it as a task if needed.

.gitignore
greatbacon: Is it okay if we punt on this? This `deploy` stuff is going to be relatively in flux and I don't really want to add a new directory just for deployments

src/deploy.task.ts
greatbacon: Gotcha. Either way, I just ran it and pushed the changes but the build is still breaking. It works locally so I have no idea what's causing it 😧

ryanatkn: This is a situation where I'm tempted to put it under Gro's umbrella, but feels better to hand vscode the ball and add the `.vscode` directory.

ryanatkn:
Looks good to me. If you want examples:

- args: https://github.com/feltcoop/gro/blob/main/src/serve.task.ts
- env vars: https://github.com/feltcoop/gro/blob/main/src/server/https.ts

These patterns are still in flux but I'm happy enough with how simple and useful they are right now. They don't do advanced validation.

ryanatkn (approved):

ryanatkn:
Is this after installing a new version of npm? Just saw the lockfile changes.

I think CI will pass if you rebase on main and force push. (that's the workflow I normally do, not sure it's the best)

src/deploy.task.ts
```diff
@@ -0,0 +1,29 @@
...
+			'deploy.tar',
+			DIST_DIR_NAME,
+			'package.json',
+			'package-lock.json',
+		]);
+		//scp to server
+		//your ssh key will need to be added to linode account
+		await spawnProcess('scp', ['deploy.tar', 'root@96.126.116.174:deploy.tar']);
+		//unpack & start server
+		await spawnProcess('ssh', ['root@96.126.116.174', 'tar -xvf deploy.tar;', 'npm i;']);
```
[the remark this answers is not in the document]
greatbacon: I think only `ssh` allows for composing commands like that. Also noted on the children thing.

greatbacon:
this PR is now in a weird bad state. I'm gonna close & try again with a clean branch

ryanatkn:
That works. For future reference another choice is to do something like this:

```bash
git reset --soft HEAD^
```

Repeat until you get to a good commit - like 1 back works in this case.

Then:

```bash
git commit -am "new commit"
git push --force
```

The `force` is dangerous but we have protection rules on `main`, so you can only screw up a branch you were going to close anyway.

ryanatkn:
Also merge commits, I don't get them, they make the diff unreadable.

PR #12, opened 2021-04-12
add websockets

ryanatkn:
This is a first pass at websocket support. It adds one top-level dependency, [`ws`](https://github.com/websockets/ws), and its types. It's a popular and plenty good library, and we could always swap it out for something else one day.

There are a few big things to figure out to make this work with auth. Followup work:

- proxy instead of using a separate port
- figure out auth (depends on proxy)
- display the count of connected clients on each client (and then later, show the count per community)


svelte.config.cjs
```diff
@@ -16,6 +16,11 @@ module.exports = {
 			server: {
 				proxy: {
 					'/api': 'http://localhost:3001/', // import('@feltcoop/gro/dist/config/defaultBuildConfig.js').API_SERVER_DEFAULT_PORT_DEV
+					// TODO what's the right way to configure this?
+					// '/ws': 'ws://localhost:3002/',
+					// '/ws': 'ws://localhost:3001/',
+					// 'ws://localhost:3000': 'ws://localhost:3002/',
+					// 'ws://localhost:3000': 'ws://localhost:3001/',
```
ryanatkn, line 23: I think I tried everything but I'm sure that's not true.

greatbacon (approved):
This looks like a good first pass; excited to play around with this locally! Especially in a DB context.

PR #13, opened 2021-04-12
Add communities view

greatbacon:
So this PR covers two elements related to a communities view

1. It extends the DB seed file to add a `communities` table & a relational `accounts_communities` table, plus dummy data
2. It adds a SideNav component that I've filled with some stub buttons (to be generated programmatically later).

Once this gets merged I'm going to add a route to get all the communities for a given account, then see if I can't populate the sidebar with that info programmatically.

Here's a REPL example of the sidenav:
https://svelte.dev/repl/588ac19e1ea54d99a41ac2e3734d3c75?version=3.37.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,1665 @@
 {
   "name": "@feltcoop/felt-server",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@feltcoop/felt-server",
+      "version": "0.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "@feltcoop/gro": "^0.19.0",
+        "@polka/send-type": "^0.5.2",
+        "body-parser": "^1.19.0",
+        "cookie-session": "^1.4.0",
+        "polka": "^1.0.0-next.13",
+        "postgres": "^1.0.2",
+        "sirv": "^1.0.11"
+      },
+      "devDependencies": {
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -1,8 +1,1665 @@
 {
   "name": "@feltcoop/felt-server",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
ryanatkn, line 4: npm version problem? maybe I need to upgrade..

greatbacon: No idea lol; I just googled & it said "commit your lock file" & the file had changed, so here we are :shrug:

ryanatkn: Can you share your `npm -v`? I'm on 6.14.11 and I'm guessing you're on 7? If so can you try to delete package-lock.json and see if it re-creates the same files? It's weird that it's duplicating everything.

ryanatkn (approved):
Looks good, just need to figure out the lockfile. I upgraded to npm 7.

I think we should fully refresh the lockfile because seems to be getting into weird states. It won't fully refresh it unless you delete node_modules it seems (or it's another issue) so try this:

```bash
rm -rf node_modules package-lock.json
npm i
```

ryanatkn:
Looks like we're good, bit SvelteKit has a breaking change, so change this line:

```ts
// src/hooks.ts
export const handle: Handle = async (request, render) => {
```
to: 
```ts
export const handle: Handle = async ({request, render}) => {
```

PR #14, opened 2021-04-14
create db

ryanatkn:
This adds three tasks to `src/db`:

- `db/create` - destroy+seed
- `db/destroy` - composed by `db/create`
- `db/seed` - composed by `db/create`

In the next database PR, we'll add the equivalent of `db/up` that gets composed into `db/create` between `db/destroy` and `db/seed`, extracting the schema creation from the `seed` function. The library we'll use is `ley`.

Task composition is useful but I don't want to add tasks that aren't designed for standalone use. For example, the `db/create` as implemented calls `db/seed`, but it also includes a flag to not seed, allowing users to create an empty database and then run `db/seed` manually, or anything else or nothing, if wanted, while still using `db/create`. We want to keep things simple but make the weird stuff possible; I'm not sure what things should be composed as tasks, what things should just be function calls, and what the final set of tasks exposed to users is.

This needs to be designed for our workflows, whatever those are!

One of the changes here is that the database is no longer created on startup - you now have to run `db/create` or it'll error. Not sure of the best fix that won't cause weird surprises. Maybe we can still create the database on startup if not detected, but only in development? I'm not sure.

- [x] basics
- [x] any API changes?
- [x] docs: https://github.com/feltcoop/felt-server/blob/e4dc2e640888d4239465c8acb1505118fd8d7ce1/src/db/README.md


src/db/create.task.ts
```diff
@@ -0,0 +1,22 @@
+import type {Task} from '@feltcoop/gro';
+
+import {obtainDb} from './obtainDb.js';
+
+// name? maybe `init` or `reset` is clearer?
+
+export interface TaskArgs {
+	'no-seed'?: boolean;
+	S?: boolean; // alias for `no-seed`
```
greatbacon, line 9: so if we pass an `S` to this command (i.e. `gro db/create S`) it won't seed?

src/db/destroy.task.ts
```diff
@@ -0,0 +1,20 @@
+import type {Task} from '@feltcoop/gro';
+
+import {obtainDb} from './obtainDb.js';
+import {defaultPostgresOptions} from './postgres.js';
+
+export const task: Task = {
+	description: 'destroy the database schema and delete all data',
```
greatbacon, line 7: Does this destroy all schema in a DB? Or just the `felt` one?

greatbacon (approved):
Just had a couple of clarifying questions, but otherwise LGTM!

src/db/create.task.ts
ryanatkn: Correct. I don't like the `-S` convention at all but that's the normal Bash thing right? I'm alright with removing the alias completely. Ideally this would be a `seed` flag and `-s` but because doesn't work because of Bash's lack of falsy args.

src/db/destroy.task.ts
ryanatkn: Just for the configured `PGDATABASE` which defaults to `felt`. Is there a clearer way to state this? It is a bit weird because we call Postgres a "database" but Postgres names its unit of composition a "database".

ryanatkn: I may have misunderstood the question - this raw SQL command is run during `db/destroy`, and it'll only affect the PGDATABASE.

```ts
		await db.sql.unsafe(`
			drop schema public cascade;
			create schema public;
			alter schema public owner to postgres;
			grant all on schema public to postgres;
			grant all on schema public to ${defaultPostgresOptions.username};
			grant all on schema public to public;
		`);
```

src/db/create.task.ts
ryanatkn: I removed the alias so only `gro db/create --no-seed` works. Eventually Gro will have support for aliases with automatic docs and no manual work, and until then it makes sense to avoid aliases unless they're particularly convenient.

src/db/destroy.task.ts
greatbacon: Maybe clarify it'll just remove the configured "app database"?

PR #15, opened 2021-04-15
Add route to return a community by id

greatbacon:
This is mostly an exercise for me to make sure I understand the sql driver & can interpret our code & architecture.

This adds the functionality of return a Community object if passed a `community_id` @ the `/api/v1/accounts/:id/communities/:id` endpoint.

I am attached to nothing so I look forward to seeing the notes on this.


src/communities/community.ts
```diff
@@ -0,0 +1,4 @@
+export interface Community {
+	community_id?: number;
+	name: String;
```
ryanatkn, line 3: Just a JS thing, this should be `string` - the `String` constructor object is different and best ignored. You'll probably never see it outside of niche library code.

src/server/ApiServer.ts
```diff
@@ -100,7 +101,8 @@ export class ApiServer {
 				send(res, 200, toClientContext(req));
 			})
 			.post('/api/v1/login', toLoginMiddleware(this))
-			.post('/api/v1/logout', toLogoutMiddleware(this));
+			.post('/api/v1/logout', toLogoutMiddleware(this))
+			.get('/api/v1/accounts/:account_id/communities/:community_id', toCommunityMiddleware(this));
```
ryanatkn, line 105: This prompted me to make a little PR here. https://github.com/feltcoop/felt-server/pull/16/files

I'm interpreting the `accounts/:account_id` section of the route as communicating ownership. Is that roughly right?

I think instead we want to drop that part of the route:

```ts
'/api/v1/accounts/:account_id/communities/:community_id'
// becomes
'/api/v1/communities/:community_id'
```

and then instead of the middleware's `console.log('[communityMiddlware] user', req.params.account_id);`, it'll use `req.account` (currently named `req.user`, will be changed) - see this small PR for a small change to the middleware that makes `req.user` available: https://github.com/feltcoop/felt-server/pull/16/files

I think the best answer for this PR is to have no security, and in a followup we'll connect the database accounts with `req.account` and do auth in the middleware.

src/communities/community.ts
greatbacon: Types as lowercase is gonna be a hard habit to break lol

src/server/ApiServer.ts
greatbacon: `I'm interpreting the accounts/:account_id section of the route as communicating ownership. Is that roughly right?`
Yeah, I think that's the REST way to construct the URL, but I also toyed with just eating the `accounts/:account_id` part since most of that will be living as a value in a session cookie anyway. I'll go ahead and drop it.

I like the sound of security as a follow up. I've added this card (https://github.com/orgs/feltcoop/projects/1#card-59130567) to track the work.

src/communities/communityMiddleware.ts
```diff
@@ -0,0 +1,13 @@
...
+import type {ApiServer, Middleware} from '../server/ApiServer.js';
+
+export const toCommunityMiddleware = (server: ApiServer): Middleware => {
+	const {db} = server;
+	return async (req, res) => {
+		console.log('[communityMiddlware] user', req.params.account_id); // TODO logging
+		console.log('[communityMiddlware] community', req.params.community_id);
+
+		const findCommunityResult = await db.repos.communities.findById(req.params.community_id);
+		return send(res, 200, {community: findCommunityResult}); // TODO API types
```
ryanatkn, line 11: This is a place where we'll get nice helpers and type safety, eventually.

- the `findCommunityResult` type needs to be introspected. If it's not `ok`, it needs to return a non-200 status code
- the result from `findById` in the repo is returning an array - generally `find` in the JS world means a single item, and things like `filter`/`search`/`get`/`query` either signal collections or work for both. Fixing this in the DB repo will be verbose, but we'll start thinking about helpers and patterns to keep things DRY and consistent.

src/db/Database.ts
```diff
@@ -116,5 +117,24 @@ export class Database {
...
+		communities: {
+			findById: async (
+				community_id: string,
+			): Promise<Result<{value: Community[]}, {type: 'noCommunityFound'; reason: string}>> => {
+				console.log(`[db] preparring to query for community id: ${community_id}`);
+				const data: Community[] = await this.sql`
+				select community_id, name from communities where community_id = ${community_id}
+				`;
+				console.log('[db] community data', data);
+				if (data) {
```
ryanatkn, line 129: I think this code path never gets hit because the driver is returning an empty array if nothing is found.

src/communities/community.ts
ryanatkn: haha yeah so it might be worse than you think, most types are uppercase, it's the JS primitives that aren't - `string`, `number`, `boolean`, `undefined`, `null`, probably some others

src/db/Database.ts
```diff
@@ -116,5 +117,24 @@ export class Database {
 				};
 			},
 		},
+		communities: {
+			findById: async (
+				community_id: string,
+			): Promise<Result<{value: Community[]}, {type: 'noCommunityFound'; reason: string}>> => {
+				console.log(`[db] preparring to query for community id: ${community_id}`);
+				const data: Community[] = await this.sql`
```
ryanatkn, line 125: I just learned this, we should change calls to `sql` from:

```ts
const data: Community[] = await this.sql`
```

to:

```ts
```

The driver returns an array extended with some useful properties like `command` and `columns`.

src/communities/community.ts
greatbacon: okay, actually that's not too far off for the JVM actually; it's just that Strings in Java are a Class, not a primitive🙃

ryanatkn (approved):
Looks good!

src/communities/communityMiddleware.ts
```diff
@@ -0,0 +1,24 @@
+import send from '@polka/send-type';
+import type {ApiServer, Middleware} from '../server/ApiServer.js';
+
+export const toCommunitiesMiddleware = (server: ApiServer): Middleware => {
+	const {db} = server;
+	return async (req, res) => {
+		console.log('[communityMiddlware] account', req.user); // TODO logging
+
+		const findCommunitiesResult = await db.repos.communities.filterByAccount(req.user);
```
ryanatkn, line 9: To make this typecheck, you could either check for the existence of `req.user` and branch if the user is not logged in, or change it to `req.user!` to tell TypeScript to ignore `null`/`undefined` possibilities.

In the future we'll handle this automatically in a config, so each route won't need this logic.

ryanatkn:
The formatting of the `package-lock.json` and `package.json` looks like a bug in Gro, I'll look into that now.

PR #17, opened 2021-04-16
Populate navbar

greatbacon:
Attempt to populate the navbar on the lefthand side with Community info from the server

### DIFF
--- a/src/lib/SideNav.svelte
+++ b/src/lib/SideNav.svelte
@@ -1,4 +1,6 @@
-<script>
+<script lang="ts">
+	export let communities: {community_id: number; name: string}[];
+	console.log(communities)
 	const submitName = async () => {
 		console.log('button');
 	};
@@ -8,6 +10,9 @@
 	<button type="button" on:click={submitName}> F </button>
 	<button type="button"> S </button>
 	<button type="button"> D </button>
+	<!-- {#each communities as {name} (name)}
+		<button type="button">{ name } </button>
+	{/each} -->
 </div>
 
 <style>
--- a/src/lib/SideNav.svelte
+++ b/src/lib/SideNav.svelte
@@ -1,13 +1,13 @@
 <script lang="ts">
-	export let communities: {community_id: number; name: string}[];	
-	const submitName = async (name: string) => {		
+	export let communities: {community_id: number; name: string}[];
+	const submitName = async (name: string) => {
 		console.log(name);
 	};
 </script>
 
-<div class="sidenav">	
+<div class="sidenav">
 	{#each communities as {name} (name)}
-		<button type="button"  on:click={submitName(name)}>{ name } </button>
+		<button type="button" on:click={submitName(name)}>{name} </button>
 	{/each}
 </div>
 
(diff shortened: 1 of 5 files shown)


src/lib/SideNav.svelte
```diff
@@ -1,13 +1,14 @@
...
+		console.log(name);
 	};
 </script>
 
 <div class="sidenav">
-	<button type="button" on:click={submitName}> F </button>
-	<button type="button"> S </button>
-	<button type="button"> D </button>
+	{#each communities as {name} (name)}
+		<button type="button" on:click={submitName(name)}>{name} </button>
```
ryanatkn, line 10: This is a rite of passage for most Svelte developers. (it was for me, and I saw it constantly asked in the community)

This code:

```
 on:click={submitName(name)}
```

is assigning the return value of `submitName` to the click handler, which is undefined, so no click events will be processed.

Because you need to capture `name`, doing `on:click={() => submitName(name)}` is probably the best fix.

If instead each button was its own component, and each component had `name` as a prop, you could do `on:click={submitName}` and have it use `name` from its closure.

src/lib/SideNav.svelte
```diff
@@ -1,13 +1,14 @@
-<script>
-	const submitName = async () => {
-		console.log('button');
+<script lang="ts">
+	export let communities: {community_id: number; name: string}[];
```
ryanatkn, line 2: Can import this type: `import type {Community} from '../communities/community.js';`

src/lib/SideNav.svelte
greatbacon: I'll keep the button as component thing in mind. I guess I'm a little confused though, because you're saying this shouldn't work, but it does work when I was testing it locally.

I'll definitely change the syntax to your suggestion, just curious about why something that isn't supposed to work is working.

ryanatkn: Turns out you can assign promises to `on:click` handlers for some especially surprising behavior!

src/routes/index.svelte
```diff
@@ -3,14 +3,25 @@
 	import Echo from '$lib/Echo.svelte';
 	import AccountForm from '$lib/AccountForm.svelte';
 	import SideNav from '$lib/SideNav.svelte';
+	import {onMount} from 'svelte';
 
 	const title = 'felt-server';
+	$: communities = [];
```
ryanatkn, line 9: The `$: communities =` is a mistake here - that's for reactive computed expressions, not assignable variables. `let communities =`  creates assignable reactive variables. (`const` if you don't plan on reassigning)

ryanatkn (approved):

src/routes/index.svelte
```diff
@@ -1,16 +1,33 @@
...
+	});
 </script>
 
 <svelte:head><title>{title}</title></svelte:head>
 
 <main>
-	<SideNav />
+	{#if user && !user.guest}
+		<SideNav {communities} />
+	{/if}
```
ryanatkn, line 30: :100:

Issue #7, opened 2021-04-08
Replace database write to disk w/ write to DB


greatbacon:
/src/db/database file currently write JSON to the file system, it should use our shiny new PG system.

Be mindful of not changing the interface. They are big and scary but not that big of a deal.

PR #18, opened 2021-04-17
Replace disk write

greatbacon:
Story for issue #7 

This PR removes the original disk based DB & ports to the postgres system. There may be some lines that change `User` -> `Account` but the full refactor will come in a later PR. But it will now login/create/read user accounts from the DB.

It also currently has an issue where when you logout or login; the SideNav component doesn't update the underlying `community` prop & I'm not quite sure how to fix that yet.

### DIFF
--- a/src/communities/communityMiddleware.ts
+++ b/src/communities/communityMiddleware.ts
@@ -4,13 +4,18 @@ import type {ApiServer, Middleware} from '../server/ApiServer.js';
 export const toCommunitiesMiddleware = (server: ApiServer): Middleware => {
 	const {db} = server;
 	return async (req, res) => {
+		if (!req.session.name) {
+			//TODO centralize error message strings
+			console.log('[communityMiddleware] no account to search for communities');
+			return send(res, 404, {reason: 'no communities found'});
+		}
 		console.log('[communityMiddleware] account', req.user); // TODO logging
 
 		const findCommunitiesResult = await db.repos.communities.filterByAccount(req.user!);
 		if (findCommunitiesResult.ok) {
 			return send(res, 200, {communities: findCommunitiesResult.value}); // TODO API types
 		} else {
-			console.log('no communities found');
+			console.log('[communityMiddleware] no communities found');
 			const code = findCommunitiesResult.type === 'noCommunitiesFound' ? 404 : 500;
 			return send(res, code, {reason: findCommunitiesResult.reason});
 		}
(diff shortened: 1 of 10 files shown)


src/communities/communityMiddleware.ts
```diff
@@ -4,13 +4,18 @@ import type {ApiServer, Middleware} from '../server/ApiServer.js';
 export const toCommunitiesMiddleware = (server: ApiServer): Middleware => {
 	const {db} = server;
 	return async (req, res) => {
+		if (!req.session.name) {
+			//TODO centralize error message strings
+			console.log('[communityMiddleware] no account to search for communities');
+			return send(res, 404, {reason: 'no communities found'});
```
ryanatkn, line 10: I think this should be a 401 and say you're not logged in.

src/routes/index.svelte
```diff
@@ -5,14 +5,13 @@
 	import AccountForm from '$lib/AccountForm.svelte';
 	import SideNav from '$lib/SideNav.svelte';
 	import {onMount} from 'svelte';
-	import type {ClientSession} from '../session/clientSession.js';
+	import type {ClientSession} from 'src/session/clientSession.js';
```
ryanatkn, line 8: I like `src/a` better than the `$a` pattern, happy to see this supported.

src/routes/index.svelte
```diff
@@ -5,14 +5,13 @@
...
 
 	const title = 'felt-server';
 	let user: ClientSession;
 	$: user = $session?.user;
-	let communities: Community[];
-	communities = [];
+	let communities: Community[] = [];
 
 	onMount(async () => {
 		const res = await fetch(`/api/v1/communities`);
```
ryanatkn, line 17: As a quick fix here, you could only call `fetch` if `user && !user.guest`. I think the proper fix is querying this data in the session, so `db.repos.session.loadClientSession`. Probably makes for a good standalone PR.

ryanatkn (approved):
This is awesome. Real data!

Fixing the issue for new user communities probably can be left to the session implementation (mentioned in a comment as a possible followup PR), I think that's the real fix. You could hack in something to fetch a new user's data, but it'll be fixed automatically when implemented on the `ClientSession`.

src/routes/index.svelte
greatbacon: Okay yeah, I wasn't really sure what should be living where in terms of "what goes in session" & "what gets loaded after DOM". Long term something this foundational living in the session makes sense to me.

I think a Gordian knot answer is "what if we just reload the page on login/logout". Happy to punt on both and have a sync convo about it tho.

ryanatkn: Yeah this is a good conversation to have, making a note. Punting implementation makes sense to me.

This ties into a complex architecture question. Long term we have very detailed questions to ask about caching, startup speed, etc, but for the short and medium terms, data like the communities a user belongs to is a good candidate to put on the global session.

ryanatkn (approved):
Cool the reload works for now, we'll probably remove that when we make a pass on the session.

PR #21, opened 2021-04-20
Session setting working on login/logout

greatbacon:
This moves the community data into the session for the front end.

This makes it work when logging in or out, but if you refresh the page the data disappears, and my node-fu is failing me in how to solve that.

I think it has to do with `getSession` stuff, but I'm not 100% sure on what's happening to come up with a solution.


src/session/loginMiddleware.ts
```diff
@@ -66,7 +66,11 @@ export const toLoginMiddleware = (server: ApiServer): Middleware => {
 
 		console.log('[login]', account.name); // TODO logging
 		req.session.name = account.name;
-		const clientSession = await db.repos.session.loadClientSession(account.name);
-		return send(res, 200, {session: clientSession}); // TODO API types
+		const clientSessionResult = await db.repos.session.loadClientSession(account.name);
+		if (clientSessionResult.ok) {
+			return send(res, 200, {session: clientSessionResult.value}); // TODO API types
+		} else {
+			return send(res, 500, {reason: 'problem loading client session'});
```
ryanatkn, line 73: This differs from the above pattern of `{reason: clientSessionResult.reason}` but I also like seeing it side by side, because it's not a final pattern anyway. Begs the question: should our db layer be responsible for generating user-facing messages? Or is that responsibility something to put in the middleware aka API handler layer?

src/routes/index.svelte
```diff
@@ -4,23 +4,14 @@
...
-
-	onMount(async () => {
-		const res = await fetch(`/api/v1/communities`);
-		if (res.ok) {
-			const data = await res.json();
-			communities = data.communities;
-			console.log(communities);
-		}
-	});
+	$: communities = $session?.communities;
```
ryanatkn, line 14: Thanks to this change, we can change the below code:

`{#if account && !$session.guest}`

to

`{#if communities}`

src/db/Database.ts
```diff
@@ -103,11 +108,23 @@ export class Database {
...
+					};
+				}
+
+				console.log(`[db] preparring to query for communities account: ${account.account_id}`);
 				const data = await this.sql<Community[]>`
-				SELECT c.community_id, c.name FROM communities c JOIN account_communities ac ON c.community_id=ac.community_id AND ac.account_id=1
+				SELECT c.community_id, c.name FROM communities c JOIN account_communities ac ON c.community_id=ac.community_id AND ac.account_id= ${account?.account_id}
 				`;
 				console.log('[db] community data', data);
 				if (data.length) {
```
ryanatkn, line 130: I think this probably shouldn't return an error if no communities are found. I think it should just return an empty array? That would solve what's currently an error for newly created users.

src/db/Database.ts
```diff
@@ -103,11 +108,23 @@ export class Database {
...
+				Result<
+					{value: Community[]},
+					{type: 'noCommunitiesFound'; reason: string} | {type: 'missingAccountId'; reason: string}
+				>
+			> => {
+				if (account.account_id == null) {
+					return {
+						ok: false,
+						type: 'missingAccountId',
+						reason: `No id present for account: ${account}`,
```
ryanatkn, line 121: `${account}` -> `${account.name}`

ryanatkn:
re: data gone on refresh

In `ApiServer`, for our hacky context route we have:

```ts
			.get('/api/v1/context', (req, res) => {
				send(res, 200, toClientContext(req));
			})
```

The problem (or part of it) is that the account is loaded because it's part of the session, but the rest of the info isn't there.

`toClientContext` could be changed to be `async`, so `send(res, 200, await toClientContext(req));`, and then load the communities from the database. in there. Could be refactored too.

`ClientAccountContext` needs `communities: Community[];` to match `AccountSession`. We'll refactor/rename/etc when SvelteKit is more stable.

and then in `hooks.ts` the `getSession` function needs to return the context's communities.

ryanatkn (changes requested):
I think this'll get us to a working state, or close.

src/session/loginMiddleware.ts
greatbacon: To your last point: I think at least having a common interface for client messaging (like `ClientResponse` or something) would be a decent soft reminder to not put sensitive information there.

To the bigger point of how to handle our error messaging, I'm not really sure. A pattern Kotlin has that I generally like is that you have discrete errors at the level they happen which can be logged for system purposes, then there is a catchall at the top level that is responsible for turning the lower level system errors into more generic/user friendly errors to return to the caller. Not sure what that would look like in a Node system but food for thought.

I think we may just have a bigger convo around our error handling architecture.

src/routes/index.svelte
greatbacon: I think guest == accountless is a good semantic pattern.

For the sidebar I'll adopt `{#if communities}` but in this case that's because I actually do want to hide the sidebar if there are zero communities, even if the user is logged in.

src/session/loginMiddleware.ts
ryanatkn: Yeah that makes sense, it's basically what I'm thinking, it's some top-level middleware or response handler or wrapper that standardizes everything. We'll architect that around Polka (and JSON schema validation stuff) when the time comes.

src/communities/communityMiddleware.ts
```diff
@@ -15,8 +15,8 @@ export const toCommunitiesMiddleware = (server: ApiServer): Middleware => {
 		if (findCommunitiesResult.ok) {
 			return send(res, 200, {communities: findCommunitiesResult.value}); // TODO API types
 		} else {
-			console.log('[communityMiddleware] no communities found');
-			const code = findCommunitiesResult.type === 'noCommunitiesFound' ? 404 : 500;
+			console.log('[communityMiddleware] error while searching for communities');
+			const code = findCommunitiesResult.type === 'missingAccountId' ? 400 : 500;
```
ryanatkn, line 19: I think `missingAccountId` should be a 500, so this could be simplified. Because the account is something we attach via the session, the value ought to be secure, meaning our internal systems goofed somewhere.

src/db/Database.ts
```diff
@@ -103,21 +108,21 @@ export class Database {
 			},
 			filterByAccount: async (
 				account: Account,
```
ryanatkn, line 110: What do you think of changing this API to take the `accountId` instead of the `account`? You could remove the `missingAccountId` error. I'm not sure if future requirements will need more account information.

src/communities/communityMiddleware.ts
greatbacon: accountId comes from the session header I thought, so wouldn't it be a malformed request to the server? Hence the 400.

src/db/Database.ts
greatbacon: yeah good call, principals of least concern and all that 🙃

src/session/sessionAccountMiddleware.ts
```diff
@@ -9,22 +9,14 @@ export const toSessionAccountMiddleware = (server: ApiServer): Middleware => {
...
 			console.log('resetting session, none found');
 			req.session = null!;
-			const code =
-				findAccountResult.type === 'noAccountFound'
-					? 404
-					: findAccountResult.type === 'invalidName'
-					? 400
-					: 500;
-			return send(res, code, {reason: findAccountResult.reason});
+			return send(res, 404, {reason: 'no session found'});
```
ryanatkn, line 18: I think the purpose of this middleware is to look for the `req.session` and attach the account info if it's available. Its previous role - returning the error - really belongs in a different layer. When those layers e.g. need an account and don't have one available, they'll return a `401`: <https://httpstatusdogs.com/401-unauthorized>

src/session/sessionAccountMiddleware.ts
```diff
@@ -9,22 +9,14 @@ export const toSessionAccountMiddleware = (server: ApiServer): Middleware => {
 		}
 
 		console.log('[sessionAccountMiddleware]', req.session.name); // TODO logging
```
ryanatkn, line 11: Because the purpose of this middleware is to either add the `req.accountSession` or clear the `req.session`, I think we should do a check for `req.accountSession` here - if it exists, let's throw an error for ourselves.

src/server/ApiServer.ts
```diff
@@ -25,15 +25,16 @@ import {
...
 import type {Database} from '../db/Database.js';
+import type {Community} from 'src/communities/community.js';
 
 const log = new Logger([blue('[ApiServer]')]);
 
 // TODO not sure what these types should look like in their final form,
 // there's currently some redundancy and weirdness
 export interface Request extends PolkaRequest, CookieSessionRequest {
-	account?: Account;
+	accountSession?: AccountSession;
```
ryanatkn, line 37: good stuff

src/session/sessionAccountMiddleware.ts
greatbacon: So I think this is the last blocking piece, but I'm not really sure I understand what you're asking for/describing here. Could you draft a more concrete example of what you have in mind?

ryanatkn: This is what I was thinking, does it make sense?

```ts
import send from '@polka/send-type';

import type {ApiServer, Middleware} from '../server/ApiServer.js';

export const toSessionAccountMiddleware = (server: ApiServer): Middleware => {
	return async (req, res, next) => {
		if (req.accountSession) {
			return send(res, 500, {reason: 'invalid server configuration'});
		}
```

ryanatkn: Incidentally, I think this change will make CI pass.

PR #23, opened 2021-04-28
Initial pass added Space

greatbacon:
This PR adds a Space model/interface; defines a table for it & some initial data in the seed file; and has a simple query to read from the DB for a space given an id (testable from the browser url `localhost:3000/api/v1/spaces/2`

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -12,6 +12,8 @@ export interface Options {
 	sql: PostgresSql;
 }
 
+//TODO create seperate models used by the front end (w/ camelCase attributes) from the repo models
+// and snake_case for the DB stuff
 export class Database {
 	sql: PostgresSql;
 
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -122,7 +122,7 @@ export class Database {
 			): Promise<Result<{value: Space[]}, {type: 'noSpaceFound'; reason: string}>> => {
 				console.log(`[db] preparring to query for space id: ${spaceId}`);
 				const data = await this.sql<Space[]>`
-				select spaceId, url, mediaType, content from spaces where spaceId = ${spaceId}
+				select space_id, url, mediaType, content from spaces where space_id = ${spaceId}
 				`;
 				console.log('[db] space data', data);
 				if (data.length) {
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -3,6 +3,7 @@ import {unwrap} from '@feltcoop/gro';
 
 import type {AccountSession} from '../session/clientSession.js';
 import type {Community} from '../communities/community.js';
+import type {Space} from '../spaces/space.js';
 import type {Account} from '../vocab/account/account.js';
 import type {Entity} from '../vocab/entity/entity.js';
 import type {PostgresSql} from './postgres.js';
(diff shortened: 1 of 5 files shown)


src/db/Database.ts
```diff
@@ -115,5 +116,24 @@ export class Database {
 				return {ok: true, value: data};
 			},
 		},
+		spaces: {
+			findById: async (
+				spaceId: string,
+			): Promise<Result<{value: Space[]}, {type: 'noSpaceFound'; reason: string}>> => {
+				console.log(`[db] preparring to query for space id: ${spaceId}`);
+				const data = await this.sql<Space[]>`
+				select spaceId, url, mediaType, content from space where spaceId = ${spaceId}
```
ryanatkn, line 125: `space` -> `spaces`

src/spaces/space.ts
```diff
@@ -0,0 +1,6 @@
+export interface Space {
+	spaceId?: number;
```
ryanatkn, line 2: We're running afoul of Postgres' case insensitivity here. We need to either go lowercase in the database or quote everything. I'm partial to underscores for property names, but things like the `mediaType` below are ActivityStreams vocab - which is awesome - but I'm wondering what the best long term convention is.

greatbacon: okay, I wasn't sure if that was a PSQL thing or something overlooked. I'm fine with `snake_case` propagating or maybe creating DB level interfaces with functions to convert them into the broader `camelCase` form?

src/db/Database.ts
```diff
@@ -122,7 +122,7 @@ export class Database {
 			): Promise<Result<{value: Space[]}, {type: 'noSpaceFound'; reason: string}>> => {
 				console.log(`[db] preparring to query for space id: ${spaceId}`);
 				const data = await this.sql<Space[]>`
-				select spaceId, url, mediaType, content from space where spaceId = ${spaceId}
+				select spaceId, url, mediaType, content from spaces where spaceId = ${spaceId}
```
ryanatkn, line 125: i leave a file unsaved about every week, and it is never not baffling

src/spaces/space.ts
ryanatkn: I like that, we can build views with `camelCase`, and we'd probably use views anyway.

src/db/seed.ts
```diff
@@ -44,7 +44,7 @@ export const seed = async (db: Database): Promise<void> => {
 
 	const createSpacesTableResult = await sql`
 	create table if not exists spaces (
-		spaceId serial primary key,
+		space_id serial primary key,
 		url text,
 		mediaType text,
```
ryanatkn, line 49: we could call this `mime` for now, or `media_type` if you prefer

greatbacon: oh yep, the other 🐍 convert

I'll stick with `media_type` since that's the AS defined term

greatbacon:
@ryanatkn are there any outstanding notes on this PR, or am I good to merge?

ryanatkn:
My bad, I was looking specifically for that media_type commit and I must've gotten tangled with my notifications and missed it.

ryanatkn (approved):

greatbacon:
@ryanatkn no worries, there was some wonkiness in the change getting pushed yesterday, so I'm not surprised it slipped under the radar. Thanks!

PR #24, opened 2021-04-30
Added  community-space relations

greatbacon:
So this adds db & seed data for relating a space to a community.

I also want to add a component that will render the url of the spaces in a nav drawer to the right of the community buttons, but I'm not really 100% sure where to start on the architecture...

would appreciate thoughts & feedback!

### DIFF
--- a/src/server/ApiServer.ts
+++ b/src/server/ApiServer.ts
@@ -27,7 +27,7 @@ import {
 	toCommunityMiddleware,
 	toCommunitiesMiddleware,
 } from '../communities/communityMiddleware.js';
-import {toSpaceMiddleware} from '../spaces/spaceMiddleware.js';
+import {toSpaceMiddleware, toSpacesMiddleware} from '../spaces/spaceMiddleware.js';
 import type {ClientAccount, AccountSession} from '../session/clientSession.js';
 import type {Database} from '../db/Database.js';
 import type {Community} from 'src/communities/community.js';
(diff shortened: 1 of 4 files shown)


src/server/ApiServer.ts
```diff
@@ -121,7 +121,8 @@ export class ApiServer {
 			.post('/api/v1/logout', toLogoutMiddleware(this))
 			.get('/api/v1/communities', toCommunitiesMiddleware(this))
 			.get('/api/v1/communities/:community_id', toCommunityMiddleware(this))
-			.get('/api/v1/spaces/:spaceId', toSpaceMiddleware(this));
+			.get('/api/v1/spaces/:spaceId', toSpaceMiddleware(this))
+			.get('/api/v1/communities/:community_id/spaces', toSpacesMiddleware(this));
```
ryanatkn, line 125: nice

src/spaces/spaceMiddleware.ts
```diff
@@ -17,3 +17,19 @@ export const toSpaceMiddleware = (server: ApiServer): Middleware => {
 		}
 	};
 };
+
+//Returns all spaces in a given community
+export const toSpacesMiddleware = (server: ApiServer): Middleware => {
```
ryanatkn, line 22: I like that you put this here

ryanatkn:
Not something to incorporate here - but we'll probably adjust our usage of `middleware` to be the generic stuff, and then have a single resource middleware that handles our resource patterns, standardizing as much as makes sense over one entrypoint.

ryanatkn:
I'd say let's merge this if that's ok, and then we'll do a single PR that gets the spaces to the client, so we can see all of the pieces that are involved. We need to heavily refactor that part of that app, and seeing that entire flow with no other noise will be helpful.

ryanatkn (approved):

greatbacon:
@ryanatkn I think a nested data structure makes sense here (since spaces can currently only be in the context of 1 community) but I'll go ahead and do that work in another PR. 🚢

PR #26, opened 2021-05-04
Add spaces to context

greatbacon:
This PR adds `spaces` as an attribute of a community, pushes it up with an account's context, and then renders it.

@ryanatkn I think I'm barking up the wrong tree as far as reactivity goes. I was trying to get the selected community back up into the workspace and down into the `SpacesNav` but it clearly isn't working.

All the front end architecture is WIP, so open to any suggestions or paths forward there.

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -110,9 +110,17 @@ export class Database {
 				};
 			},
 			filterByAccount: async (accountId: number): Promise<Result<{value: Community[]}>> => {
-				console.log(`[db] preparring to query for communities account: ${accountId}`);
-				const data = await this.sql<Community[]>`
-				SELECT c.community_id, c.name FROM communities c JOIN account_communities ac ON c.community_id=ac.community_id AND ac.account_id= ${accountId}
+				console.log(`[db] preparring to query for communities & spaces account: ${accountId}`);
+				const data = await this.sql<Community[]>`		
+  					select c.community_id, c.name,
+    				(
+      				select array_to_json(array_agg(row_to_json(d)))
+      				from (
+        				SELECT s.space_id, s.url, s.media_type, s.content FROM spaces s JOIN community_spaces cs ON s.space_id=cs.space_id AND cs.community_id=c.community_id      				
+      				) d
+    				) as spaces
+  				from communities c JOIN account_communities ac
+  				ON c.community_id=ac.community_id AND ac.account_id=${accountId}				
 				`;
 				console.log('[db] community data', data);
 				return {ok: true, value: data};
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -114,7 +114,7 @@ export class Database {
 				const data = await this.sql<Community[]>`		
   					select c.community_id, c.name,
     				(
-      				select array_to_json(array_agg(row_to_json(d)))
+      				select array_to_json(coalesce(array_agg(row_to_json(d)), '{}'))
       				from (
         				SELECT s.space_id, s.url, s.media_type, s.content FROM spaces s JOIN community_spaces cs ON s.space_id=cs.space_id AND cs.community_id=c.community_id      				
       				) d
--- a/src/lib/CommunityNav.svelte
+++ b/src/lib/CommunityNav.svelte
@@ -2,10 +2,11 @@
 	import type {Community} from '../communities/community.js';
 
 	export let communities: Community[];
-	export let communityFocus: Community;
-	const changeFocus = async (index: number) => {
-		$: communityFocus = communities[index];
-	};
+	export let selectedCommunity: Community;
+	//export const selectCommunity;
+	let selectedCommunityIndex: number | null = null;
+	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];
+	const changeFocus = (index: number) => selectedCommunityIndex = index;	
 </script>
 
 <div class="sidenav">
--- a/src/lib/CommunityNav.svelte
+++ b/src/lib/CommunityNav.svelte
@@ -5,13 +5,12 @@
 	export let selectedCommunity: Community;
 	export let selectCommunity: (community: Community) => void;
 	let selectedCommunityIndex: number | null = null;
-	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];
-	const changeFocus = (index: number) => selectCommunity(communities[index]);
+	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];	
 </script>
 
 <div class="sidenav">
 	{#each communities as community, i}
-		<button type="button" on:click={() => changeFocus(i)}>{community.name} </button>
+		<button type="button" on:click={() => selectCommunity(communities[i])}>{community.name} </button>
 	{/each}
 </div>
 
--- a/src/lib/CommunityNav.svelte
+++ b/src/lib/CommunityNav.svelte
@@ -2,14 +2,15 @@
 	import type {Community} from '../communities/community.js';
 
 	export let communities: Community[];
-	const submitName = async (name: string) => {
-		console.log(name);
+	export let communityFocus: Community;
+	const changeFocus = async (index: number) => {
+		$: communityFocus = communities[index];
 	};
 </script>
 
 <div class="sidenav">
-	{#each communities as {name} (name)}
-		<button type="button" on:click={() => submitName(name)}>{name} </button>
+	{#each communities as community, i}
+		<button type="button" on:click={() => changeFocus(i)}>{community.name} </button>
 	{/each}
 </div>
 
@@ -18,9 +19,10 @@
 		border: 1px outset grey;
 		background-color: lightGreen;
 		height: 75px;
-		width: 90%;
+		width: 75px;
 		cursor: pointer;
 		margin: 5%;
+		word-wrap: break-word;
 	}
 
 	button:active {
@@ -28,9 +30,8 @@
 	}
 
 	.sidenav {
-		width: 5%;
+		width: 85px;
 		height: 100%;
-		border: 2px outset black;
 		position: fixed;
 	}
 </style>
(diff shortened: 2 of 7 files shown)


src/lib/CommunityNav.svelte
```diff
@@ -1,16 +1,16 @@
 <script lang="ts">
 	import type {Community} from '../communities/community.js';
 
-	//TODO there is an issue with the communities prop not properly clearing/updating on login/logout
 	export let communities: Community[];
-	const submitName = async (name: string) => {
-		console.log(name);
+	export let communityFocus: Community;
+	const changeFocus = async (index: number) => {
+		$: communityFocus = communities[index];
```
ryanatkn, line 7: The `$: ` labels don't work inside function bodies, only at the top-level of a `script` in components. I think you want to track a `let focusedCommunityIndex: number | null = null;` declaration, then `$: focusedCommunity = communities[focusedCommunityIndex];`, and then a `changeFocus` function could just assign `focusedCommunityIndex = index;`.

ryanatkn: Was looking at the code below, you could just store the community directly:

```ts
let selectedCommunity: Community | null = null;
cost selectCommunity = (community: Community) => {
  selectedCommunity = community;
};
```

```svelte
{#each communities as community}
	<button type="button" on:click={() => selectCommunity(community)}>{community.name} </button>
{/each}
```

ryanatkn: I wrote these before seeing the usage in `Workspace.svelte`. So what I think you want is to make `selectedCommunity` the source of truth that lives as `let selectedCommunity: Community | null = null;` in `Workspace.svelte`, and then `CommunityNav.svelte` takes a `selectCommunity` function as a prop.

src/routes/index.svelte
```diff
@@ -40,17 +30,18 @@
 
 	main {
 		text-align: center;
-		padding: 1em;
-		margin: 0 auto;
 		padding: 0 auto;
 	}
 
+	body {
```
ryanatkn, line 36: This isn't actually being applied is it, because it's scoped to the component and there's no `<body>` in here?

src/lib/Workspace.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
+	import CommunityNav from '$lib/CommunityNav.svelte';
+	import SpaceNav from '$lib/SpaceNav.svelte';
+	import type {Community} from 'src/communities/community.js';
+
+	export let communities: Community[];
+	let communityFocus: Community;
+	$: communityFocus = communities[0];
+	$: console.log(`the current community spaces are ${communityFocus.spaces}`);
```
ryanatkn, line 9: to recap the stuff above, the straightforward way to make this reactive is to pass a function like this to child components:

```ts
let selectedCommunity = communities[0];
cost selectCommunity = (community: Community) => {
  selectedCommunity = community;
};
```

src/routes/index.svelte
```diff
@@ -17,16 +15,8 @@
 <svelte:head><title>{title}</title></svelte:head>
 
 <main>
-	{#if communities}
-		<SideNav {communities} />
-	{/if}
+	<Workspace {communities} />
```
ryanatkn, line 18: :+1:

ryanatkn (changes requested):
I can't test atm, will try to diagnose when we talk. (I doubt it's anything in this PR)

src/routes/index.svelte
greatbacon: yeah, this is leftover from something else, will be cleared up.

src/lib/Workspace.svelte
greatbacon: I did a little searching, but I'm not really turning up a great example of how to pass a function as a prop? Any guides/pointers?

src/lib/CommunityNav.svelte
```diff
@@ -3,10 +3,11 @@
 
 	export let communities: Community[];
 	export let selectedCommunity: Community;
-	//export const selectCommunity;
+	export let selectCommunity: (community: Community) => void;
 	let selectedCommunityIndex: number | null = null;
-	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];
-	const changeFocus = (index: number) => selectedCommunityIndex = index;	
+	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];	
+	const changeFocus = (index: number) => selectCommunity(communities[index]);
```
ryanatkn, line 9: To simplify I'd remove this and call `selectCommunity` directly inside the #each

src/lib/SpaceNav.svelte
```diff
@@ -8,9 +8,11 @@
 </script>
 
 <div class="sidenav">
-	{#each spaces as {url} (url)}
-		<button type="button" on:click={() => submitName(url)}>{url} </button>
-	{/each}
+	{#if spaces}
```
ryanatkn, line 11: Since the type of `spaces` is non-nullable we don't want `{#if spaces}`, and it makes sense we'd only mount this component if `spaces` was an array.

If you wanted to handle the case where there's an empty array, there's actually an optional `{:else}` block:

```svelte
	{#each spaces as {url} (url)}
		<button type="button" on:click={() => submitName(url)}>{url} </button>
	{:else}
		(no spaces)
	{/each}
```

Though my guess is that's not very helpful in this example.

greatbacon: The browser was throwing this error if I tried to iterate over it

```
index.mjs?v=33425075:1650 Uncaught (in promise) Error: {#each} only iterates over array-like objects.
    at validate_each_argument (index.mjs?v=33425075:1650)
    at Object.update [as p] (SpaceNav.svelte:11)
    at update (index.mjs?v=33425075:764)
    at flush (index.mjs?v=33425075:732)
```
Maybe that's more indicative of a server side issue passing a null back for spaces instead of an empty array... 🤔

src/lib/CommunityNav.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+	import type {Community} from '../communities/community.js';
+
+	export let communities: Community[];
+	export let selectedCommunity: Community;
+	export let selectCommunity: (community: Community) => void;
+	let selectedCommunityIndex: number | null = null;
+	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];
```
ryanatkn, line 8: Sorry I think this was confused by my multiple comments. The two lines here should be removed since they're being passed in as props. My previous comments assumed they were local to the component.

src/lib/CommunityNav.svelte
```diff
@@ -0,0 +1,39 @@
...
+</script>
+
+<div class="sidenav">
+	{#each communities as community, i}
+		<button type="button" on:click={() => selectCommunity(communities[i])}
```
ryanatkn, line 13: `selectCommunity(communities[i])` can be `selectCommunity(community)`

src/lib/CommunityNav.svelte
```diff
@@ -0,0 +1,39 @@
...
+
+	export let communities: Community[];
+	export let selectedCommunity: Community;
+	export let selectCommunity: (community: Community) => void;
+	let selectedCommunityIndex: number | null = null;
+	$: selectedCommunity = selectedCommunityIndex ?? communities[selectedCommunityIndex];
+</script>
+
+<div class="sidenav">
+	{#each communities as community, i}
```
ryanatkn, line 12: With the change below this you don't need the index, so:

`{#each communities as community (community.community_id)}`

src/lib/Workspace.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
+	import CommunityNav from '$lib/CommunityNav.svelte';
+	import SpaceNav from '$lib/SpaceNav.svelte';
+	import type {Community} from 'src/communities/community.js';
+
+	export let communities: Community[];
+	$: selectedCommunity = communities[0];
+	const selectCommunity = (community: Community) => {
+		$: selectedCommunity = community;
```
ryanatkn, line 9: no need for the `$: ` here

src/lib/Workspace.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
+	import CommunityNav from '$lib/CommunityNav.svelte';
+	import SpaceNav from '$lib/SpaceNav.svelte';
+	import type {Community} from 'src/communities/community.js';
+
+	export let communities: Community[];
+	$: selectedCommunity = communities[0];
```
ryanatkn, line 7: should be `let selectedCommunity = communities[0];`

src/lib/SpaceNav.svelte
ryanatkn: Yeah we'd want to fix that upstream I think, an empty array makes usage much cleaner. I think we want to set any `null` values to `[]`. If Postgres doesn't make that easy we can just hackily iterate over them and assign `[]`, and revisit it when appropriate.

src/db/Database.ts
```diff
@@ -114,7 +114,7 @@ export class Database {
 				const data = await this.sql<Community[]>`		
   					select c.community_id, c.name,
     				(
-      				select array_to_json(array_agg(row_to_json(d)))
+      				select array_to_json(coalesce(array_agg(row_to_json(d)), '{}'))
```
ryanatkn, line 117: Not sure at all about this, but should it be `'[]'` instead?

greatbacon: That's what the documentation said to use & I confirmed it was returning an empty JSON array as expected. I think `{}` is the SQL notation for an empty set, which then gets picked up by the next `array_to_json`

ryanatkn: Got it thanks for the explanation.

greatbacon:
@ryanatkn anything else you can see? Or are we g2g?

ryanatkn (approved):

ryanatkn:
gtg! it's coming allliiiiive

PR #28, opened 2021-05-06
Render posts by space

greatbacon:
Here's a PR that connects posts in the DB to their respective spaces on the front end.

### DIFF
--- a/src/lib/ChatRoom.svelte
+++ b/src/lib/ChatRoom.svelte
@@ -3,13 +3,14 @@
 	import type {Post} from '../posts/post.js';
 
 	export let spaceId: number;
-	export let props: Object;
+    export let props: Object;
+    
+    $: console.log(`[chatRoom] fetching posts for ${spaceId}`);
 
 	let posts: Post[];
 	$: posts = [];
 
 	onMount(async () => {
-		console.log(`[chatRoom] fetching posts for ${spaceId}`);
 		const res = await fetch(`/api/v1/spaces/${spaceId}/posts`);
 		if (res.ok) {
 			const data = await res.json();
--- a/src/lib/ChatRoom.svelte
+++ b/src/lib/ChatRoom.svelte
@@ -6,9 +6,10 @@
 	export let props: Object;
 
 	let posts: Post[];
-	posts = [];
+	$: posts = [];
 
 	onMount(async () => {
+		console.log(`[chatRoom] fetching posts for ${spaceId}`);
 		const res = await fetch(`/api/v1/spaces/${spaceId}/posts`);
 		if (res.ok) {
 			const data = await res.json();
--- a/src/lib/ChatRoom.svelte
+++ b/src/lib/ChatRoom.svelte
@@ -1,23 +1,24 @@
 <script lang="ts">
-	import {onMount} from 'svelte';
 	import type {Post} from '../posts/post.js';
+	import {browser} from '$app/env';
+	$: browser && loadPosts(spaceId);
 
 	export let spaceId: number;
-    export let props: Object;
-    
-    $: console.log(`[chatRoom] fetching posts for ${spaceId}`);
+	export let props: Object;
+
+	$: console.log(`[chatRoom] fetching posts for ${spaceId}`);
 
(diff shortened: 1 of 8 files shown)


src/lib/ChatRoom.svelte
```diff
@@ -0,0 +1,31 @@
...
+	import {browser} from '$app/env';
+	$: browser && loadPosts(spaceId);
+
+	export let spaceId: number;
+	export let props: Object;
+
+	$: console.log(`[chatRoom] fetching posts for ${spaceId}`);
+
+	let posts: Post[];
+	$: posts = [];
```
ryanatkn, line 12: since `posts` is a variable that gets assigned, not computed, we don't need the `$:`, so these two lines should be: `let posts: Post[] = []`

src/lib/ChatRoom.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+	import type {Post} from '../posts/post.js';
+	import {browser} from '$app/env';
+	$: browser && loadPosts(spaceId);
+
+	export let spaceId: number;
+	export let props: Object;
```
ryanatkn, line 7: This looks copied from some of my other code, we don't need `props` here yet, it's for mounting views generically in the WIP UI system.

src/lib/Workspace.svelte
```diff
@@ -1,21 +1,36 @@
...
 	};
+	let selectedSpace = selectedCommunity.spaces[0];
+	const selectSpace = (space: Space) => {
+		selectedSpace = space;
+		console.log(`[ss] ${selectedSpace.url}`);
+	};
+	let selectedSpaceContent: Object;
+	if (selectedSpace.media_type == 'application/json') {
+		selectedSpaceContent = JSON.parse(selectedSpace.content);
+	}
```
ryanatkn, line 21: This is interesting, to make it reactive:

```ts
$: selectedSpaceContent = selectedSpace.media_type === 'application/json'
	? JSON.parse(selectedSpace.content)
    : {props: {}}; // or whatever default value makes sense
```

Another option would be to interpret the content as plain text if the media type isn't json, and then render it differently if it's a string or object. Depends on what use case you're thinking of, feel free to do whatever, doesn't have to be thought through.

```ts
$: selectedSpaceContent = selectedSpace.media_type === 'application/json'
	? JSON.parse(selectedSpace.content)
    : {props: {}}; // or whatever default value makes sense
```

(small aside: use `===` except in one circumstance, `== null` which catches `null` and `undefined` - we will probably start using eslint to catch things like this)

ryanatkn (approved):
awesome work!

src/lib/Workspace.svelte
```diff
@@ -1,21 +1,36 @@
...
 	<section class="communitynav">
 		<CommunityNav {communities} {selectedCommunity} {selectCommunity} />
 	</section>
-	<section class="spacenav"><SpaceNav spaces={selectedCommunity.spaces} /></section>
-	<div class="viewfinder">"hello this is where a chat box would go"</div>
+	<section class="spacenav">
+		<SpaceNav spaces={selectedCommunity.spaces} {selectedSpace} {selectSpace} />
+	</section>
+	<div class="viewfinder">
+		<ChatRoom spaceId={selectedSpace.space_id} props={selectedSpaceContent.props} />
```
ryanatkn, line 32: Not something to necessarily change, more of a question to think about - should this `ChatRoom` component take the space as a prop? That way it could do the interpreting/rendering of its content, simplifying the external interface.

src/lib/ChatRoom.svelte
greatbacon: I'll leave a `//TODO refactor?` here & glad to hear you have a vision, I was just like "I think we'll need this in the component, so how do I do that?" lol

src/lib/Workspace.svelte
greatbacon: Yeah, that's a good point! Especially considering these sorts of components are essentially specific implementations of the more generic `Space`. I think that could be a solid pattern to establish now.

PR #29, opened 2021-05-12
Add create posts flow

greatbacon:
This is an initial pass at a create posts flow. It takes text from a form, passes it as content to an endpoint with the actor & space context, builds a post, and inserts it into the DB.

This flow will not dynamically re-render new posts made by anyone in the space atm.

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -155,10 +155,15 @@ export class Database {
 			},
 		},
 		posts: {
-			insert: async (post: Post): Promise<Result<{value: Post}>> => {
-				console.log(`[db] inserting post ${post}`);
+			insert: async (
+				actor_id: number,
+				space_id: string,
+				content: string,
+			): Promise<Result<{value: Post}>> => {
 				const data = await this.sql<Post[]>`
-					INSERT INTO posts ${this.sql(post, 'content', 'actor_id', 'space_id')}
+					INSERT INTO posts (actor_id, space_id, content) VALUES (
+						${actor_id},${space_id},${content}
+					)
 				`;
 				console.log(data);
 				return {ok: true, value: data[0]};
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -8,6 +8,7 @@ import type {Post} from '../posts/post.js';
 import type {Account} from '../vocab/account/account.js';
 import type {Entity} from '../vocab/entity/entity.js';
 import type {PostgresSql} from './postgres.js';
+import {addSyntheticLeadingComment} from 'typescript';
 
 export interface Options {
 	sql: PostgresSql;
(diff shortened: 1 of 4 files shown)


src/db/Database.ts
```diff
@@ -8,7 +8,6 @@ import type {Post} from '../posts/post.js';
 import type {Account} from '../vocab/account/account.js';
 import type {Entity} from '../vocab/entity/entity.js';
 import type {PostgresSql} from './postgres.js';
-import {addSyntheticLeadingComment} from 'typescript';
```
ryanatkn: vscode has automatic import functionality that can lead to *very* surprising results, happens to me often

src/server/ApiServer.ts
```diff
@@ -124,6 +124,7 @@ export class ApiServer {
 			.get('/api/v1/communities/:community_id', toCommunityMiddleware(this))
 			.get('/api/v1/spaces/:spaceId', toSpaceMiddleware(this))
 			.get('/api/v1/communities/:community_id/spaces', toSpacesMiddleware(this))
+			.post('/api/v1/spaces/:spaceId/posts', createPostMiddleware(this))
```
ryanatkn, line 127: following the others this should be named `toCreatePostMiddleware`, the leading "create" is more confusing than normal because you mean "create posts" not "create middleware"

these names aren't great, and they'll be factored out post-MVP

src/lib/ChatRoom.svelte
```diff
@@ -23,7 +44,13 @@
 	{#each posts as post (post.post_id)}
 		<div>{post.actor_id} said: {post.content}</div>
 	{/each}
+	<input type="text" placeholder="> chat" on:keydown={onKeyDown} bind:value={text} />
 </div>
 
 <style>
+	.chatRoom {
```
ryanatkn, line 51: normally the CSS class naming convention would be dash-case, so `.chat-room` and `class="chat-room"`

It might make more sense to name that `<div>` a `"posts"` instead or something. I started using the names of the components less in places like this - the browser dev tools offer more visibility into how components translate to markup, but I haven't used them in a while. <https://github.com/RedHatter/svelte-devtools>

src/posts/postMiddleware.ts
```diff
@@ -20,3 +20,26 @@ export const toPostsMiddleware = (server: ApiServer): Middleware => {
...
+			//TODO centralize error message strings
+			console.log('[postMiddleware] no account to search for communities');
+			return send(res, 401, {reason: 'not logged in'});
+		}
+		console.log('[postMiddleware] space', req.params.spaceId);
+		console.log('[postMiddleware] body', req.body);
+
+		//TODO take content from body & build post to pass along with it
+
+		const findPostsResult = await db.repos.posts.insert(req.body);
```
ryanatkn, line 37: Just a suggestion, we could keep this API simpler for now and just take a `content` value from the client, which we'll interpret as plain text.

So the API would change:

```ts
const findPostsResult = await db.repos.posts.insert(accountId, req.params.spaceId, req.body.content);
```

and those API changes would flow up/downstream

also, that `accountId` - is it `req.accountSession.account.account_id`?

I wonder if we want `session.accountId` instead of, or in addition to, `session.name`. I'm being lazy right now, I could answer more of this in a followup.

greatbacon: We'll probably want to refactor how/what is being stored in the account session. I'm gonna punt on that for now tho

ryanatkn (approved):

src/lib/ChatRoom.svelte
```diff
@@ -14,7 +15,27 @@
...
+			console.log('post sent');
+		} else {
+			console.error('error sending post');
+		}
+		text = '';
+	};
+
+	const onKeyDown = (e: KeyboardEvent) => {
+		if (e.key === 'Enter') {
+			createPost();
```
ryanatkn, line 38: as a hacky "see this work" thing, I think you could change this to:

```ts
	const onKeyDown = async (e: KeyboardEvent) => {
		if (e.key === 'Enter') {
			await createPost();
			await loadPosts(space.space_id);
		}
	};
```

PR #30, opened 2021-05-12
fix a type

ryanatkn:
I think this is a better pattern for now, fixes an error in a Svelte component.

`SpacePartial` should be used in places where we don't yet have the id.

It could probably use a better name, but the idea is that we have a separate document type in some circumstances.

### DIFF
--- a/src/spaces/space.ts
+++ b/src/spaces/space.ts
@@ -1,6 +1,10 @@
+import type {PartialOnly} from '@feltcoop/gro';
+
 export interface Space {
-	space_id?: number;
+	space_id: number;
 	url: string;
 	media_type: string;
 	content: string;
 }
+
+export type SpacePartial = PartialOnly<Space, 'space_id'>;
--- a/src/spaces/space.ts
+++ b/src/spaces/space.ts
@@ -1,10 +1,6 @@
-import type {OmitStrict} from '@feltcoop/gro';
-
 export interface Space {
 	space_id: number;
 	url: string;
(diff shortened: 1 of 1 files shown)


src/spaces/space.ts
```diff
@@ -1,6 +1,10 @@
...
+
 export interface Space {
-	space_id?: number;
+	space_id: number;
 	url: string;
 	media_type: string;
 	content: string;
 }
+
+export type SpacePartial = PartialOnly<Space, 'space_id'>;
```
greatbacon, line 10: Does this PartialOnly give us a subset of Space that excludes `space_id`? Or the inverse where it's only `space_id`?

ryanatkn: Subset that makes `space_id` optional:

```ts
type SpacePartial = {
    space_id?: number | undefined;
} & {
    url: string;
    media_type: string;
    content: string;
}
```

The inverse is `PartialExcept`.

ryanatkn: Oh hmm, actually wait, I have a new PR.

greatbacon:
Yeah, that optional param was something I picked up for inserts with incrementing IDs handled by the DB. However with the first pass on creating something, I ended up not even using the pattern that necessitated the optionality in the first place. 🤔

ryanatkn:
Gotcha, then yeah no need for the type at all.

ryanatkn:
never mind, breaks other things and isn't worth worrying about now

PR #31, opened 2021-05-17
Broadcast posts

greatbacon:
This PR is a first draft at utilizing the websocket tooling to broadcast messages to other clients as they are created.

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -163,9 +163,9 @@ export class Database {
 				const data = await this.sql<Post[]>`
 					INSERT INTO posts (actor_id, space_id, content) VALUES (
 						${actor_id},${space_id},${content}
-					)
+					) RETURNING *
 				`;
-				console.log(data);
+				console.log('[db] create post', data);
 				return {ok: true, value: data[0]};
 			},
 			filterBySpace: async (spaceId: string): Promise<Result<{value: Post[]}>> => {
--- a/src/lib/ChatRoom.svelte
+++ b/src/lib/ChatRoom.svelte
@@ -1,25 +1,25 @@
 <script lang="ts">
-	import type {Post} from '../posts/post.js';
 	import {browser} from '$app/env';
 	import type {Space} from '../spaces/space.js';
-	import MessageList from '$lib/MessageList.svelte';
-	import {messages} from '$lib/messagesStore';
+	import PostList from '$lib/PostList.svelte';
+	import {posts} from '$lib/postStore';
+	import type {Json} from '@feltcoop/gro/dist/utils/json.js';
 	import type {SocketStore} from './socketStore.js';
+	import {getContext} from 'svelte';
 
 	export let socket: SocketStore;
+	socket = getContext('socket');
 
 	export let space: Space;
 	export let text = '';
 	$: browser && loadPosts(space.space_id);
 	$: console.log(`[chatRoom] fetching posts for ${space.space_id}`);
 
-	let posts: Post[] = [];
-
 	const loadPosts = async (spaceId: number) => {
 		const res = await fetch(`/api/v1/spaces/${spaceId}/posts`);
 		if (res.ok) {
 			const data = await res.json();
-			posts = data.posts;
+			posts.update(() => data.posts);
 		}
 	};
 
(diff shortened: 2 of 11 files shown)


src/db/Database.ts
```diff
@@ -163,9 +163,9 @@ export class Database {
 				const data = await this.sql<Post[]>`
 					INSERT INTO posts (actor_id, space_id, content) VALUES (
 						${actor_id},${space_id},${content}
-					)
```
greatbacon, line 166: This returns the object that was inserted (which gives it back to the client to work with how it will)

src/lib/ChatRoom.svelte
```diff
@@ -26,26 +31,36 @@
 			body: JSON.stringify({content: text}),
 		});
 		if (res.ok) {
-			console.log('post sent');
+			console.log('post sent, broadcasting to server');
+			const data = await res.json();
+			broadcastPost(data);
```
greatbacon, line 36: Obviously this leaves us with 2 round trips per message sent; follow up to this work in another PR is to just bundle all this together.

src/lib/postStore.ts
```diff
@@ -0,0 +1,7 @@
+import {writable} from 'svelte/store';
+
+import type {Post} from '../posts/post.js';
```
greatbacon, line 3: I went ahead and made this second form of the Store since we haven't talked about the server interface & I didn't want to muck with the "Message" stuff.

src/lib/socketStore.ts
```diff
@@ -80,6 +82,9 @@ export const createSocketStore = () => {
 			console.log('[socket] message', message);
 			if (message.type === 'Create') {
 				messages.update(($messages) => [message, ...$messages]);
+			} else {
```
greatbacon, line 85: This was a hack to get past the "Message"/"Post" question. I think we should take some time to talk through it before I move forward with anymore websocket stuff.

src/lib/ChatRoom.svelte
```diff
@@ -1,20 +1,25 @@
...
 	$: console.log(`[chatRoom] fetching posts for ${space.space_id}`);
 
-	let posts: Post[] = [];
-
 	const loadPosts = async (spaceId: number) => {
 		const res = await fetch(`/api/v1/spaces/${spaceId}/posts`);
 		if (res.ok) {
 			const data = await res.json();
-			posts = data.posts;
+			posts.update(() => data.posts);
```
ryanatkn, line 22: `store.update` is useful when you need the previous value, otherwise `store.set` avoids the wrapper function, so `posts.set(data.posts)`, and in Svelte files we get store sugar, so `$posts = data.posts`.

src/lib/ChatRoom.svelte
```diff
@@ -1,20 +1,25 @@
...
 	import {browser} from '$app/env';
 	import type {Space} from '../spaces/space.js';
+	import PostList from '$lib/PostList.svelte';
+	import {posts} from '$lib/postStore';
+	import type {Json} from '@feltcoop/gro/dist/utils/json.js';
+	import type {SocketStore} from './socketStore.js';
+	import {getContext} from 'svelte';
+
+	export let socket: SocketStore;
+	socket = getContext('socket');
```
ryanatkn, line 11: The socket should probably be either a prop or in context but not both, so we can change this line to `const socket = getContext('socket');`. We'll probably make this a helper in the future, so the type annotation won't be needed, and the error-prone string won't be used, but I don't have ideal patterns there yet, so we should punt.

src/lib/ChatRoom.svelte
ryanatkn: Nice, to make this a little closer to the final form, I think it makes sense to put `broadcastPost(data)` directly before `await fetch`, and then `{content: text}` is used as its data param along with the fetch call, so it's doing the driving and the restful hack is the afterthought.

src/lib/ChatRoom.svelte
```diff
@@ -26,26 +31,36 @@
...
 		}
 		text = '';
 	};
 
+	const broadcastPost = async (data: Json) => {
+		if (!$socket.connected) {
+			console.error('expected socket to be connected to chat');
+			return;
+		}
+		if (!text) return;
```
ryanatkn, line 48: can delete

ryanatkn (approved):
amaze, I can *talk to myself* **!!!!!**

src/lib/ChatRoom.svelte
greatbacon: Ah yeah! I originally had the prop, and then got lazy and just pulled the socket in directly here. I assume long term the call for the socket context will live at the route layer & then get passed as prop all the way down the respective chains.

greatbacon:
> amaze, I can _talk to myself_ **!!!!!**

Right!? I'm excited for the point where we can talk _to each other_!!!

PR #32, opened 2021-05-18
Create spaces

greatbacon:
This adds the serverside code for creating a new space in a given community, as well as starting to flesh out the front end flow for space creation.

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -153,6 +153,26 @@ export class Database {
 				console.log('[db] community data', data);
 				return {ok: true, value: data};
 			},
+			insert: async (
+				community_id: string,
+				url: string,
+				media_type: string,
+				content: string,
+			): Promise<Result<{value: Space}>> => {
+				const data = await this.sql<Space[]>`
+					INSERT INTO spaces (url, media_type, content) VALUES (
+						${url},${media_type},${content}
+					) RETURNING *
+				`;
+				console.log('[db] created space', data);
+				const association = await this.sql`
+			   INSERT INTO community_spaces (space_id, community_id) VALUES (
+					 ${(data[0].space_id, community_id)}
+				 )
+				`;
+				console.log('[db] created community_space', association);
+				return {ok: true, value: data[0]};
+			},
 		},
 		posts: {
 			insert: async (
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -167,6 +167,7 @@ export class Database {
 				console.log('[db] created space', data);
 				const space_id: number = data[0].space_id!;
 				console.log(community_id);
+				//TODO more robust error handling or condense into single query
 				const association = await this.sql<any>`
 			   INSERT INTO community_spaces (space_id, community_id) VALUES (
 					 ${space_id},${community_id}
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -165,9 +165,11 @@ export class Database {
 					) RETURNING *
 				`;
 				console.log('[db] created space', data);
-				const association = await this.sql`
+				const space_id: number = data[0].space_id!;
+				console.log(community_id);
+				const association = await this.sql<any>`
 			   INSERT INTO community_spaces (space_id, community_id) VALUES (
-					 ${(data[0].space_id, community_id)}
+					 ${space_id},${community_id}
 				 )
 				`;
 				console.log('[db] created community_space', association);
--- a/src/lib/SpaceNav.svelte
+++ b/src/lib/SpaceNav.svelte
@@ -1,15 +1,27 @@
 <script lang="ts">
 	import type {Space} from '../spaces/space.js';
 
+	export let communityId: number;
 	export let spaces: Space[];
 	export let selectedSpace: Space;
 	export let selectSpace: (community: Space) => void;
 
 	const createSpace = async () => {
-		//Trigger component with input form
+		//TODO: Trigger component with input form
 		//Needs to collect url(i.e. name for now), type (currently default json/application), & content (hardcoded JSON struct)
-		//Finally submit as a JSON package to server
-		console.log('create');
+		const url = '/hello/world';
+		const text = {
+			url: url,
+			media_type: 'json/application',
+			content: `{"type": "ChatRoom", "props": {"data": "${url}/posts"}}`,
+		};
+		const res = await fetch(`/api/v1/communities/${communityId}/spaces`, {
+			method: 'POST',
+			headers: {'Content-Type': 'application/json'},
+			body: JSON.stringify(text),
+		});
+		const data = await res.json();
+		spaces.push(data.space);
 	};
 </script>
 
--- a/src/lib/SpaceNav.svelte
+++ b/src/lib/SpaceNav.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
 	import type {Space} from '../spaces/space.js';
+	import type {Community} from '../communities/community.js';
 
-	export let communityId: number;
+	export let community: Community;
 	export let spaces: Space[];
 	export let selectedSpace: Space;
 	export let selectSpace: (community: Space) => void;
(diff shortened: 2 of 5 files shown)


src/db/Database.ts
```diff
@@ -153,6 +153,28 @@ export class Database {
...
+				const data = await this.sql<Space[]>`
+					INSERT INTO spaces (url, media_type, content) VALUES (
+						${url},${media_type},${content}
+					) RETURNING *
+				`;
+				console.log('[db] created space', data);
+				const space_id: number = data[0].space_id!;
+				console.log(community_id);
+				const association = await this.sql<any>`
+			   INSERT INTO community_spaces (space_id, community_id) VALUES (
```
greatbacon, line 171: This needs some more robust error handling long term or if it could be combined into a single query, even better.

src/lib/SpaceNav.svelte
```diff
@@ -1,25 +1,47 @@
...
+			media_type: 'json/application',
+			content: `{"type": "ChatRoom", "props": {"data": "${url}/posts"}}`,
+		};
+		const res = await fetch(`/api/v1/communities/${communityId}/spaces`, {
+			method: 'POST',
+			headers: {'Content-Type': 'application/json'},
+			body: JSON.stringify(text),
+		});
+		const data = await res.json();
+		spaces.push(data.space);
```
greatbacon, line 24: I tried to get some reactive behavior off of this, but my Svelte-fu is failing me here. Any thoughts?

src/lib/SpaceNav.svelte
```diff
@@ -1,25 +1,47 @@
...
+	};
 </script>
 
 <div class="sidenav">
+	<div class="header">
+		<button type="button" class="button-emoji" on:click={() => createSpace()}>➕</button>|
```
greatbacon, line 30: I'm not sure if this was the "right" way to get this behavior (Emoji acting as a button while still just looking like an emoji) but hey, it works!

ryanatkn: Yep this is on the right track. The usage of `<button>` is the key here, for accessibility and semantic markup reasons. One addition I know of its to improve accessibility, we want the [`aria-label` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) so e.g. screenreaders can give useful information to people - buttons with a readable label don't need one.

My accessibility knowledge isn't great, so I'll keep trying to improve here as we go, rather than saving it for "later".

fyi the `type="button"` is a tricky one - you might see some code I write always use it, but it's for a specific contextual reason that usually doesn't apply: if you have a normal `<button>` inside a `<form>`, it gets `type="submit"` by default, which *reloads the current webpage in the browser*. This can lead to really surprising behavior, but TBH, we should probably get in the habit of using `type="button"` only when needed, or in generic components that could be nested inside `<form>`.

src/lib/SpaceNav.svelte
ryanatkn: This is one of the things that every Svelte dev learns. Svelte has no "magic methods" for its reactivity - instead, it uses assignment to detect when to tell a component to update. 

So to Svelte, `spaces.push(data.space)` just looks like any other line of code. For reactivity, you'll need a second line, `spaces = spaces`, or to use immutability, change the line to `spaces = spaces.concat(data.space)`.

This ties into a bigger conversation about Svelte's immutability flag, which Gro has been using by default. (but we do not currently configure this for SvelteKit; we should soon, to all templates+projects) By default, both versions above work. With the immutability flag on, Svelte is able to optimize things by using reference equality to compare objects. What this means is that it can help components avoid a lot of wasted work, sometimes enough to be significant for end users.

So I think it's good practice to use the immutability flag. But there are situations where we'll want to opt out, because immutability has its own tradeoffs, and isn't ideal for every situation.

Once we have the flag enabled on all projects, to *opt out* of immutability, components can add this:

`<svelte:options immutable={false}>`

For now I'd probably just use the concat form, so we prefer immutable patterns by default.

src/lib/SpaceNav.svelte
```diff
@@ -1,25 +1,47 @@
...
 	export let spaces: Space[];
 	export let selectedSpace: Space;
 	export let selectSpace: (community: Space) => void;
+
+	const createSpace = async () => {
+		//TODO: Trigger component with input form
+		//Needs to collect url(i.e. name for now), type (currently default json/application), & content (hardcoded JSON struct)
+		const url = '/hello/world';
+		const text = {
+			url: url,
```
ryanatkn, line 14: can shorthand this to `url,`

src/db/Database.ts
ryanatkn: Cool ok, maybe leave a comment for future cleanup? I've got nothing useful to add here.

ryanatkn (approved):
Awesome, I'd say a good followup PR would be adding an input box next to the add space button to name it, or just using the `window.prompt` function as a quick hack.

src/lib/SpaceNav.svelte
```diff
@@ -1,25 +1,47 @@
...
+	export let communityId: number;
 	export let spaces: Space[];
 	export let selectedSpace: Space;
 	export let selectSpace: (community: Space) => void;
+
+	const createSpace = async () => {
+		//TODO: Trigger component with input form
+		//Needs to collect url(i.e. name for now), type (currently default json/application), & content (hardcoded JSON struct)
+		const url = '/hello/world';
+		const text = {
```
ryanatkn, line 13: I'd probably name this `doc`.

src/lib/SpaceNav.svelte
```diff
@@ -1,25 +1,47 @@
...
 </script>
 
 <div class="sidenav">
+	<div class="header">
+		<button type="button" class="button-emoji" on:click={() => createSpace()}>➕</button>|
+		<button type="button" class="button-emoji" on:click={() => console.log('search')}>🔍</button>
+	</div>
 	{#each spaces as space (space.space_id)}
-		<button type="button" on:click={() => selectSpace(space)}>{space.url}</button>
+		<button type="button" class="button-nav" on:click={() => selectSpace(space)}>{space.url}</button
```
ryanatkn, line 34: Was going back and forth on this, I think as a general rule we should avoid leaving unused classes like `button-nav` around. Our tools won't check for this automatically (though that's something I have working in another project), Svelte only warns if it sees a class in the `<style>` that's unused in the markup.

src/lib/Workspace.svelte
```diff
@@ -22,7 +22,12 @@
 		<CommunityNav {communities} {selectedCommunity} {selectCommunity} />
 	</section>
 	<section class="spacenav">
-		<SpaceNav spaces={selectedCommunity.spaces} {selectedSpace} {selectSpace} />
+		<SpaceNav
+			communityId={selectedCommunity.community_id}
```
ryanatkn, line 26: I think we can assume we'll always have the community object loaded when we mount a `SpaceNav`, so I think passing `community` instead of `communityId` down as props is better. Mainly I'm thinking any future grandchildren might want the full object, and looking it up on demand would be wasteful.

src/lib/SpaceNav.svelte
greatbacon: I'll drop a TODO to put an IconButton component in Felt & get these converted over once the Felt deploy is all squared away.

Thanks for the button theory talk too! Good call on the accessibility concern.

As far as always vs not-always using the `type="button"` is there a downside to being explicit about it everywhere?

src/lib/SpaceNav.svelte
greatbacon: Yeah, I think that's a good general principal. I left it there as a stub, I'll add some styling to it though since I think the current unstyled buttons aren't a _great_ look.

greatbacon:
> Awesome, I'd say a good followup PR would be adding an input box next to the add space button to name it (or its `url`, or both). If you want to avoid UI concerns you could just use `window.prompt` - the real work is in the data structures.

I'm already working on a follow up PR that will trigger a modal with a space for adding the name, which'll just go in the URL space for now

src/lib/SpaceNav.svelte
ryanatkn: Basically no downside except small amounts of noise and confusion to the reader, it's more that Svelte is so terse and convenient, and buttons are so common, that it aged poorly to my eyes.

`<form>` elements are rare enough and testing is not some corner case you hit, so in hindsight my putting `type="button"` everywhere was an overreaction, like I was protesting an API choice I don't like and never want to think about again. :shrug:

src/lib/SpaceNav.svelte
ryanatkn: Sounds good, I like that, gives us something to look at and think about extraction. Lots of styling stuff for us to figure out, so this is like raw material for thinking.

PR #34, opened 2021-05-21
Create spaces modal

greatbacon:
This PR adds logic to bring up a modal when going to create a new space, and ties that back to the creation logic upon hitting 'enter'

### DIFF
--- a/src/lib/Modal.svelte
+++ b/src/lib/Modal.svelte
@@ -1,40 +1,26 @@
 <script lang="ts">
-	let isOpen = false;
-	function open() {
-		isOpen = true;
-	}
-	function close() {
-		isOpen = false;
-	}
+	let opened = false;
+	const open = () => {
+		opened = true;
+	};
+	const close = () => {
+		opened = false;
+	};
 </script>
 
-<slot name="trigger" {open}>
-	<!-- fallback trigger to open the modal -->
-	<button on:click={open}>Open</button>
+<slot name="trigger" {open} {close}>
+	<button on:click={open}>open</button>
 </slot>
 
-{#if isOpen}
+{#if opened}
 	<div class="modal">
 		<div class="backdrop" on:click={close} />
-
-		<div class="content-wrapper">
-			<slot name="header">
-				<!-- fallback -->
-				<div>
-					<h1>Your Modal Heading Goes Here...<button on:click={close}>close</button></h1>
-				</div>
-			</slot>
-
+		<div class="container">
+			<slot name="header" />
 			<div class="content">
 				<slot name="content" />
 			</div>
-
-			<slot name="footer" {close}>
-				<!-- fallback -->
-				<div>
-					<h1>Your Modal Footer Goes Here...</h1>
-				</div>
-			</slot>
+			<slot name="footer" />
 		</div>
 	</div>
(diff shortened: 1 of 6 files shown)


src/lib/Modal.svelte
```diff
@@ -0,0 +1,78 @@
+<script lang="ts">
```
greatbacon, line 1: This Modal was more or less directly lifted from a Svelte tutorial, so please feel free to tear apart/comment/rebuild

src/lib/SpaceNav.svelte
```diff
@@ -23,18 +34,34 @@
...
+					on:click={() => trigger()}>➕</button
+				>
+			</span>
+			<div slot="header">
+				<h1>Create a new space</h1>
+			</div>
+
+			<div slot="content">
+				<p>
+					<input type="text" placeholder="> chat" on:keydown={onKeyDown} bind:value={newName} />
```
greatbacon, line 58: So there's still a piece of the puzzle I'm missing, which is how to pass the `{exit}` function up there on Line 43 (which will close the modal) into the `[onKeyDown}` function so that it can be called after the space is successfully created. Also let me know if this is totally off base.

ryanatkn: This looks good, a simple version could be:

`on:keydown={(e) => onKeyDown(e, exit)}`

I think behaves the same:

`on:keydown={toOnKeyDown(exit)}`

in the second case the new function signature would be:

`const onKeyDown = (exit: () => void) => async (e: KeyboardEvent) => {`

src/lib/Modal.svelte
ryanatkn: Looks like a fine start, similar to things I've written and adds enough features to get me thinking without much code.

src/lib/Modal.svelte
```diff
@@ -0,0 +1,62 @@
...
+	<button on:click={open}>open</button>
+</slot>
+
+{#if opened}
+	<div class="modal">
+		<div class="backdrop" on:click={close} />
+		<div class="container">
+			<slot name="header" />
+			<div class="content">
+				<slot name="content" {open} {close} />
```
greatbacon, line 21: @ryanatkn I got your stuff merged in, I had to add these to this slot to get the close on submit behavior working properly tho.  Just wanted to run it past ya and get some clarity on it, otherwise we're g2g!

ryanatkn: I was getting confused here by the slot shorthand, so apparently these are identical:

```svelte
		<Modal>
			<span slot="trigger" let:open={openModal}>
```


```svelte
		<Modal let:open={openModal}>
			<span slot="trigger">
```

Our `Modal` component is providing an API with `openModal` and `closeModal`, so we should make sure that's available on each defined slot:

```svelte
			<slot name="header" {open} {close} />
			<div class="content">
				<slot name="content" {open} {close} />
			</div>
			<slot name="footer" {open} {close} />
```

Then the consumer code has a choice of how it binds to these functions: see the 2 snippets above. I guess the shorthand with `let:` on the component itself looks nice? I wouldn't expect any downsides from this.

ryanatkn (approved):

PR #36, opened 2021-05-25
Add create community flow

greatbacon:
This PR adds the frontend and backend components for creating new communities on a felt-server instance

### DIFF
--- a/src/communities/communityMiddleware.ts
+++ b/src/communities/communityMiddleware.ts
@@ -27,6 +27,12 @@ export const toCommunitiesMiddleware = (server: ApiServer): Middleware => {
 export const toCommunityMiddleware = (server: ApiServer): Middleware => {
 	const {db} = server;
 	return async (req, res) => {
+		if (!req.accountSession) {
+			//TODO centralize error message strings
+			console.log('[communityMiddleware] no account to search for communities');
+			return send(res, 401, {reason: 'not logged in'});
+		}
+
 		console.log('[communityMiddleware] account', req.accountSession?.account.account_id!); // TODO logging
 		console.log('[communityMiddleware] community', req.params.community_id);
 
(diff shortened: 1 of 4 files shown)


src/communities/communityMiddleware.ts
```diff
@@ -40,3 +46,22 @@ export const toCommunityMiddleware = (server: ApiServer): Middleware => {
...
+//Creates a new community for an instance
+export const toCreateCommunityMiddleware = (server: ApiServer): Middleware => {
+	const {db} = server;
+	return async (req, res) => {
+		console.log('[communityMiddleware] creating community', req.body);
+
+		const createCommunityResult = await db.repos.communities.insert(
+			req.body.name,
+			req.accountSession?.account.account_id!,
+		);
```
ryanatkn, line 59: Instead of guarding around the `accountSession` with the `?` and `!`, let's do the same `if (!req.accountSession) {` auth check as you added to `toCommunityMiddleware` above.

greatbacon: Oops, that what happens when I forget to scroll while copy-pastaing 🙃

src/db/Database.ts
```diff
@@ -126,6 +126,26 @@ export class Database {
...
 			},
+			insert: async (name: string, account_id: number): Promise<Result<{value: Community}>> => {
+				const data = await this.sql<Community[]>`
+					INSERT INTO communities (name) VALUES (
+						${name}
+					) RETURNING *
+				`;
+				console.log('[db] created community', data);
+				//TODO is there a way to do this via sql? Or do we just add a default `/general` space to all new communities?
+				data[0].spaces = [];
```
ryanatkn, line 137: Interesting q, no opinions here, I like both choices

src/lib/CommunityNav.svelte
```diff
@@ -1,19 +1,91 @@
...
+	let newName = '';
+
+	const onKeyDown = async (e: KeyboardEvent, closeModal: () => void) => {
+		if (e.key === 'Enter') {
+			await createCommunity();
+			closeModal();
+		}
+	};
+
+	const createCommunity = async () => {
```
ryanatkn, line 18: To make this more amenable to refactoring, and to make it clearer, I'd parameterize this by `newName` instead of relying on the closure. In the context of the `createCommunity` function, I'd name the arg `name`.

ryanatkn (approved):

src/communities/communityMiddleware.ts
ryanatkn: There's so much boilerplate in our server routes it's hard to see the forest from the trees!

src/lib/CommunityNav.svelte
greatbacon: Good call

PR #37, opened 2021-05-27
Community invite

greatbacon:
This PR adds a first pass to inviting people into communities

NB: This relies on some less specified terms/relations (`friend`, `member`) & the system itself is currently non-consentful (I.e. no chance to reject an invite into a community).

### DIFF
--- a/src/communities/community.ts
+++ b/src/communities/community.ts
@@ -1,7 +1,9 @@
 import type {Space} from '../spaces/space.js';
+import type {Member} from '../members/member.js';
 
 export interface Community {
 	community_id?: number;
 	name: string;
 	spaces: Space[];
+	members: Member[];
 }
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -77,12 +77,15 @@ export class Database {
 			},
 		},
 		members: {
+			//TODO: this is a hack to stub out "friends" for inviting to a Community.
+			//This should use a community_id to filter or something
 			getAll: async (): Promise<Result<{value: Member[]}, {reason: string}>> => {
 				const data = await this.sql<Member[]>`
 				select account_id, name from accounts
 				`;
 				return {ok: true, value: data};
 			},
+			//TODO: refactor this code to return 'Member'
 			create: async (account_id: number, community_id: string): Promise<Result<{value: any}>> => {
 				const data = await this.sql`
 				INSERT INTO account_communities (account_id, community_id) VALUES (
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -83,6 +83,15 @@ export class Database {
 				`;
 				return {ok: true, value: data};
 			},
+			create: async (account_id: number, community_id: string): Promise<Result<{value: any}>> => {
+				const data = await this.sql`
+				INSERT INTO account_communities (account_id, community_id) VALUES (
+					${account_id},${community_id}
+				) RETURNING *			
+				`;
+				console.log('[db] created account_communities', data);
+				return {ok: true, value: data};
+			},
 		},
 		entities: {
 			findByAccount: async (
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -5,6 +5,7 @@ import type {AccountSession} from '../session/clientSession.js';
 import type {Community} from '../communities/community.js';
 import type {Space} from '../spaces/space.js';
 import type {Post} from '../posts/post.js';
+import type {Member} from '../members/member.js';
 import type {Account} from '../vocab/account/account.js';
 import type {Entity} from '../vocab/entity/entity.js';
 import type {PostgresSql} from './postgres.js';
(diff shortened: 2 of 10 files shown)


src/communities/community.ts
```diff
@@ -1,7 +1,9 @@
 import type {Space} from '../spaces/space.js';
+import type {Member} from '../members/member.js';
 
 export interface Community {
 	community_id?: number;
 	name: string;
 	spaces: Space[];
+	members: Member[];
```
ryanatkn, line 8: Nice I like this, clear place to compose a `Persona` in each `Member`.

src/db/Database.ts
```diff
@@ -37,10 +38,11 @@ export class Database {
 				let communities: Community[] = unwrap(
 					await this.repos.communities.filterByAccount(account.account_id!),
 				);
+				let friends: Member[] = unwrap(await this.repos.members.getAll());
```
ryanatkn, line 41: Nothing to change here, just thinking out loud:

If a `Member` encapsulates a `Persona`'s relationship to one `Community`, I'm wondering what the eventual abstraction for friends will be. If we reuse the `Community` concept like we briefly talked about, so each user gets their own personal `Community` that contextually has its own `Members` to model other people, this seems right on.

src/db/Database.ts
```diff
@@ -74,6 +76,23 @@ export class Database {
 				return {ok: false, type: 'noAccountFound', reason: `No account found with name: ${name}`};
 			},
 		},
+		members: {
+			getAll: async (): Promise<Result<{value: Member[]}, {reason: string}>> => {
```
ryanatkn, line 80: Don't need to change this now, I'm guessing this should take the community id as an arg to scope it.

src/db/Database.ts
```diff
@@ -74,6 +76,23 @@ export class Database {
...
+				const data = await this.sql<Member[]>`
+				select account_id, name from accounts
+				`;
+				return {ok: true, value: data};
+			},
+			create: async (account_id: number, community_id: string): Promise<Result<{value: any}>> => {
+				const data = await this.sql`
```
ryanatkn, line 87: I think `{value: Member}` and `this.sql<Member>` are the correct types?

src/lib/CommunityNav.svelte
```diff
@@ -1,11 +1,17 @@
...
 	import type {Community} from '../communities/community.js';
 	import Modal from '$lib/Modal.svelte';
+	import type {Member} from 'src/members/';
 
+	export let friends: Member[];
 	export let communities: Community[];
 	export let selectedCommunity: Community;
 	export let selectCommunity: (community: Community) => void;
 
+	let invitableFriends = friends.filter(
```
ryanatkn, line 11: To make this reactive, change to `$: invitableFriends`.

src/routes/index.svelte
```diff
@@ -4,12 +4,15 @@
...
 	import type {ClientAccount} from 'src/session/clientSession.js';
 	import type {Community} from 'src/communities/community.js';
+	import type {Member} from 'src/members/member.js';
 
 	const title = 'felt-server';
 	let account: ClientAccount;
 	$: account = $session?.account;
 	let communities: Community[] = [];
 	$: communities = $session?.communities;
+	let friends: Member[] = [];
```
ryanatkn, line 14: Here and 2 lines above, the `= []` initialization should be removed, and ideally we don't need these declarations.

There's one tricky caveat in Svelte with TypeScript right now, where we can't use the non-null assertion operator `!`, which tells TypeScript, "ignore null and undefined values here, just trust me".

Svelte template code cannot use any TypeScript right now, so we can't sprinkle `!` in there, so if you get template errors in this case after removing the declaration line, this should fix it: (but don't add the `!` unless it fixes an error)

`$: friends = $session?.friends!;`

ryanatkn (approved):
very cool, very social

src/lib/CommunityNav.svelte
```diff
@@ -1,11 +1,17 @@
...
+		(x) => !selectedCommunity.members.some((y) => x.account_id == y.account_id),
```
ryanatkn, line 12: This is good code, not suggesting to change anything. Given that it's an `O=Nsquared` algorithm that often suggests improvements to data structures or access patterns upstream. Just something to think about when we iterate on the design.

src/db/Database.ts
greatbacon: I think `Friend` as a concept actually transcends the `Community` model, and instead represents a direct relationship between two `Persona`s

src/db/Database.ts
greatbacon: I think I have a comment on whatever calls this method, but I'll go ahead and put one here too.

src/db/Database.ts
greatbacon: Indeed you are correct!

src/lib/CommunityNav.svelte
greatbacon: Yeah, I'm not sure what the process looks like per se, but my gut says that a lot of session things that are arrays in the front end data store should probably be hashes with the `object_id` as key, or something to that effect.

PR #38, opened 2021-05-31
Added some front end logic to fix state errors with new accounts

greatbacon:
This PR fixes an issue where new accounts were bugging out due to missing communities & spaces. This adds some more front end validation to help render the screen for when accounts have no community, or when a community has no spaces.

I also tweaked the styling to make the community nav stand out, & put the chat input box at the bottom of it's component.

### DIFF
--- a/src/lib/Workspace.svelte
+++ b/src/lib/Workspace.svelte
@@ -8,20 +8,16 @@
 
 	export let friends: Member[];
 	export let communities: Community[];
-	let selectedCommunity = communities ? communities[0] : {};
-	let selectedCommunitySpaces = selectedCommunity ? selectedCommunity.spaces : [];
+	let selectedCommunity = communities[0] || null;
+	$: selectedCommunitySpaces = selectedCommunity?.spaces || null;
 	const selectCommunity = (community: Community) => {
 		selectedCommunity = community;
-		selectedCommunitySpaces = community.spaces;
 	};
 	let selectedSpace = selectedCommunity ? selectedCommunity.spaces[0] : null;
 	const selectSpace = (space: Space) => {
 		selectedSpace = space;
 		console.log(`[ss] ${selectedSpace.url}`);
 	};
-	if (selectedSpace) {
-		console.log('selectedSpace', selectedSpace);
-	}
 </script>
 
 <div class="workspace">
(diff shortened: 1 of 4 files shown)


src/lib/Workspace.svelte
```diff
@@ -8,31 +8,40 @@
 
 	export let friends: Member[];
 	export let communities: Community[];
-	let selectedCommunity = communities[0];
+	let selectedCommunity = communities ? communities[0] : {};
```
ryanatkn, line 11: should this be:

`let selectedCommunity = communities ? communities[0] : {};`

?

src/lib/Workspace.svelte
```diff
@@ -8,31 +8,40 @@
...
+	let selectedCommunitySpaces = selectedCommunity ? selectedCommunity.spaces : [];
```
ryanatkn, line 12: `let selectedCommunitySpaces = selectedCommunity ? selectedCommunity.spaces : null;`

I think

src/lib/Workspace.svelte
```diff
@@ -8,31 +8,40 @@
...
 	const selectCommunity = (community: Community) => {
 		selectedCommunity = community;
+		selectedCommunitySpaces = community.spaces;
```
ryanatkn, line 15: remove, can be reactive

src/lib/Workspace.svelte
```diff
@@ -8,31 +8,40 @@
...
 	};
-	let selectedSpace = selectedCommunity.spaces[0];
+	let selectedSpace = selectedCommunity ? selectedCommunity.spaces[0] : null;
 	const selectSpace = (space: Space) => {
 		selectedSpace = space;
 		console.log(`[ss] ${selectedSpace.url}`);
 	};
+	if (selectedSpace) {
+		console.log('selectedSpace', selectedSpace);
+	}
```
ryanatkn, line 24: could change these three lines to:

`$: console.log('selectedSpace', selectedSpace)`

to get reactive logging

ryanatkn (approved):

src/lib/Workspace.svelte
greatbacon: I wasn't sure what was better idiomatically so I went with empty object. If we're in the clear to use `null` here, I'll happily use null here lol

ryanatkn: Yeah gotcha, in this case prefer `null`, empty object is almost never something we'd do in JS -- empty arrays are common, but in this case we want to use `null` meaningfully.

PR #41, opened 2021-06-14
Upgrade sveltekit

greatbacon:
This PR updates sveltekit & the node-adaptor we use for building. It's currently still somewhat a work in progress.

Currently the big outstanding issue is using the `cookieSession` to actually populate the session object we return via the getSession hook.


.gitignore
```diff
@@ -10,3 +10,4 @@ node_modules
 *.ignore.*
 ignore/
 *.pem
+deploy.tar
```
ryanatkn, line 13: The `/functions` line can be removed above, and `/.svelte` needs to be changed to `/.svelte-kit`. (I'm trying to find the "add suggestions" feature)

.svelte-kit/dev/components/layout.svelte
```diff
@@ -0,0 +1 @@
+<slot></slot>
```
ryanatkn, line 1: this and other files will disappear when the gitignore is changed above

package.json
```diff
@@ -22,27 +22,29 @@
...
-    "prettier-plugin-svelte": "^2.2.0",
+    "prettier": "^2.3.1",
+    "prettier-plugin-svelte": "^2.3.0",
     "svelte": "^3.29.0",
     "svelte-preprocess-esbuild": "^2.0.0",
-    "tslib": "^2.0.0",
-    "typescript": "^4.2.4",
+    "tslib": "^2.3.0",
+    "typescript": "^4.3.2",
     "vite": "^2.1.5"
```
ryanatkn, line 35: SvelteKit now iincludes Vite internally, so `npm uninstall vite`

src/hooks.ts
```diff
@@ -80,39 +38,3 @@ export const handle: Handle = async ({request, render}) => {
 	// 	},
 	// };
 };
-
-// TODO move this or better yet, get rid of this completely and don't make http calls in here.
-// see the discussion at the top of this file
-const fetchJson = async (url: string, options: http.RequestOptions): Promise<any> =>
```
ryanatkn, line 86: let's pretend this never happened

package.json
```diff
@@ -22,27 +22,29 @@
...
+    "@types/ws": "^7.4.4",
     "body-parser": "^1.19.0",
+    "bufferutil": "^4.0.1",
     "cookie-session": "^1.4.0",
-    "polka": "^1.0.0-next.14",
+    "polka": "^1.0.0-next.15",
     "postgres": "^1.0.2",
-    "sirv": "^1.0.11",
+    "sirv": "^1.0.12",
+    "utf-8-validate": "^5.0.2",
```
ryanatkn, line 47: Did you install `utf-8-validate` and `bufferutil` to fix errors? I don't see the imports so that's my best guess. I had to do this with `locate-character` in Gro, but I think it was just a temporary problem with Svelte types that I was solving. (reminds me..I'll see if I can delete that soon)

ryanatkn (changes requested):
Great start, I'll see if I can dig up some links for cookie stuff. I've seen some examples that entail replacing our current `cookie-session` middleware with a cookie library and a bit more code.

package.json
greatbacon: Yeah, they're marked as optional dependencies of `ws` (i think...) and it was blocking shit from starting at run time.

ryanatkn: Gotcha ok, looks like `bufferutil` is a major performance increase to `ws`, so it's good, just weird that they're not optional!

ryanatkn (approved):
The gitignore change isn't retroactive to committed files, so I think you can delete the `.svelte-kit` directory, commit the changes, and then this PR should look good.

greatbacon:
> The gitignore change isn't retroactive to committed files, so I think you can delete the `.svelte-kit` directory, commit the changes, and then this PR should look good.

I couldn't remember so I left it alone, thanks for pointing that out lol!

PR #42, opened 2021-06-16
Fix cookie session

greatbacon:
Currently just a draft.

I thought I had found a way to parse the cookie from outside the middleware/ApiServer context, but I've run into a weird bug. 

As is, the server fails to load at runtime citing the following error:
```
ReferenceError: require is not defined
    at /node_modules/cookie-session/index.js:15:13
    at instantiateModule (/home/hreed/git/felt-server/node_modules/vite/dist/node/chunks/dep-bc228bbb.js:68693:166)
```

Presumably because of the work I'm doing in the `hooks.ts` file.

But here comes the weird bit. If I think fudge the `cookie-session` import in the `package.json`, everything works!? Server loads, name gets pulled out of cookie, everything is grand. Of course then an npm i fails.

@ryanatkn do you have any insight into what might be going on & any potentially remedies or paths I should be stepping back onto?

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -1,6 +1,7 @@
-import type {GetSession, Handle} from '@sveltejs/kit';
+import type {GetSession} from '@sveltejs/kit';
 
 import type {ClientSession} from './session/clientSession.js';
+import {apiServer} from './server/server.js';
 import type {CookieSessionRequest} from 'cookie-session';
 
 import tryGetSession from 'cookie-session';
@@ -17,7 +18,6 @@ const dev = process.env.NODE_ENV !== 'production';
 const TODO_SERVER_COOKIE_KEYS = ['TODO', 'KEY_2_TODO', 'KEY_3_TODO'];
 
 export const getSession: GetSession<CookieSessionRequest, ClientSession> = (req) => {
-	//console.log('[getSession] authenticated:', parse(request.headers.cookie));
 	tryGetSession({
 		keys: TODO_SERVER_COOKIE_KEYS,
 		maxAge: 1000 * 60 * 60 * 24 * 7 * 6, // 6 weeks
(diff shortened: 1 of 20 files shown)


ryanatkn:
This appears to be an open issue with Vite: sveltejs/kit ... issues/1570

There might be a better workaround (maybe with the `noExternals` - I'm not following what you mean about fudging `package.json` but it makes me thing it's linked, and another workaround may be as simple as manually adding some package names to the existing `noExternals` config in `svelte.config.js`.

I'm opening a PR with the workaround I found -- feel free to close if you like something else.

src/hooks.ts
```diff
@@ -1,7 +1,10 @@
-import type {GetSession, Handle} from '@sveltejs/kit';
+import type {GetSession} from '@sveltejs/kit';
 
 import type {ClientSession} from './session/clientSession.js';
-import type {ClientContext} from './server/ApiServer.js';
+import {apiServer} from './server/server.js';
```
greatbacon, line 4: @ryanatkn so I'm trying to pull in this export to access the DB param on the server for loading session data, but I keep getting the following error
```
Could not resolve './server/server.js' from src/hooks.ts
> Could not resolve './server/server.js' from src/hooks.ts
```
am I missing something? Am I even on the right track for this?

ryanatkn: Try removing the `.js` extension -- if that doesn't work (it might cause a new error) there's some other things we can try. This is an incompatibility between Vite and TypeScript that we currently just need to work around. Vite wants us to use no extension or `.ts`, but TypeScript errors on `.ts` imports deliberately -- I wrote up more in a PR on Gro but that's the gist, I don't know of a good solution other than to just omit the `.js`. It works for `import type` because they get stripped, and TypeScript can handle it.

ESM requires a file extension, so I don't like the workaround long term, but it's fine for now.

src/hooks.ts
```diff
@@ -10,32 +14,22 @@ import type {ClientContext} from './server/ApiServer.js';
...
-	// 			friends: context.friends,
-	// 			entities: [], // TODO load
-	// 	  }
-	// 	: {guest: true}; // TODO is swallowing `context.error`, only return in dev mode? look for "reason"?
-	return {guest: true}; //Hack until cookies are working again...
-};
+//TODO source this from wherever ApiServer.js does
+const dev = process.env.NODE_ENV !== 'production';
+const TODO_SERVER_COOKIE_KEYS = ['TODO', 'KEY_2_TODO', 'KEY_3_TODO'];
+let db = new Database({sql: postgres(defaultPostgresOptions)})
```
greatbacon, line 20: @ryanatkn okay, I'm trying a slightly different tack. I started removing file extensions and realized it was going to want me to do that for the whole import chain. So I thought (at least to get it working short term) to just have a db connection load here. But now vite is throwing a handful of errors like the following:
```
1:21:19 PM [vite] new dependencies found: @feltcoop/gro/dist/utils/env.js, @feltcoop/gro, updating...
 > node_modules/@feltcoop/gro/dist/utils/path.js:1:9: error: No matching export in "browser-external:path" for import "extname"
    1 │ import { extname, basename, dirname } from 'path';
    ```
Is this because the gro version specified in the lock file is behind? More vite chaos for imports?

ryanatkn: That seems likely, probably should upgrade both Gro and Felt, so `npm i @feltcoop/gro@latest @feltcoop/felt@latest`. I'll be able to help with testing tomorrow.

ryanatkn:
Working on these two PRs, and will continue on this when they're ready: 

- [ ] publish Felt with fixes: https://github.com/feltcoop/felt/pull/63
- [ ] update API server adapter: https://github.com/feltcoop/gro/pull/223

ryanatkn:
Awesome, looks good to me. We'll have to make adjustments for the production build (no clue what the final structure looks like) but this is wonderful -- we can do things with code now! :shipit:

ryanatkn (approved):

PR #47, opened 2021-06-29
update felt and gro

ryanatkn:
Upgrades Felt and Gro for the `felt-server` usecases, needed for #42 to proceed.


greatbacon (approved):
:shipit:

PR #59, opened 2021-07-13
Add logic for rendering names in chat

greatbacon:
Quick fix to get account names rendering for posts

### DIFF
--- a/src/lib/ui/Post_List.svelte
+++ b/src/lib/ui/Post_List.svelte
@@ -9,7 +9,7 @@
 <ul>
 	{#each posts as post (post.post_id)}
 		<li>
-			<small>{members.find((member) => member.account_id == post.actor_id)?.name}</small>:
+			<small>{members.find((member) => member.account_id === post.actor_id)?.name}</small>:
 			{post.content}
 		</li>
 	{/each}
--- a/src/lib/ui/Post_List.svelte
+++ b/src/lib/ui/Post_List.svelte
@@ -9,6 +9,7 @@
 <ul>
 	{#each posts as post (post.post_id)}
 		<li>
+			<!--TODO: fix the backend datastruct to not need this-->
 			<small>{members.find((member) => member.account_id === post.actor_id)?.name}</small>:
 			{post.content}
 		</li>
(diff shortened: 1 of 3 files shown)


src/lib/ui/Post_List.svelte
```diff
@@ -1,13 +1,15 @@
...
 
 	export let posts: Post[];
+	export let members: Member[];
 </script>
 
 <ul>
 	{#each posts as post (post.post_id)}
 		<li>
-			<small>{post.actor_id}</small>:
+			<small>{members.find((member) => member.account_id == post.actor_id)?.name}</small>:
```
ryanatkn, line 12:
```suggestion
			<small>{members.find((member) => member.account_id === post.actor_id)?.name}</small>:
```

src/lib/ui/Workspace.svelte
```diff
@@ -9,6 +9,8 @@
 	export let friends: Member[];
 	export let communities: Community[];
 	let selected_community = communities[0] || null;
+	$: selected_community_members = selected_community.members;
```
ryanatkn, line 12: This will error if `selected_community` is null, which can happen if there are no loaded communities. (e.g. a new account). Copying the pattern below:

```suggestion
	$: selected_community_members = selected_community?.members || null;
```

ryanatkn (approved):
Looks good, there's some cleanup to the interfaces needed here but I'll make a pass another time, touches too many lines of code.

src/lib/ui/Workspace.svelte
ryanatkn, line 12:
```suggestion
```

Isn't being used -- breaks patterns with other things but the fix is more complex than needed for this PR.

src/lib/ui/Post_List.svelte
ryanatkn: Might also want to add `?.name || 'unknown'` or some default, currently will just be undefined if it can't find the member. If you want to assume the member exists, change it to `!.name`.

greatbacon: Yeah, this is just a client side hack to get names displayed. I agree that this is something to fix on the backend.

PR #63, opened 2021-07-19
Added actor name to Post query

greatbacon:
This takes the iniefficient front end hack to get names rendering and actually implements it properly as a database query

### DIFF
--- a/src/lib/db/Database.ts
+++ b/src/lib/db/Database.ts
@@ -222,7 +222,7 @@ export class Database {
 			filter_by_space: async (space_id: string): Promise<Result<{value: Post[]}>> => {
 				console.log(`[db] preparing to query for space posts: ${space_id}`);
 				const data = await this.sql<Post[]>`
-					SELECT p.post_id, p.content, p.actor_id, p.space_id FROM posts p WHERE p.space_id= ${space_id}
+				SELECT p.post_id, p.content, p.actor_id, p.space_id, a.name FROM posts p JOIN accounts a ON p.actor_id=a.account_id WHERE p.space_id=${space_id}
 				`;
 				console.log('[db] space posts', data);
 				return {ok: true, value: data};
--- a/src/lib/db/Database.ts
+++ b/src/lib/db/Database.ts
@@ -222,7 +222,8 @@ export class Database {
 			filter_by_space: async (space_id: string): Promise<Result<{value: Post[]}>> => {
 				console.log(`[db] preparing to query for space posts: ${space_id}`);
 				const data = await this.sql<Post[]>`
-				SELECT p.post_id, p.content, p.actor_id, p.space_id, a.name FROM posts p JOIN accounts a ON p.actor_id=a.account_id WHERE p.space_id=${space_id}
+					SELECT p.post_id, p.content, p.actor_id, p.space_id, a.name FROM posts p 
+					JOIN accounts a ON p.actor_id=a.account_id WHERE p.space_id=${space_id}
 				`;
 				console.log('[db] space posts', data);
 				return {ok: true, value: data};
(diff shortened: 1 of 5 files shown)


src/lib/ui/Post_List.svelte
```diff
@@ -1,17 +1,21 @@
...
-	import type {Member} from '$lib/members/member.js';
 
 	export let posts: Post[];
-	export let members: Member[];
 </script>
 
 <ul>
 	{#each posts as post (post.post_id)}
 		<li>
 			<!--TODO: fix the backend datastruct to not need this-->
```
ryanatkn, line 10:
```suggestion

```

ryanatkn:
Nice, what do you think of `author_name` or `actor_name` instead of just `name`?

ryanatkn (approved):

src/lib/db/Database.ts
```diff
@@ -222,7 +222,7 @@ export class Database {
 			filter_by_space: async (space_id: string): Promise<Result<{value: Post[]}>> => {
 				console.log(`[db] preparing to query for space posts: ${space_id}`);
 				const data = await this.sql<Post[]>`
-					SELECT p.post_id, p.content, p.actor_id, p.space_id FROM posts p WHERE p.space_id= ${space_id}
+				SELECT p.post_id, p.content, p.actor_id, p.space_id, a.name FROM posts p JOIN accounts a ON p.actor_id=a.account_id WHERE p.space_id=${space_id}
```
ryanatkn, line 225: This is the indentation I mentioned, I find it easier to read when the template strings are indented, but I'm open to either.

ryanatkn:
For the CI error, we'll want to think about maintaining two types for each object, one for input to the db, and one for output, This also solves the issue where ids are marked as optional, which causes user code to unnecessarily handle a possible `undefined` value. For this PR I'd just hack it unless you see a good solution.

greatbacon:
> Nice, what do you think of `actor_name` instead of just `name`? That mirrors the `actor_id` that's already there, though perhaps `author_name` is better naming for both.
> 
> Longer term we may want to have just the `actor_id` and then we perform a lookup to get any info needed, but this is a big question and tons of difficult tradeoffs, so caching the value like this PR does seems best for now. Since we're not building to scale, I like the idea of loading the full document of each community member into memory, so any frontend requirements can be easily and quickly met without a network request. This design will dramatically simplify our db querying and realtime updates, because the frontend will want the normalized form, same as the database, and normalizing means the source of truth is always in 1 place for updates.
> 
> Just mentioning for reference, in ActivityStreams this actor/author field would be `attributedTo`, which is a more specific version of its `actor` property.

`actor_name` works for me!

For the long term architectural question I guess it boils down to do we simply surface filtered tables to the front end & have it do any complex joining and/or querying, or do we simply pass the precomposed data to the front end and just have it render there. Technically we already had this information on the front end, I think it was just in a format that made it hard to query for the name from the `member` store. That might have just been my javascript skills failing me though...

greatbacon:
> For the CI error, we'll want to think about maintaining two types for each object, one for input to the db, and one for output. I'm not sure of the most maintainable way to do this yet, and it overlaps with data validation concerns. This also solves the issue some where ids are marked as optional, which causes user code to unnecessarily handle a possible `undefined` value. For this PR I'd just hack it unless you see a good solution.

Yeah, I was thinking about that. I've worked on some teams that used the term `____Document` for objects that directly relate to a datastore (SQL or NoSQL) & then referred to the object that was passed back via the api as the model or domain object.

greatbacon:
Going to close this PR & go for a different approach (front-end driven) in a separate PR per a convo @ryanatkn had this morning

ryanatkn:
followup implementation: #65

PR #64, opened 2021-07-20
add devmode

ryanatkn:
Adds the `devmode` top-level component and context store. To use it from anywhere in the component tree during component initialization:

```ts
import {get_devmode} from '@feltcoop/felt/ui/devmode.js';
const devmode = get_devmode(); // reads from context
$devmode; // writable store
```

Also had to add the `optimizeDeps` Vite config option to fix bundling, but unlike the SSR error we don't need the `ssr` config. Without the `optimizeDeps` override, it fails to prebundle properly, messing up Svelte's initialization. The newly imported Felt code uses Svelte's context helpers, so that's the root cause of this specific issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "@feltcoop/felt": "^0.4.5",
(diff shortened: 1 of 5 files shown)


greatbacon (approved):
🆒 👨‍💻

PR #65, opened 2021-07-20
Fix data query for member names on front end

greatbacon:
This fixes an inefficient name lookup for users in a chat room. Per a conversation with Ryan today we are going to try and keep things normalized until they reach the client, where we can (at a later date) build out a better data access architecture for the system.

For now, this just get's things looking nice & fixes an O(n) lookup


ryanatkn (approved):
# :racehorse: :cityscape:

PR #68, opened 2021-07-23
fix socket disconnect

ryanatkn:
Fixes the socket disconnect issue, so when the user logs out it'll disconnect. It's being encapsulated in the `Socket_Connection.svelte` component, so mounting/unmounting that component controls the socket lifecycle. I don't love the flexibility this gives us with putting the socket in context and having it tied to the component lifecycle, but it's fine for now. There's a source of truth issue here with the socket url that I don't know how to fix. We know we need multiple concurrent sockets to various servers, so I suspect we'll want to rework this to be more data-driven from the top.

Also refactors the `Socket_Store` to match the pattern elsewhere, giving us better types.

### DIFF
--- a/src/lib/ui/Socket_Connection.svelte
+++ b/src/lib/ui/Socket_Connection.svelte
@@ -2,12 +2,13 @@
 	import {onMount} from 'svelte';
 	import {get_devmode} from '@feltcoop/felt/ui/devmode.js';
 
-	import type {Socket_Store} from '$lib/ui/socket.js';
+	import {get_socket} from '$lib/ui/socket';
+
+	const socket = get_socket();
 
 	const devmode = get_devmode();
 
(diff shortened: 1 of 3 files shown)


src/lib/ui/Socket_Connection.svelte
```diff
@@ -2,19 +2,23 @@
...
 	let url = `ws://localhost:3002/ws`;
-	export let socket: Socket_Store;
 
 	onMount(() => {
 		const {hostname} = window.location;
 		console.log(hostname);
 		url = `ws://${hostname}:3002/ws`;
 		console.log('created socket store', socket, url);
 		socket.connect(url); // TODO should be reactive to `url` changes
+		return () => {
```
greatbacon, line 19: 👏  Rich Harris

greatbacon (approved):

PR #73, opened 2021-07-28
More websocket work

greatbacon:
This PR continues some of the earlier working of getting web sockets to run locally and prod. This does not actually authorize any requests still. Using cookie security will be in a follow up PR.

### DIFF
--- a/src/lib/server/server.ts
+++ b/src/lib/server/server.ts
@@ -1,31 +1,13 @@
-import type {Server as HttpServer} from 'http';
-import type {Server as HttpsServer} from 'https';
-import {createServer as create_https_server} from 'https';
 import polka from 'polka';
 import postgres from 'postgres';
 import {createServer as create_http_server} from 'http';
-import fs from 'fs';
 
 import {Api_Server} from '$lib/server/Api_Server.js';
 import {Database} from '$lib/db/Database.js';
 import {default_postgres_options} from '$lib/db/postgres.js';
 import {Websocket_Server} from '$lib/server/Websocket_Server.js';
 
-const create_server = (): HttpServer | HttpsServer => {
-	if (process.env.NODE_ENV === 'production') {
-		return create_https_server({
-			//TODO double check this aligns with GRO standard (load_https_credentials)
-			//cert: fs.readFileSync('/etc/letsencrypt/live/staging.felt.dev/fullchain.pem'),
-			//key: fs.readFileSync('/etc/letsencrypt/live/staging.felt.dev/privkey.pem'),
-			cert: fs.readFileSync('localhost.crt'),
-			key: fs.readFileSync('localhost.key'),
-		});
-	} else {
-		return create_http_server();
-	}
-};
-
-const server = create_server();
+const server = create_http_server();
 
 //TODO I think we still need to pass an https server in prod
 
--- a/src/lib/server/server.ts
+++ b/src/lib/server/server.ts
@@ -1,13 +1,33 @@
+import type {Server as HttpServer} from 'http';
+import type {Server as HttpsServer} from 'https';
+import {createServer as create_https_server} from 'https';
 import polka from 'polka';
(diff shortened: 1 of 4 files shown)


src/lib/server/server.ts
```diff
@@ -1,13 +1,33 @@
...
+			key: fs.readFileSync('localhost.key'),
+		});
+	} else {
+		return create_http_server();
+	}
+};
+
+const server = create_server();
+
+//TODO I think we still need to pass an https server in prod
```
ryanatkn, line 30: Looks like we can configure nginx

https://stackoverflow.com/questions/42761992/configuring-https-for-express-and-nginx


in the `deploy` task we have the nginx config, it needs to add the https passthrough here:

//	listen 80;

ryanatkn (changes requested):
I think should be able to configure to stay in http only mode locally by default, which incurs less overhead (see stackoverflow link in other comment), but maybe we should enable an option at startup, like a `FELT_HTTPS_CRT` and `FELT_HTTPS_KEY`? Since the code is mostly written already? That'll allow some production deployments, like to native, to optionally be sure by just specifying a cert and key.

src/lib/server/server.ts
greatbacon: that stuff gets set up by Certbot when you install a cert. I'll leave a TODO to revise NGINX setup with HTTPS in mind for the pre Certbot stage.

greatbacon:
> I think should be able to configure to stay in http only mode locally by default, which incurs less overhead (see stackoverflow link in other comment), but maybe we should enable an option at startup, like a `FELT_HTTPS_CRT` and `FELT_HTTPS_KEY`? Since the code is mostly written already? That'll allow some production deployments, like to native, to opt into https by just specifying a cert and key.

I agree that the way we're passing in the keys right now isn't great, but I don't think we should allow people to opt into HTTP for production deploys. Unless I'm misreading your comment?

ryanatkn:
Sorry this is a little confusing. What I'm saying is that in a typical deployment to a VPS, Nginx handles incoming HTTP and HTTPS traffic with this following lines:

```
listen 80;
listen 443 ssl;
```

And then our Node server runs locally on insecure HTTP, avoiding both the complexity and overhead of HTTPS, making it so the only HTTPS setup we need is with Nginx and LetsEncrypt. We'll then configure our firewall so only Nginx receives external traffic.

(the second line is the one I'm saying was missing from the commented-out Nginx config in `deploy.task.ts`, but I don't know what's being deployed)

Eventually we'll want to support more kinds of deployments, but we can avoid the issue completely for now by just assuming the server will be running behind Nginx. So in Node code, we never have to worry about certs or HTTPS servers.

src/lib/server/server.ts
ryanatkn: Gotcha yeah just looked at the old server Nginx config, and these lines are there:

```
  listen [::]:443 ssl ipv6only=on; # managed by Certbot
  listen 443 ssl; # managed by Certbot
```

Should we make a `src/lib/deploy` directory for the Nginx config and deployment docs?

ryanatkn: I'm suggesting this change:

```
const create_server = (): HttpServer | HttpsServer => {
	if (process.env.NODE_ENV === 'production') {
		return create_https_server({
			//TODO double check this aligns with GRO standard (load_https_credentials)
			//cert: fs.readFileSync('/etc/letsencrypt/live/staging.felt.dev/fullchain.pem'),
			//key: fs.readFileSync('/etc/letsencrypt/live/staging.felt.dev/privkey.pem'),
			cert: fs.readFileSync('localhost.crt'),
			key: fs.readFileSync('localhost.key'),
		});
	} else {
		return create_http_server();
	}
};

const server = create_server();
```

becomes:

```ts
const server = create_http_server();
```

ryanatkn (approved):

PR #77, opened 2021-08-04
add space type rendering for notes and iframes

ryanatkn:
- renders spaces based on their `media_type`
- renames `Chat` from `Chat_Room`
- adds a `Notes` component and seeds one
- adds an `Iframe` component and seeds one
- upgrades Felt

I originally tried to iframe Felt's library, but it loads, renders, and then errors with `The operation is insecure.`. Quick search didn't yield a workaround; not sure how to proceed right now, so I changed it to load dealt.dev/tar. Might be a sandbox attribute or something in SvelteKit? Based on the error, my guess is it's related to how SvelteKit is doing routing but I'm alright with iframing something else for now.

### DIFF
--- a/src/lib/db/Database.ts
+++ b/src/lib/db/Database.ts
@@ -218,19 +218,19 @@ export class Database {
 					this.repos.spaces.insert(community_id, {
 						name: 'chat',
 						url: '/chat',
-						media_type: 'application/json',
+						media_type: 'application/fuz+json',
 						content: '{"type": "Chat", "props": {"data": "/chat/posts"}}',
 					}),
 					this.repos.spaces.insert(community_id, {
 						name: 'board',
 						url: '/board',
-						media_type: 'application/json',
+						media_type: 'application/fuz+json',
 						content: '{"type": "Board", "props": {"data": "/board/posts"}}',
 					}),
 					this.repos.spaces.insert(community_id, {
 						name: 'voice',
 						url: '/voice',
-						media_type: 'application/json',
+						media_type: 'application/fuz+json',
 						content: '{"type": "Voice", "props": {"data": "/voice/stream"}}',
 					}),
 				]);
--- a/src/lib/db/seed.ts
+++ b/src/lib/db/seed.ts
@@ -289,7 +289,7 @@ export const seed = async (db: Database): Promise<void> => {
 			name: 'felt library',
 			url: '/felt-library',
 			media_type: 'application/fuz+json',
-			content: '{"type": "Iframe", "props": {"url": "https://www.felt.dev/sketch/library"}}',
+			content: '{"type": "Iframe", "props": {"url": "https://www.dealt.dev/tar"}}',
 		};
 		const space5_result = await sql`
 			insert into spaces ${sql(space5, 'url', 'media_type', 'content')}
--- a/src/lib/db/seed.ts
+++ b/src/lib/db/seed.ts
@@ -286,8 +286,8 @@ export const seed = async (db: Database): Promise<void> => {
 	const space5 = space_docs.find((d) => d.space_id === 5);
 	if (!space5) {
 		const space5: Space_Params = {
-			name: 'felt library',
-			url: '/felt-library',
+			name: 'dealt: tar',
+			url: '/tar',
 			media_type: 'application/fuz+json',
 			content: '{"type": "Iframe", "props": {"url": "https://www.dealt.dev/tar"}}',
 		};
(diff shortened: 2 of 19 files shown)


src/lib/db/Database.ts
```diff
@@ -218,19 +218,19 @@ export class Database {
 					this.repos.spaces.insert(community_id, {
 						name: 'chat',
 						url: '/chat',
-						media_type: 'application/json',
+						media_type: 'application/fuz+json',
```
greatbacon, line 221: Should we pick a better term than `fuz` to start with?

src/lib/db/seed.ts
```diff
@@ -260,15 +260,43 @@ export const seed = async (db: Database): Promise<void> => {
...
+		const space4_result = await sql`
+			insert into spaces ${sql(space4, 'url', 'media_type', 'content')}
+		`;
+		console.log('[db] create_space4_result', space4_result);
+	}
+
+	const space5 = space_docs.find((d) => d.space_id === 5);
+	if (!space5) {
+		const space5: Space_Params = {
+			name: 'felt library',
```
greatbacon, line 289: Since there was an issue with the felt library, should we update this to "dealt"?

src/lib/ui/Chat.svelte
```diff
@@ -28,31 +28,10 @@
...
-			body: JSON.stringify({content: text}),
-		});
-		if (res.ok) {
-			console.log('post sent, broadcasting to server');
-			const data = await res.json();
-			broadcast_post(data);
-		} else {
-			console.error('error sending post');
-		}
+		await api.create_post(space, text);
```
greatbacon, line 31: 🥇

src/lib/ui/Chat.svelte
```diff
@@ -72,12 +51,14 @@
 		display: flex;
 		flex-direction: column;
 		flex: 1;
+		overflow: hidden; /* make the content scroll */
```
greatbacon, line 54: So I just tested this on my local machine, and now it no longer pushes chat off the bottom (yay!) but I can't seem to scroll back through the chat history now (aww).

src/lib/ui/Iframe.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+	import type {Space} from '$lib/spaces/space.js';
+
+	export let space: Space;
+	export let url: string; // TODO type
+</script>
+
+<!-- TODO figure out sandboxing -- allow-same-origin? -->
+<iframe sandbox="allow-scripts allow-pointer-lock" frameborder="0" title={space.name} src={url} />
```
greatbacon, line 9: I noticed a little delay between clicking into an Iframe space and it actually loading. Is there an easy way to use the spin loader component here?

src/lib/ui/Space_Input.svelte
```diff
@@ -26,8 +26,8 @@
 			community.community_id,
 			new_name,
 			url,
-			'application/json',
-			`{"type": "Chat_Room", "props": {"data": "${url}/posts"}}`,
+			'application/fuz+json',
```
greatbacon, line 29:
```suggestion
//TODO : add space type picker
			'application/fuz+json',
```

greatbacon (approved):
This is really exciting to see in action! I think especially the "post" backing both a note & chat space!!!

src/lib/ui/Chat.svelte
ryanatkn: This might be a bug in both FF and Chrome:

- https://stackoverflow.com/a/37515194
- https://github.com/philipwalton/flexbugs/issues/53

Instead of using `justify-content: flex-end` I'm using `flex-direction: column-reverse` as described here: https://stackoverflow.com/a/44051405

Seems to be working :crossed_fingers:

src/lib/db/seed.ts
ryanatkn: Went with this for now, not really sure about it:

```ts
			name: 'dealt: tar',
			url: '/tar',
```

src/lib/ui/Iframe.svelte
ryanatkn: Added a pending animation overlay component in this commit: https://github.com/feltcoop/felt-server/pull/77/commits/79dc66c885db879d33e5ce76fb325e9fcbab94d7

We'll probably want to tweak Felt's pending animation now that we're seeing it in multiple use cases -- I think it could be a bit snappier, feels a bit mushy right now.

Another possible improvement could be adding an animation to the button, to be reused whenever a button does something async.

ryanatkn:
I think I addressed all the points, will wait if you want to take another look before merging.

src/lib/db/Database.ts
ryanatkn: Keeping for now after discussing.

greatbacon (approved):
Cool! Thanks for the fixes & clarifications! :shipit:

PR #78, opened 2021-08-09
Add initialize task and config file for nginx

greatbacon:
This PR adds a gro task for initializing a clean install Linux(Ubuntu) distro into the place where it can run a production instance of the felt-server.

### DIFF
--- a/src/infra/felt-server.conf
+++ b/src/infra/felt-server.conf
@@ -1,0 +1,20 @@
+server {
+        server_name <put your URL here>;
+
+        location /api {
+                proxy_pass http://localhost:3001;
+        }
+
+        location /ws {
+                proxy_set_header Upgrade $http_upgrade;
+                proxy_set_header Connection "upgrade";
+                proxy_http_version 1.1;
+                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+                proxy_set_header Host $host;
+                proxy_pass http://localhost:3001;
+        }
+
+        location / {
+                proxy_pass http://localhost:3000;
+        }
+}
--- a/src/infra/felt-server.conf
+++ b/src/infra/felt-server.conf
@@ -1,4 +1,5 @@
 server {
+        #use `gen` to leverage env/config vars and build this string per PR#78
         server_name <put your URL here>;
 
         location /api {
--- a/src/infra/initialize.task.ts
+++ b/src/infra/initialize.task.ts
@@ -1,0 +1,46 @@
+import type {Task} from '@feltcoop/gro';
+import {spawn} from '@feltcoop/felt/util/process.js';
+import {DIST_DIRNAME} from '@feltcoop/gro/dist/paths.js';
+
+export const task: Task = {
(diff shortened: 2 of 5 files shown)


ryanatkn:
Looks good, we can refactor `root@50.116.30.248` and `certbot --nginx -d example.com -d www.example.com` with environment variables once we have that in. I was planning to get on that sometime this week.

ryanatkn (approved):

src/infra/initialize.task.ts
```diff
@@ -0,0 +1,46 @@
...
+			'root@50.116.30.248',
+			`apt update;
+			apt install unzip;
+      curl -fsSL https://fnm.vercel.app/install | bash;`,
+		]);
+		//Splitting these tasks here let's fnm get picked up from the bash profile
+		await spawn('ssh', [
+			'root@50.116.30.248',
+			`fnm install 16.6.0;
+		 	apt install npm;`,
```
ryanatkn, line 21: This will pull in whatever version of `npm` is in apt, which might be outdated. `npm` should be available once `fnm` is ready.

src/infra/initialize.task.ts
```diff
@@ -0,0 +1,46 @@
...
+		await spawn('ssh', [
+			'root@50.116.30.248',
+			`apt update;
+			apt install unzip;
+      curl -fsSL https://fnm.vercel.app/install | bash;`,
+		]);
+		//Splitting these tasks here let's fnm get picked up from the bash profile
+		await spawn('ssh', [
+			'root@50.116.30.248',
+			`fnm install 16.6.0;
```
ryanatkn, line 20: Should probably change this to `16` or `16.6`. I think the former should be compatible -- the latter will make sure at least the patches get in, 16.6.1 for example has a security fix.

src/infra/initialize.task.ts
```diff
@@ -0,0 +1,46 @@
+import type {Task} from '@feltcoop/gro';
```
ryanatkn, line 1: What do you think of naming this `setup` instead? Mainly just less to type, or `init` if you like those semantics better?

src/infra/felt-server.conf
```diff
@@ -0,0 +1,20 @@
+server {
+        server_name <put your URL here>;
```
ryanatkn, line 2: I think we should generate this file with `gen`, but after we get the env/config stuff implemented. The way it'd work is you make a sibling file at `src/infra/felt-server.gen.conf.ts` and then write a function to return the constructed string -- here's an example: https://github.com/ryanatkn/corpus-activity-streams/blob/29fdb7960fe081553044a34eae028f024aec0d44/src/activity_streams_examples_ast.gen.json.ts

src/infra/initialize.task.ts
greatbacon: I like `setup`, I think that actually captures the spirit a little better.

src/infra/felt-server.conf
greatbacon: Cool! I'll add a TODO comment to leverage `gen` here once we figure out our env/config stuff.

PR #79, opened 2021-08-09
improve mobile view

ryanatkn:
Makes the mobile view much better, using animations and responsive layout on the main menu.

Also adds an `app` context helper to keep things succinct. I think it's a good call but I could be off.

Possible followup TODOs for later PRs:

- [ ] keep the menu button fixed onscreen z-index overlaying the menu as it slides in/out, so quick tapping always hits it
- [ ] consider a transition out animation
- [ ] consider ways to share the media query size with the JS variable to DRY things up, tweak the transitions for desktop vs mobile, and avoid inserting the bg element into the DOM when not used -- gets tricky with SSR

### DIFF
--- a/src/lib/ui/Main_Nav.svelte
+++ b/src/lib/ui/Main_Nav.svelte
@@ -9,6 +9,11 @@
 
 	const {data, ui, api} = get_app();
 
+	// TODO share these vars with CSS
+	const MAIN_NAV_WIDTH = 320; // TODO this is the same as `column_width_min`
+	const NAV_TRANSITION_IN_DURATION = 410;
+	const BG_TRANSITION_IN_DURATION = 237;
+
 	$: members = $data.members;
 	$: communities = $data.communities;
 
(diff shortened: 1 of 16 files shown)


src/lib/ui/Main_Nav.svelte
```diff
@@ -28,8 +28,14 @@
 	// $: console.log('[Main_Nav] selected_space', selected_space);
 </script>
 
-<div class="main-nav">
+<div
+	class="main-nav-bg"
+	in:fade={{duration: 237}}
```
greatbacon, line 33: Should this be extracted as a CSS/JS var?

greatbacon (approved):
I like the `app` refactor! I noticed you had a bunch of `rethink` comments around it, so I'll be curious to hear what other thoughts you have about it.

src/lib/ui/Main_Nav.svelte
ryanatkn: Oh yeah good idea, extracted to JS vars and added a TODO. I think they'll belong in Felt and be shared between CSS and JS -- we don't have a system for that yet but it's not too far away.

ryanatkn:
Glad you like it, I have a handful of questions, none of which would prevent me from merging it as is:

- does `app` tie together too many concerns, making components that import it overly reliant on this particular app's structure, harming portability? I think it's ok for components that live inside `felt-server`, but say we want to upstream a component to Felt -- if it imports `devmode` via `app`, that import will have to be changed to import `devmode` from its module directly, unless we add `app` import functionality similar to how SvelteKit does things with the `$app/*` imports: https://kit.svelte.dev/docs#modules
- is it overly restrictive for `app` to be a plain object with stores, rather than a reactive store itself? As is, 
-

PR #81, opened 2021-08-10
Add personas

greatbacon:
This PR introduces the Persona table and interjects it as a interface between an Account (which has a 1:1 representational relationship with a human; and a persona which covers the different contexts we interact with communities under.

Currently there is no way to make a new persona, delete a persona, or switch between multiple personas, so it is functionally similar to the pre-persona account behavior. 

- [ ] There's still a bug where when switching between accounts that don't share communities the member list doesn't update properly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "@feltcoop/felt": "^0.6.0",
+        "@feltcoop/felt": "^0.7.0",
         "@polka/send-type": "^0.5.2",
         "@types/ws": "^7.4.4",
         "body-parser": "^1.19.0",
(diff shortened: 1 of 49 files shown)


ryanatkn (approved):
Nice this looks great. What's the thinking behind the `active_persona` in the session? I'd expect that to be temporary state in the client, and still load everything by account.

greatbacon:
> Nice this looks great. What's the thinking behind the `active_persona` in the session? I'd expect that to be temporary state in the client, and still load everything by account.

That was a shortcut in response to swapping the Persona with Account as the main relation for communities & posts. I suppose we could load all communities for all personas on the backend and pass it to the client. I guess I assumed the `active_persona` was the temporary state in the client. If that's not the right solution what should we do instead?

greatbacon:
> > I suppose we could load all communities for all personas on the backend and pass it to the client.
> 
> That sounds good to me, thick session payload. A future optimization could let clients specify only a subset of personas/communities, but for now the full take of (meta)data makes sense to send on load.

Cool; in that case is session still an okay place for storing the "active" persona to render from the thick payload? Or is there a better place we could/should be storing that for now?

ryanatkn:
I think the right spot with current patterns is `src/ui/ui.ts` alongside state like `selected_community_id`. We can assume that there's one persona per community for now.

greatbacon:
Okay, I'm going to go ahead and merge what I have, and open another PR for managing any frontend state & rendering work!

PR #82, opened 2021-08-11
security pass

ryanatkn:
Changes the session to use `account_id`, and then sends only a subset of fields back to the client. Not sure about the naming of the various `Account` interfaces -- I left a todo with some ideas, and it fits the current pattern. Also removes the old hacky `/context` API route.

### DIFF
--- a/src/lib/db/Database.ts
+++ b/src/lib/db/Database.ts
@@ -7,7 +7,7 @@ import type {Space, Space_Params} from '$lib/spaces/space.js';
 import type {Post} from '$lib/posts/post.js';
 import type {Member} from '$lib/members/member.js';
 import type {Account, Account_Model, Account_Params} from '$lib/vocab/account/account.js';
-import {account_model_properties} from '$lib/vocab/account/account';
+import {account_properties, account_model_properties} from '$lib/vocab/account/account';
 import type {Postgres_Sql} from '$lib/db/postgres.js';
 
 export interface Options {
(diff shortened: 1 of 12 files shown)


src/lib/db/Database.ts
```diff
@@ -31,24 +32,29 @@ export class Database {
...
-			load_client_session: async (name: string): Promise<Result<{value: Account_Session}>> => {
-				console.log('[db] load_client_session', name);
-				let account: Account = unwrap(await this.repos.accounts.find_by_name(name));
-				let communities: Community[] = unwrap(
-					await this.repos.communities.filter_by_account(account.account_id!),
+			load_client_session: async (
+				account_id: number,
+			): Promise<Result<{value: Account_Session}>> => {
+				console.log('[db] load_client_session', account_id);
+				// TODO refactor this hardcoded array with `Account_Model` data, make typesafe with columns
```
greatbacon, line 39: Is this comment referring to the hardcoded array at line 72, or something else?

src/lib/server/Api_Server.ts
```diff
@@ -106,9 +104,6 @@ export class Api_Server {
 				log.info('echo', req.body);
 				send(res, 200, req.body);
 			})
-			.get('/api/v1/context', (req, res) => {
```
greatbacon, line 109: 🧹 💨

src/lib/session/login_middleware.ts
```diff
@@ -47,11 +47,14 @@ export const to_login_middleware = (server: Api_Server): Middleware => {
 			// There's already an account, so proceed to log in after validating the password.
 			account = find_account_result.value;
 			if (account.password !== password_hash) {
-				return send(res, 400, {reason: 'wrong password'});
+				return send(res, 400, {reason: 'invalid account name or password'});
```
greatbacon, line 50: 🔒

greatbacon (approved):
LGTM! Just had one comment about what (maybe?) is a stray comment line in the DB repo

src/lib/db/Database.ts
ryanatkn: Good catch, it referred to code that has since changed but also applies at line 72, I'll delete this comment and extract 72 to a var.

PR #83, opened 2021-08-11
make nav a permanent onscreen object

ryanatkn:
This makes the `Main_Nav` a permanent fixture onscreen. It also flips the Svelte compiler flag `immutable` on.

### DIFF
--- a/src/lib/ui/Main_Nav.svelte
+++ b/src/lib/ui/Main_Nav.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import Markup from '@feltcoop/felt/ui/Markup.svelte';
 	import {icons} from '@feltcoop/felt';
+	import {to_random_seeded} from '@feltcoop/felt/util/random_seeded.js';
 
 	import Actor_Icon from '$lib/ui/Actor_Icon.svelte';
 	import Community_Nav from '$lib/ui/Community_Nav.svelte';
@@ -8,7 +9,6 @@
 	import Socket_Connection from '$lib/ui/Socket_Connection.svelte';
 	import Account_Form from '$lib/ui/Account_Form.svelte';
 	import {get_app} from '$lib/ui/app';
-	import {to_random_seeded} from '@feltcoop/felt/util/random_seeded.js';
 
 	const {data, ui, api} = get_app();
 
--- a/src/lib/ui/Main_Nav.svelte
+++ b/src/lib/ui/Main_Nav.svelte
@@ -1,11 +1,14 @@
 <script lang="ts">
 	import Markup from '@feltcoop/felt/ui/Markup.svelte';
+	import {icons} from '@feltcoop/felt';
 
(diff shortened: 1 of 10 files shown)


src/lib/ui/Main_Nav.svelte
```diff
@@ -26,53 +29,67 @@
 	// $: console.log('[Main_Nav] communities', communities);
 	// $: console.log('[Main_Nav] selected_community', selected_community);
 	// $: console.log('[Main_Nav] selected_space', selected_space);
+
+	// TODO not sure about this `--hue` api -- see in 2 places
+	// TODO refactor to some client view-model for the actor
+	$: hue = to_random_seeded($data.account.name)() * 360; // TODO random int
```
greatbacon, line 35: This seems like a good way to get dynamic hue coloring (since I don't think you can do something like this inside the CSS itself?). Might be worth tagging as a util function to pull up to `felt` as well.

src/lib/ui/style.css
```diff
@@ -7,8 +7,11 @@
 	--luggage_size: var(--icon_size_md);
 	--navbar_size: var(--luggage_size);
 
-	--transition_duration_md: 0.32s;
-	--transition_duration_sm: 0.24s;
+	--transition_duration_xl: 0.75s;
+	--transition_duration_lg: 0.56s;
+	--transition_duration_md: 0.36s;
+	--transition_duration_sm: 0.22s;
+	--transition_duration_xs: 0.08s;
```
greatbacon, line 14: Is there a theory/principle behind these time chunks? My first blush is that it seems odd to not use similar/rounder differences between each size.

greatbacon (approved):
Had a couple questions/comments, but nothing blocking. LGTM!

src/lib/ui/Main_Nav.svelte
ryanatkn: Good idea, does it belong in a hypothetical `src/util/color.ts` or the already-existing `src/util/random.ts`? I imagine we're going to add more color functions soon now that we're designing with it, so I think the new file could make sense.

src/lib/ui/style.css
ryanatkn: 2 are haphazard and 3 are deliberate but rough, am open to changing. In particular `transition_duration_md` and `transition_duration_lg` are not used -- they were just eyeballed by me to fill in the gaps between `transition_duration_xl`, `transition_duration_sm`, and `transition_duration_xs`, which do have values that I chosen to get a particular effect.

src/lib/ui/Main_Nav.svelte
ryanatkn: I extracted these to a color helper here, with a todo to upstream to Felt (after some API stuff is figured out): https://github.com/feltcoop/felt-server/pull/83/commits/2ebcd760264e3ce1698445534efa135485c6857e

PR #84, opened 2021-08-14
add board and forum spaces

ryanatkn:
Fills out the default spaces a bit more, adding `Board` and `Forum`. Not much thought has gone into them, just trying to rough things in and change them enough so they're distinct, so we can think more about them.

Also factors out a `create_space` seed helper, just a temporary pattern while we think about how we actually want to structure things.


src/lib/db/seed.ts
```diff
@@ -227,130 +229,64 @@ export const seed = async (db: Database): Promise<void> => {
...
-	if (!space5) {
-		const space5: Space_Params = {
-			name: 'dealt: tar',
-			url: '/tar',
-			media_type: 'application/fuz+json',
-			content: '{"type": "Iframe", "props": {"url": "https://www.dealt.dev/tar"}}',
-		};
-		const space5_result = await sql`
-			insert into spaces ${sql(space5, 'url', 'media_type', 'content')}
+	const create_space = async (
```
greatbacon, line 232: Is there a way to start leveraging the DB repo functions in this file?

greatbacon (approved):
Just had one comment on the db seed file (really more of just a thought than anything). LGTM tho!

src/lib/db/seed.ts
ryanatkn: lol yes good one, here's the commit: https://github.com/feltcoop/felt-server/pull/84/commits/39755cfe055ea087c582d3a289cc134ed4ccfd9b

PR #88, opened 2021-08-19
Render personas

greatbacon:
Okay, so this piece now adds Personas to the session data & ui store; updates the community nav based on which persona is selected, and provides a mechanism for picking between personas.

It is not perfectly styled though, as a caveat.

I know this wasn't exactly what we had previously discussed, but this was what I could figure out to get a workable state 🙃


ryanatkn:
Got the piece for setting the initial state here: https://github.com/feltcoop/felt-server/pull/92

src/lib/ui/ui.ts
```diff
@@ -66,16 +84,35 @@ export const to_ui_store = () => {
...
 						main_nav_view: 'explorer',
 					};
 				}
 			});
 		},
+		select_persona: (persona_id: number | null) => {
+			console.log(typeof persona_id);
+			update(($ui) => ({
+				...$ui,
+				selected_persona_id: persona_id,
```
ryanatkn, line 100: When selecting a persona we'll want to update `selected_community_id` as well. With better code this won't be needed -- the `selected_community_id` should be computed, not updated manually like this -- but for now this makes things behave better.

```
selected_community_id: $ui.selected_community_id_by_persona[persona_id],
```

(also, remove the `| null` from the type signature or you'll get a type error, it's not there in the store signature, and I'm puzzled why it's not erroring but w/e)

src/lib/ui/MainNav.svelte
```diff
@@ -43,15 +47,17 @@
 		<div class="header">
 			<!-- TODO how to do this? -->
 			<div class="icon-button button-placeholder" />
+			<select class="persona-selector" on:change={(event) => selectPersona(event)}>
+				{#each personas as persona (persona)}
+					<option value={persona.persona_id}>{persona.name}</option>
+				{/each}
+			</select>
```
ryanatkn, line 54: No need to change this since we'll likely rewrite it soon, but you could simplify this `<select>` with Svelte's `bind:group`, which allows you to use the persona objects directly without drawbacks: https://svelte.dev/examples#select-bindings

ryanatkn (approved):
Looks good! Good step towards the functionality we want. I had a couple comments with a 1 line change/bugfix. Not far off from the behavior we talked about, and we can pair up next if you'd like.

PR #98, opened 2021-09-01
add env

ryanatkn:
Adds support for a `.env` file in the root directory.

There are a few categories of variables so far:

- server-only vars like the server hosts
- client-accessible vars like the websocket URL
- secrets like the cookie keys
- deployment vars like the VPS IP address

The client-accessible vars are supported by Vite, and must contain the `VITE_` prefix so it can safely expose them to the client. Here's Vite's docs: https://vitejs.dev/guide/env-and-mode.html

The `dotenv` library is used by both Vite and the Felt server to load environment variables: https://github.com/motdotla/dotenv

Vite and `dotenv` recommend different processes -- the former recommends using different files for modes and `.local` secrets, and the latter recommends against doing this. I'm not sure which to go with yet, so I'm starting with the simplest version, a single `.env` file that gets bootstrapped from `src/infra/.env.default` if it doesn't exist. We'll need to do some more work to support production deployments.

Also updates SvelteKit and the Node adapter.

- [x] basic impl
- [x] fix build failing due to `import.meta.env` usage
- [x] production and development versions
- [x] move non-secrets to `src/lib/constants.js`

In a followup we'll want to refactor this to do proper validation and use it for the Postgres vars.


src/infra/.env.production.default
```diff
@@ -0,0 +1,7 @@
+VITE_WEBSOCKET_URL=wss://staging.felt.dev/ws
+
+COOKIE_KEYS=TODO_key1__key2__key3
+
+DEPLOY_IP=50.116.30.248
```
greatbacon, line 5: Long term I think this should be stuff that lives outside the `.env` context; probably as a config file or something that lives in an `infra` context.

greatbacon (approved):

src/infra/.env.production.default
ryanatkn: Oh yeah! I went ahead and refactored to do that, so now everything but secrets live in `src/lib/constants.js`, as a JS file so `svelte.config.js` can import it during development.

I think this works well for now, but later we'll need to design how to override this in user projects. We'll probably extend  `gro.config.ts` somehow.

PR #101, opened 2021-09-03
add salt to password check and refactor database seeding

ryanatkn:
Adds salt to the password hashing and verification steps. Uses the constant time check `timingSafeEqual` to mitigate timing attacks on password checks.

Previously the seeded accounts used hardcoded password hashes, but now that the process uses a random salt per entry, we need to use the database repo methods. I refactored `lib/db/seed.ts` to use repo methods and data. The next PRs here will be to further refactor, add additional seed data, and then set up migrations.

Also adds tests because the API seems stable and is of more than minor importance. :grin: 

Changed a couple of `insert` repo methods to `create` to better match vocab semantics.

- [x] implement
- [x] test
- [x] fix hardcoded seed passwords (use repo functions)
- [x] refactor some repo functions, like making `account.create` not make an initial persona and community, and making `persona.create` making the initial community
- [x] refactor more of `lib/db/seed.ts` to use repo functions and add default posts with valid personas

Hashing and checking passwords is computationally heavy and can block the main thread to a degree that'll noticeably degrade performance for end users; after the next milestone I made a note to use [worker_threads](https://nodejs.org/api/worker_threads.html) here to avoid blocking the main thread.


src/lib/vocab/community/communityRepo.ts
```diff
@@ -45,26 +45,22 @@ export const communityRepo = (db: Database) => ({
...
 		const community_id = community.community_id;
-		console.log(community_id);
 		// TODO more robust error handling or condense into single query
-		const association = await db.sql<any>`
+		await db.sql<any>`
       INSERT INTO persona_communities (persona_id, community_id) VALUES (
       ${persona_id},${community_id}
       )
     `;
-		console.log('[db] created persona_communities', association);
```
greatbacon, line 64: How come you removed the `insert default spaces` chunk here?

src/lib/vocab/account/accountRepo.ts
```diff
@@ -13,16 +13,8 @@ export const accountRepo = (db: Database) => ({
       insert into accounts (name, password) values (
         ${name}, ${password}
       ) RETURNING *`;
-		console.log(data);
+		console.log('[db] created account', data);
 		const account = data[0];
-		const persona_response = await db.repos.persona.create(name, account.account_id);
```
greatbacon, line 18: Without this block new accounts won't have a default persona associated with them, which means no communities for an account.

ryanatkn: Yikes good catch, the PR breaks creating new accounts. The reasoning was that having a persona with the same name as your account seemed like a bug. I was thinking (but didn't implement) that after logging into a fresh account, the app shows no personas and asks you to create one.

I'll revert this for now. The issues it was causing may be aesthetic only.

src/lib/vocab/community/communityRepo.ts
ryanatkn: This was another mistake. The half-baked motivating thought was that creating default spaces should be lifted out of the repo into the service layer, but that's not necessarily a good idea and it's out of scope for this PR, and all I did here was delete it from the repo and move it to the seed, which is broken. I reverted this change

greatbacon (approved):
Thanks for fixing those regressions! I agree with the spirit of both removals, but yeah, I think those can be tackled in later PRs.

PR #102, opened 2021-09-03
extract repos to standalone modules

ryanatkn:
Extracts all of the repo functions to standalone modules, all in `$lib/vocab` except for `session`. 

The repos are stateless, just an object wrapper instantiated with dependencies like `db` around some named functions. Type inference is used instead of defining the return values to keep things simple.

I changed the repo names to be singular to mirror the vocab instead of the database tables. I'm not 100% sure about this but it feels right at the moment -- the reasoning is that repo functions primarily deal with vocabulary, and the database tables are an implementation detail. I think it's possible we might have a vocabulary term and corresponding repo using two database tables, and the term has a name that references neither table. This also makes sense if we refactor the database to have polymorphic tables for files/activities. I'm open to changing this.


greatbacon (approved):
I think not having repos be 1:1 with the tables & instead align with vocabulary makes sense to me and is a good pattern.

PR #104, opened 2021-09-06
add authorization middleware

### LINKED ISSUE #19: Add session authn tooling to platform
When looking up user info on the server we should have some sort of tool/util/interceptor to confirm an account is authenticated to access the resources

ryanatkn:
closes #19 

Centralizes authorization to a single middleware instead of having each route manually do its own checks for a `401`. For now, this makes all `/api/*` routes require an authenticated account. We'll want to change this to allow public publishing, but for now it's simple and secure to do it this way.

- [x] understand why the login middleware doesn't fail authorization given Polka's middleware behavior: https://github.com/lukeed/polka#middleware-sequence -- it does fail if re-ordered to follow the authorization middleware, counter to my understanding of Polka's docs (answer: 1.0@next has different behavior than the stable 0.5.x docs)


greatbacon (approved):
🔒

PR #105, opened 2021-09-06
add generic service middleware

ryanatkn:
This is a large PR that adds more robust service definitions. The goal is to make services -- which correspond to vocab verbs/events -- usable by both http requests and websocket events. It adds two new top-level dependencies, [`ajv`](https://github.com/ajv-validator/ajv) for JSON schema validation and [`@sinclair/typebox`](https://github.com/sinclairzx81/typebox) to build typed schemas. (so be sure to `npm i`)

- [x] convert endpoint middlewares to services
- [x] http middleware
- [x] hacky websocket middleware
- [x] service response types
- [x] declarative schemas for the input params of each service

Future PRs:

- add declarative schemas for the output response of each service (probably validated only in dev mode)
- handle multiple validation errors instead of just the first one
- parse instead of validate to handle additional formats
- maybe confirm to [JSON RPC 2.0](https://www.jsonrpc.org/specification)
- refine APIs


src/lib/server/WebsocketServer.ts
```diff
@@ -8,13 +8,19 @@ import {to_cookie_session_middleware} from '$lib/session/cookie_session';
...
+	(websocket_server: WebsocketServer, socket: ws, message: any, account_id: number): Promise<void>;
+}
+
 export class WebsocketServer {
 	readonly wss: ws.Server;
 	readonly server: HttpServer | HttpsServer;
+	readonly handle: HandleWebsocketServerMessage;
 
-	constructor(server: HttpServer | HttpsServer) {
+	constructor(server: HttpServer | HttpsServer, handle: HandleWebsocketServerMessage) {
```
greatbacon, line 20: Is this supposed to relate to `HandleWebsocketMessage`? I don't really see anything that ties the two together from an interface/implementation level?

src/lib/server/server.ts
```diff
@@ -6,15 +6,21 @@ import {ApiServer} from '$lib/server/ApiServer.js';
...
 
-export const api_server = new ApiServer({
+// TODO this explicit type declaration is needed because of the awkward
+// `handle_websocket_message` API, there's some circularity in deps to maybe refactor
+export const api_server: ApiServer = new ApiServer({
 	server,
 	app: polka({server}),
-	websocket_server: new WebsocketServer(server), // TODO probably pass `{server}` when fixing socket auth
+	websocket_server: new WebsocketServer(server, (websocket_server, socket, message, account_id) =>
+		handle_websocket_message(api_server, websocket_server, socket, message, account_id),
```
greatbacon, line 22: Is this not something that can happen inside the WebsocketServer constructor itself?

greatbacon (approved):
Cool! I really just had the 1 question around how the websocket stuff is behaving, otherwise lgtm

ryanatkn:
Did some minor stuff since you reviewed -- https://github.com/feltcoop/felt-server/pull/105/commits/cca67289e41c339ff7e70fe1701731d11bd733d3 --

- cleaned up the docs and notes in $lib/server/service.ts`
- renamed the `service_handlers` map to `services`, and `handler.handle(` to `service.handle`. I'm wondering if there's a better name than `handle`.

I should have mentioned, the middleware to service transformation was pretty simple but it's hard to follow because the files were also renamed -- it mostly just changes from using `send(` to a return value and wraps in the new `Service` definition.

src/lib/server/server.ts
ryanatkn: That's a messy part of the PR, maybe the messiest bit. Two things here:

First is this awkward handle function in `$lib/server/server.ts`. The idea here was to make `WebsocketServer` generic and simple, so it has no dependency on `ApiServer` and delegates to its `handle` constructor option. (maybe needs a rename) This might be an inadvisable goal. It makes the `ApiServer` instantiation here very weird, and because of the circular dependency we can't change the second arg to `WebsocketServer` to `handle_websocket_message.bind(null, api_server)`, because `api_server` is still being created. I'll think about this a bit more before merging, would appreciate any thoughts you might have. (and lmk if it's confusing)

Second and less important is the names used on the client and server. `HandleWebsocketServerMessage` and then `HandleSocketMessage` on the client. Maybe the latter should be `HandleWebsocketClientMessage`? I dunno. It's a little odd that our code uses `Websocket` on the server, `Socket` on the client, and the server library uses `WebSocket`.

src/lib/server/WebsocketServer.ts
ryanatkn: See the comment below, motivation was to extract the `ApiServer` dependency out of the `WebsocketServer`.

src/lib/server/server.ts
ryanatkn: Ok I refactored this here, cleaning it up a bunch -- https://github.com/feltcoop/felt-server/pull/105/commits/e1eaeea709f80d29f90d43c86ba3387a4ac18f49

The assumption now is that `ApiServer` encapsulates service handler logic for both http and websockets. It was already doing it with http, so it didn't make sense (and complicated the code) to split it out. I think it's likely we'll extract more dependencies out of `ApiServer` in the future, like the services so they're more easily customized.

PR #107, opened 2021-09-09
Add space type picker

greatbacon:
This PR adds a simple selector element to the new Space dialog box which allows a user to select the Type of new space.

It also adds an enumeration object (SpaceType) although I'm not sure if it's actually in the right "semantics" for our framework.

There's also currently a weird bug, where if you click into a new space just after creating it, Svelte throws a bunch of errors (see below)
Screen Shot 2021-09-09 at 3 30 13 PM

### DIFF
--- a/src/lib/vocab/space/default_spaces.ts
+++ b/src/lib/vocab/space/default_spaces.ts
@@ -36,8 +36,7 @@ export const default_spaces: SpaceParams[] = [
 		name: 'felt library',
 		url: '/library',
 		media_type: 'application/fuz+json',
-		content:
-			`{"type": "${SpaceTypes.Iframe}", "props": {"url": "https://www.felt.dev/sketch/library"}}`,
+		content: `{"type": "${SpaceTypes.Iframe}", "props": {"url": "https://www.felt.dev/sketch/library"}}`,
 	},
 	{
 		name: 'dealt: tar',
--- a/src/lib/vocab/space/default_spaces.ts
+++ b/src/lib/vocab/space/default_spaces.ts
@@ -1,46 +1,47 @@
 import type {SpaceParams} from '$lib/vocab/space/space';
+import { SpaceTypes } from '$lib/vocab/space/space';
 
 export const default_spaces: SpaceParams[] = [
 	{
 		name: 'chat',
 		url: '/chat',
 		media_type: 'application/fuz+json',
-		content: '{"type": "Chat", "props": {"data": "/chat/files"}}',
+		content: '{"type": "'+SpaceTypes.Chat+'", "props": {"data": "/chat/files"}}',
 	},
 	{
 		name: 'board',
 		url: '/board',
 		media_type: 'application/fuz+json',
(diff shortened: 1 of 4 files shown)


src/lib/vocab/space/default_spaces.ts
```diff
@@ -1,46 +1,50 @@
 import type {SpaceParams} from '$lib/vocab/space/space';
+import {SpaceTypes} from '$lib/vocab/space/space';
 
 export const default_spaces: SpaceParams[] = [
 	{
 		name: 'chat',
 		url: '/chat',
 		media_type: 'application/fuz+json',
-		content: '{"type": "Chat", "props": {"data": "/chat/files"}}',
+		content: '{"type": "' + SpaceTypes.Chat + '", "props": {"data": "/chat/files"}}',
```
ryanatkn, line 9: Any objections to changing these lines from:

```ts
content: '{"type": "' + SpaceTypes.Chat + '", "props": {"data": "/chat/files"}}',
```
to 

```ts
content: `{"type": "${SpaceTypes.Chat}", "props": {"data": "/chat/files"}}`,
```

src/lib/vocab/space/space.ts
```diff
@@ -12,3 +12,12 @@ export interface SpaceParams {
...
 }
+
+export const SpaceTypes = {
+	Chat: 'chat',
+	Board: 'board',
+	Forum: 'forum',
+	Notes: 'notes',
+	Voice: 'voice',
+	Iframe: 'iframe',
+};
```
ryanatkn, line 23: I think we should change these to be capitalized like in ActivityStreams. That'll also give us a 1:1 lookup from type to component name. I'm wondering if we'll extend the vocabulary by adding a `Space` that extends `Object` and then has these various subtypes.

ryanatkn:
Nice this is cool to see! One small addition could be to add a button that calls `create`. I have some ideas for refactoring things but saving it for a followup PR.

ryanatkn (approved):

src/lib/vocab/space/space.ts
ryanatkn: Also as written this is probably a good usecase for TypeScript String `enum` -- https://www.typescriptlang.org/docs/handbook/enums.html#string-enums -- but I think I want to change things a bit differently, make them extensible instead of a closed set of types.

src/lib/vocab/space/default_spaces.ts
greatbacon: nope! I was thinking about how to do exactly that but spaced it in the moment 🤣

src/lib/vocab/space/space.ts
greatbacon: Yeah, capitalizing makes sense. I'll leave a todo about the enum stuff as well.

greatbacon:
As a note: the bug mentioned in the initial PR was not able to be reproduced by either Ryan or I after a few hours 🤷

PR #110, opened 2021-09-11
change services to have declarative http path and method

ryanatkn:
This continues the work to make services declarative and data-driven. It adds an `http` object to services with `method` and `path`, e.g. `'post'` and `'/api/v1/communities/:community_id'`. The `http` name could maybe be improved, like `route`? The equivalent to `method`+`path` for websockets is simply the service `name` which corresponds to events/rpc/etc.

Since services are now mostly declared as data, it should be fairly trivial to generate some basic docs. Maybe a good first usecase for adding them to the frontend.

### DIFF
--- a/src/lib/server/service.ts
+++ b/src/lib/server/service.ts
@@ -11,8 +11,7 @@ export interface Service<
 	TResponseData extends ServiceResponseData, // TODO change to an output response schema
 > {
 	name: string; // `snake_cased`
-	// TODO better name than `http`? `route`? something else?
-	http: {
+	route: {
 		path: string; // e.g. '/api/v1/some/:neat/:path'
 		// supports each `trouter` http method: https://github.com/lukeed/trouter#method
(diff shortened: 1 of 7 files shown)


src/lib/server/service.ts
```diff
@@ -11,6 +11,12 @@ export interface Service<
 	TResponseData extends ServiceResponseData, // TODO change to an output response schema
 > {
 	name: string; // `snake_cased`
+	// TODO better name than `http`? `route`? something else?
+	http: {
```
greatbacon, line 15: Yeah, I'd use `route` here instead of `http`

greatbacon (approved):

src/lib/server/service.ts
ryanatkn: Cool made that change: https://github.com/feltcoop/felt-server/pull/110/commits/d56c805430952cc74abc7b3736d23495cda669c8

PR #112, opened 2021-09-11
refactor persona and community layout

ryanatkn:
This refactors the community nav and selected persona indicator to align with what we discussed. The idea is that you can see all of your personas and communities at a glance, and select any with one click. This won't scale past a certain number of personas and communities, but we can address that issue down the road.

I'm not super happy with how the elements of the community nav are rendered, but it's flat & simple. Just might not handle the actual usage and styling we want.

- [x] merge #111 and update this

minor things, maybe for the future:

- [ ] fix the single-frame state flash that's caused by updating state in 2 ways, by link navigation and click handler
- [x] style the persona header a bit better

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -40,5 +40,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm ci
+      - run: npx gro lib/db/create
       - run: npx gro check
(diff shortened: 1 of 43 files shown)


greatbacon (approved):
Looking slick! Thanks for putting the whole UX together!

PR #113, opened 2021-09-11
onboard initial persona

ryanatkn:
- [x] do not create an initial persona when creating an account
- [x] add initial onboarding flow with a single step to create a persona
- [x] merge changes from #110 and update


greatbacon (commented):
Looks good to me overall. Couple of thoughts though:

1.) It does seem like there's a UX issue where the account tab starts off activated after login/persona creation and not the community (so it looks like nothing happened).

2.) Could pre-populating the create persona form with the account name make sense? Or should we always make a new user type something for their first persona?

ryanatkn:
1. Looks like it's because the `ui` state is preserved across logouts -- got a hacky fix here: https://github.com/feltcoop/felt-server/pull/113/commits/155e9b8a1f961436c1dd8cc5ded630ffe8ad22f5 -- it probably makes sense to reset state generically on logout. Some UI things might be better UX to preserve but resetting everything will definitely avoid issues, so I'll try to do that before merging.

greatbacon (approved):

PR #114, opened 2021-09-12
add service response schemas and tests

ryanatkn:
- [x] add service response schemas
- [x] add service tests
- [x] rename `service.perform` from `service.handle`
- [x] add db repo tests (single file right now so dependencies can be forwarded implicitly instead of trying to architect a system that keeps things concise in unit tests)
- [x] change many repos to use params objects instead of individual args
- [x] improve schema validation to be a more concise API and faster cache lookup
- [x] add helpers to setup and teardown a server (unfortunately occurs each suite, not sure of a robust way to share a global reference without timeouts -- maybe polling to see if anything is running?)
- [x] add random helpers
- [x] import `unwrap` after releasing next version of Felt

future:

- change repo tests to delete data
- instead of having a single repo test file, architect a system that keeps things concise in unit tests (need good helpers/abstractions for generating random data dependencies)
- generate randomizers automatically from schemas, using a config option to hook into specific properties that may need special sourcing through a randomizer context (e.g. to create random table rows)


src/lib/server/random.ts
```diff
@@ -0,0 +1,61 @@
...
+import type {Service} from '$lib/server/service';
+import {
+	randomFileParams,
+	randomMemberParams,
+	RandomVocab,
+	RandomVocabContext,
+} from '$lib/vocab/random';
+import {randomCommunityParams, randomSpaceParams} from '$lib/vocab/random';
+
+// TODO where does this belong? this isn't technically vocab
```
greatbacon, line 12: This is probably something that could go in `src/lib/util`

greatbacon (approved):
Looks good from the high level perusal I did

src/lib/server/random.ts
ryanatkn: I changed the TODO to consider moving to `src/lib/util`. It was originally with `src/lib/vocab/random.ts`, and I had moved this piece to `src/lib/server/random.ts`, because that's where `services.ts` lives. Maybe it should be `randomService.ts` instead of just `random.ts`? Leaving it where it is for now but it seems likely at least the filename should change.

ryanatkn:
Great, one change I made since you reviewed was to use `unwrap` in the tests more. It extracts a `Result.value` and throws if it's an error result, i.e. `ok: false`. Generally speaking we want to handle those cases explicitly and not throw errors, but for tests, we're mostly duplicating effort. Sometimes you may want to throw an error manually to add more context, but in most cases the underlying error, plus the stack trace to the test, ought to be enough context.

PR #115, opened 2021-09-13
generate server nginx config file

ryanatkn:
This changes the nginx config file to be generated using [`gro gen`](https://github.com/feltcoop/gro/blob/main/src/docs/gen.md). It also renames `$lib/constants.ts` to `$lib/config.js` so `svelte.config.js` can import it -- SvelteKit cannot import TypeScript, but luckily the file is simple enough that we don't lose any meaningful type safety.

- [x] generate `felt-server.conf` nginx config file
- [x] replace `$lib/constants.js` with `$lib/config.ts` and generate `$lib/config.json`
- [x] fix prettier formatting issue (override prettier only for `package.json`)

Note that [`gro gen`](https://github.com/feltcoop/gro/blob/main/src/docs/gen.md) does not (yet) automatically re-run when dependencies change. It wouldn't be hard to implement this in Gro's system, but `gro gen` is expected to have some heavyweight processes, including possibly using external APIs, so for now we must `gro gen` manually. If we forget to do so, `gro check` fails because it calls `gro gen --check`, so CI will go :x: . This also means `gro gen` outputs must be deterministic (though we could add an override if that usecase appears). In [the TODOs section of the gen docs](https://github.com/feltcoop/gro/blob/main/src/docs/gen.md#todo) I propose some API changes to automate this while allowing overrides for heavy processes.


greatbacon (approved):

greatbacon:
So I feel like I missed some details of these new `gro` commands along the way. Did you update the README or CONTRIBUTING docs to cover where these new commands are expected to run in the development pipeline, if they're supposed to be run manually at all?

ryanatkn:
Good idea, I'll add to this PR before merging.

ryanatkn:
Wrote some docs for both `gro gen` and `gro format`, the two manual steps I can think of (outside of db management) -- https://github.com/feltcoop/felt-server/pull/115/commits/290ba8a6a69a94dde54048a473018621bc8c99c8

ryanatkn:
I fixed `.js` imports in Gro (https://github.com/feltcoop/gro/pull/270) so `$lib/config.js` is now the config file instead of TypeScript+generated json. `gro gen` is sometimes very useful but it should be used as little as possible, and happily we can now share `.js` modules between our code and the SvelteKit config.

PR #116, opened 2021-09-13
generate services metadata

ryanatkn:
- [x] generate `$lib/server/servicesMeta.ts` that the client can import
- [x] merge the other PRs and output the response schema

future:

- import the services metadata and display it as docs on the client (where? maybe a new `/docs` route? we discussed this should be a standalone static deployment, but how do we fit that into the existing SvelteKit project?)
- add a form component that lets you automatically call services
- use JSON schema references instead of duplicating child definitions in each


greatbacon (approved):
So this was generated code right? How is it generated again? `gro gen`?

ryanatkn:
Yep that's right, more docs here: https://github.com/feltcoop/gro/blob/main/src/docs/gen.md

CI will fail if the generated output changes, `gro check` includes a call to `gro gen --check`

PR #122, opened 2021-09-15
add home space

ryanatkn:
- [x] add the `Home` space type and make it a default space (so run `gro lib/db/create`)
- [x] extract `PersonaInfo.svelte` and `SpaceInfo.svelte` and reuse them where appropriate
- [x] fix nav to home

next:

- fix `selectPersona`


src/lib/ui/CommunityNav.svelte
```diff
@@ -1,20 +1,27 @@
 <script lang="ts">
 	import type {CommunityModel} from '$lib/vocab/community/community.js';
 	import CommunityInput from '$lib/ui/CommunityInput.svelte';
-	import ActorIcon from '$lib/ui/ActorIcon.svelte';
-	import {randomHue} from '$lib/ui/color';
+	import CommunityNavButton from '$lib/ui/CommunityNavButton.svelte';
 	import type {Persona} from '$lib/vocab/persona/persona';
 
+	// TODO should this just use `ui` instead of taking all of these props?
```
greatbacon, line 7: Probably, this was a first step solution before the UI store was created

src/lib/ui/CommunityNav.svelte
```diff
@@ -1,20 +1,27 @@
...
 	export let communitiesByPersonaId: {
 		[persona_id: number]: CommunityModel[];
 	};
+	export let selectedSpaceIdByCommunity: {[key: number]: number | null};
 	// TODO this is causing a double state change (rendering an invalid in between state)
 	// because it's both navigating and setting state internally in the same user action
 	// TODO should this be an event?
 	export let selectPersona: (persona_id: number) => void;
+
+	// TODO improve the efficiency of this with better
```
greatbacon, line 22: not to be a snotty shit, but `with better` what?

src/lib/ui/CommunityNav.svelte
```diff
@@ -24,26 +31,22 @@
...
-			</a>
+			<CommunityNavButton
+				community={toPersonaCommunity(persona)}
+				{persona}
+				selected={persona === selectedPersona && toPersonaCommunity(persona) === selectedCommunity}
+				{selectedSpaceIdByCommunity}
+				{selectPersona}
+			/>
 			{#each communitiesByPersonaId[persona.persona_id] as community (community.community_id)}
 				{#if community.name !== persona.name}
```
greatbacon, line 42: Since it looks like a lot of the complexity around this button was pushed down into the component, is there a way to refactor this logic branch to be inside the `CommunityNavButton` as well?

src/lib/ui/MainNav.svelte
```diff
@@ -22,7 +23,7 @@
 	$: communitiesByPersonaId = ui.communitiesByPersonaId;
 
 	// TODO refactor to some client view-model for the account
-	$: selectedPersonaName = $selectedPersona?.name || 'guest';
+	$: selectedPersonaName = $selectedPersona?.name || GUEST_PERSONA_NAME;
```
greatbacon, line 26: 🧹

greatbacon (approved):
Just a couple of questions, but it looks good!

src/lib/ui/CommunityNav.svelte
ryanatkn: lol I probably got distracted mid-sentence, "with better data structures and caching"

src/lib/ui/CommunityNav.svelte
ryanatkn: Cool refactored here: https://github.com/feltcoop/felt-server/pull/122/commits/2ebc2c62e6e1829d4e8a3d5225deea08a157635f

src/lib/ui/CommunityNav.svelte
ryanatkn: Yeah good question, it would be better to do so I think, problem is we don't really know what this design should be, and how it should be laid out in the DOM. Right now that quirky complexity is here in the `CommunityNav` and I think it's good to keep it here until we have a better idea of what we want and how to do it.

PR #123, opened 2021-09-15
add api client

ryanatkn:
This is going to be a large PR that abstracts out an `ApiClient` interface (all names TBD) so that frontends can choose between websocket and http clients. Although our app will likely use websockets only, doing this work now ensures that we think through our interfaces before making too many assumptions, and it'll act as a secondary set of code paths that can be automatically tested.

It also implements important features we need for the websocket client, mainly the ability to `await` a service request and receive a response in a promise. Normally websocket messages are fire-and-forget, so this gives us an RPC-like interface to our services over websockets. I'm making the messages conform to the [JSON-RPC 2.0 spec](https://www.jsonrpc.org/specification) because it's simple, fits well enough, and provides grounding for our otherwise bespoke system.

- [x] typesafe `client.invoke`

The `JsonRpcResponse` type may not have the right name, because many incoming messages won't be in response to anything, or maybe it's one of at least 2 types of messages clients can receive (seems more likely). Something to fix in the future I think.


src/lib/vocab/file/fileServices.ts
```diff
@@ -24,7 +24,7 @@ export const readFilesService: Service<
 	name: 'read_files',
 	route: {
 		path: '/api/v1/spaces/:space_id/files',
-		method: 'get',
+		method: 'GET',
```
greatbacon, line 27: Should we add a common set of HTTP method enums to Felt?

greatbacon (approved):
LGTM!

src/lib/vocab/file/fileServices.ts
ryanatkn: I think so, not sure how it'd look. Currently things are semi-documented here:

https://github.com/feltcoop/felt-server/pull/123/files#diff-00b4d1fbea9877f9005ce22f576eb58ecc20d021ba1adb039843999f9152d01aR13

PR #125, opened 2021-09-16
Membership table refactor

greatbacon:
Okay, so I said I was going to split these up but it ended up not being feasible.

Sorry for the xbox huge PR, it should be mostly just renames at this point but some highlights:

1.) Database `persona_communities` table renamed to `memberships`
2.) Repo queries updated to reflect that

and the real doozy in this PR
3.) So it turns out the term "Membership" was overloaded in like 3 places. Now `membership` should only be reserved for describing the specific relationship held between a single `Community` & a single `Persona`.

It's use inside the `Community` object has now been replaced with the more specific `memberPersona`, which describes an actual `Persona` with a `Membership` in the `Community`

It's use inside the frontend UI & Invite flow has been replaced with what that object was actually representing `allPersonas`.


src/lib/vocab/community/community.ts
```diff
@@ -2,14 +2,14 @@ import {Type} from '@sinclair/typebox';
...
-import type {Member} from '$lib/vocab/member/member.js';
+import type {Persona} from '$lib/vocab/persona/persona.js';
 import {toValidateSchema} from '$lib/util/ajv';
 
 export interface Community {
 	community_id: number;
 	name: string;
 	spaces: Space[];
-	members: Member[];
+	member_personas: Persona[];
```
ryanatkn, line 12: This is referenced as `member_personas` here and the community repo, but `memberPersonas` elsewhere.

src/lib/vocab/community/communityRepo.ts
```diff
@@ -54,11 +54,11 @@ export const communityRepo = (db: Database) => ({
 				(
 					SELECT array_to_json(coalesce(array_agg(row_to_json(d)), '{}'))
 					FROM (
-						SELECT p.persona_id, p.name FROM personas p JOIN persona_communities pc ON p.persona_id=pc.persona_id AND pc.community_id=c.community_id
+						SELECT p.persona_id, p.name FROM personas p JOIN memberships m ON p.persona_id=m.persona_id AND m.community_id=c.community_id
 					) d
-				) as members
+				) as member_personas
```
ryanatkn, line 59: Here's the other reference to `member_personas` -- I'm assuming it doesn't need to be quoted in this usage?

src/lib/session/clientSession.ts
```diff
@@ -9,8 +8,8 @@ export interface ClientAccountSession {
 	personas: Persona[];
 	account: AccountModel;
 	communities: Community[];
-	//Stub for a Friends feature in future release, for now just returns all users in an instance
-	members: Member[];
+	//Stub for a Friends feature in future release, for now just returns all personas in an instance
+	allPersonas: Persona[];
```
ryanatkn, line 12: Probably too much to change for this PR, but when we start implementing normalized data it probably makes sense for this to be `personas` and then the currently used `personas` above would instead be a list of ids named something like `accountPersonas`.

src/lib/ui/MemberInput.svelte
```diff
@@ -3,18 +3,18 @@
 	import Markup from '@feltcoop/felt/ui/Markup.svelte';
```
ryanatkn, line 3: Could rename this module if you want.

ryanatkn:
Looks good, only problem I saw was `member_personas`/`memberPersonas` mismatch, commented in the 2 places.

ryanatkn (approved):

src/lib/vocab/community/communityRepo.ts
greatbacon: Ah yeah, it would need to be quoted here. This was a lazy shortcut to get loads from the db working. I'll fix this.

src/lib/session/clientSession.ts
greatbacon: Yeah, my thinking here is that since this is part of a deprecated implementation for the invite feature, we'll be removing this attribute in the near/medium term & should clarify it instead.

PR #126, opened 2021-09-16
add svelte-check to CI

ryanatkn:
context: https://github.com/feltcoop/gro/pull/271

- [x] upgrade to `@feltcoop/gro@0.42.0` which adds `svelte-check` to typechecking as described and implemented here: https://github.com/feltcoop/gro/pull/271
- [x] fix type errors in Svelte files blocking CI

There are some `svelte-check` warnings right now, they're fixed in this PR which will release with `@feltcoop/felt@0.12.0` -- https://github.com/feltcoop/felt/pull/136


greatbacon:
Have you done benchmarking on the potential growth of `gro`s dependencies?

Would this just be a growth in technical complexity, or would there be serious performance implications?

(I have no horse in this race, just trying to help you suss out an answer)

ryanatkn:
I decided it should go into Gro as a regular dependency. Here's the context: https://github.com/feltcoop/gro/pull/271

I think we use Svelte often enough to justify it. Even Node libraries usually use it for frontend docs, so we might as well just bundle everything together in the tool. In the future I'd like to slim down some of its dependencies, maybe upstreaming the changes if they're desired, but it's not worth the time right now.

The downsides:

- more dependencies which means more possible bugs and security issues and more responsibility
- some install time and a little disk space
- Gro feels significantly heavier


Here's the stats:

```
before: 91 dependencies
after: 139 dependencies (48 added)
before: 139.1 MB
after: 145.2 MB (6.1 MB added)
```

ryanatkn:
Added to gro@0.42.0: https://github.com/feltcoop/gro/pull/271/files

PR #127, opened 2021-09-18
add help messages to account and persona forms

ryanatkn:
Adds an informative message to the account and persona creation forms. The goal is to inform the user about the implications of the decision. Might need to include more info.

The idea came from this discussion: https://github.com/feltcoop/felt-server/pull/113#issuecomment-919566336


src/lib/ui/CenteredBlock.svelte
```diff
@@ -0,0 +1,14 @@
+<!-- TODO not sure about this component -->
```
greatbacon, line 1: This seems a little like overkill to me (assuming that's what this TODO is about)

src/lib/ui/PersonaInput.svelte
```diff
@@ -47,6 +49,9 @@
 		<button type="button" on:click={create} disabled={status === 'pending'}> Create </button>
 	</form>
 </Markup>
+<CenteredBlock>
+	<Message icon="‼">your persona name is visible in your communities</Message>
```
greatbacon, line 53:
```suggestion
	<Message icon="‼">your persona name is visible to others</Message>
```
Not 100% on if this is the right phrase, but the other one felt a little awkward

greatbacon (approved):

src/lib/ui/CenteredBlock.svelte
ryanatkn: Yeah alternatives would be as a class, either global or internal to the component. I'll remove this component.

PR #128, opened 2021-09-18
add marquee

ryanatkn:
Adds the "social sidebar" aka `Marquee.svelte` to the workspace to the right of the active space view. Currently only displays the community's members and has `MarqueeNav.svelte`. It contains meta info about the active community and space.

- [x] display the selected space's personas in the right sidebar
- [x] rename `PersonaInfo` to `Avatar` which I think is :fire: (usable for both personas and communities and anything with a `name` and/or `icon`)
- [x] add the `entity` vocabulary term and add helpers to extract `name` and `icon` from data that may or may not have those properties (see its file for possible renames)
- [x] add a toggle to the space meta nav, see `expandSecondaryNav`/`toggleSecondaryNav` (and maybe rename/refactor -- it's similar to `expandMainNav`)
- [x] fix space meta nav for mobile

future:

- resolve some differences between how the space meta panel works compared to the main menu (e.g. it appears instantly instead of sliding in, maybe we want to use a shared abstraction)
- add interactivity to the persona avatars (context menu)
- display space's members instead of the community's


src/lib/ui/SpaceMeta.svelte
```diff
@@ -0,0 +1,32 @@
...
+	import type {Persona} from '$lib/vocab/persona/persona';
+	import type {Community} from '$lib/vocab/community/community';
+	import Avatar from '$lib/ui/Avatar.svelte';
+	import SpaceMetaNav from '$lib/ui/SpaceMetaNav.svelte';
+	import {toIcon, toName} from '$lib/vocab/entity/entity';
+	import {getApp} from '$lib/ui/app';
+
+	const {ui} = getApp();
+
+	// TODO better name than `SpaceMeta`?
```
greatbacon, line 12: If nothing else, shouldn't this be `CommunityMeta` since it describes Community level info, not Space level info? At least I think that was how you were describing it this morning...

greatbacon: Maybe could refer to it as the `Marquee`,`Billboard`, or `Bulleteinboard` for a community?

greatbacon (approved):

src/lib/ui/SpaceMeta.svelte
ryanatkn: :laughing:  revive [the Marquee](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee)??

ryanatkn: It's both `CommunityMeta` and `SpaceMeta` I think -- like the list of members should be scoped to just the space, once we support that feature. You can also picture the space overriding the community-wide rules, or extending them.

ryanatkn: I'll keep it as `SpaceMeta` for now since it's an obviously bad name. Added it to our private vocabulary document for discussion.

PR #129, opened 2021-09-18
add persona name to query params

ryanatkn:
(extends #122 so currently has a lot of noise)

This adds `persona=${persona.name}` to the query params, e.g. `felt.social/dev?persona=alice`.

I think this is a good enough pattern for now. The idea is that the source of truth is the URL and the query params encodes path-orthogonal state like the active persona name. An alternative would be to drive the URL from our internal system events but it seems more complicated because it needs to be hydrated on load.

- will the server have wasteful cache misses because of the query params littered with unused values?


ryanatkn:
re: copypasting links could expose something you meant to keep private -- what if we used the persona's index instead? The server would probably have to ensure a stable order. It could be ignored if it overflowed the session's available personas.

greatbacon (approved):

PR #130, opened 2021-09-20
Added url uniqueness validation on create

greatbacon:
Added a check to validate URL uniqueness on constraint.

I got it working on the backend, but I wasn't sure how to take the error the API throws (`api.ts` line 191) and make it show up as a toast or temporary message to the user. Any thoughts @ryanatkn ?


ryanatkn:
For the error message, let's add it to the backlog to display error messages on the community and space input forms, inline like with the login form. Notifications/toasts I think we should punt further. #123 updates the client API to return errors in a structured way, so let's mark on the roadmap to add these error messages after that's in.

src/lib/vocab/space/spaceRepo.ts
```diff
@@ -24,13 +24,26 @@ export const spaceRepo = (db: Database) => ({
...
 		const data = await db.sql<Space[]>`
       SELECT s.space_id, s.name, s.url, s.media_type, s.content FROM spaces s JOIN community_spaces cs ON s.space_id=cs.space_id AND cs.community_id= ${community_id}
     `;
 		console.log('[db] spaces data', data);
 		return {ok: true, value: data};
 	},
+	filterByCommunityUrl: async (
+		community_id: number,
+		url: string,
+	): Promise<Result<{value: Space[]}>> => {
```
ryanatkn, line 37: I think we should change this to `findByCommunityUrl` instead of `filter` and return `Space | undefined` instead of `Space[]`. The return statement can be changed to `return {ok: true, value: data[0]};` and it'll be `undefined` if not there, just like a JS array `find` call.

src/lib/vocab/space/spaceServices.ts
```diff
@@ -129,9 +129,23 @@ export const createSpaceService: Service<
...
+			params.url,
+		);
 
-		console.log('[create_space] creating space for community', params.community_id);
+		if (!filterByCommunityUrlResult.ok) {
+			console.log('[create_space] error validating unique url for new space');
+			return {code: 500, data: {reason: 'error validating unique url for new space'}};
+		}
 
+		if (filterByCommunityUrlResult.value.length != 0) {
```
ryanatkn, line 143:
```suggestion
		if (filterByCommunityUrlResult.value.length !== 0) {
```

ryanatkn (approved):

src/lib/vocab/space/spaceServices.ts
greatbacon: Ah, my ☕  is showing 😳

greatbacon:
Thanks for the tips on how to do the error messaging! I'll just merge this in & then do the error message in a follow up PR

ryanatkn:
I'm confused as to why CI is failing; I can't reproduce locally.

ryanatkn:
So it appears GitHub runs CI on the merged to main version, which is totally news to me. To fix:

```bash
git merge main
```

Then go to `spaceServices.ts` and you should see the error message in your editor. The lines that have `code` and `data` need to be changed to `status` and `value`.

PR #134, opened 2021-09-25
pass through api result from services to clients

ryanatkn:
Extracts the `ApiResult` type to `$lib/server/api.ts` and uses it on both the client and server:

- http service middleware uses http/`globalThis.fetch` conventions for the `status` and `ok`
- websocket middleware sends the unmodified `Result` object (for now, we may need to add more features in the future)

The `ApiResult` is now the return value of each `Service`'s `perform` function. It's then passed through to clients, rather than being a bespoke form of data+error.

### DIFF
--- a/src/lib/server/serviceMiddleware.ts
+++ b/src/lib/server/serviceMiddleware.ts
@@ -57,7 +57,7 @@ export const toServiceMiddleware =
 			}
 			const result = await service.perform({server, params, account_id: req.account_id});
 			if (!result.ok) {
-				send(res, result.status || 500, {reason: result.reason}); // TODO consider returning the entire `result` for convenience (but it's less efficient)
+				send(res, result.status || 500, {reason: result.reason});
 				return;
 			}
 			if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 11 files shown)


src/lib/server/serviceMiddleware.ts
```diff
@@ -55,14 +55,18 @@ export const toServiceMiddleware =
...
-				if (!service.validateResponse()(response.value)) {
-					console.error(red('validation failed:'), response, service.validateResponse().errors);
+				if (!service.validateResponse()(result.value)) {
+					console.error(red('validation failed:'), result, service.validateResponse().errors);
 				}
 			}
-			console.log('[serviceMiddleware] result.code', response.status);
-			send(res, response.status, response.value);
+			console.log('[serviceMiddleware] result.status', result.status);
+			send(res, result.status, result.value); // TODO consider returning the entire `result` for convenience (but it's less efficient)
```
greatbacon, line 69: We chatted about this & just returning `result.value` is fine

greatbacon (approved):

PR #135, opened 2021-09-25
make the api randomly choose a client on every request

ryanatkn:
Makes the client api randomly choose between http and websockets on every request. Might surface bugs! (made me realize http service middleware needs to broadcast just like websockets). We'll remove this after adding tests.


greatbacon (approved):
🎲

PR #140, opened 2021-09-27
show different icons for users and communities

ryanatkn:
This makes the default icons different for communities and personas -- communities now have a rounded rectangle, and personas remain circles.

Adds `EntityType` that can currently be either `'Persona'` or `'Community'` to a `$lib/vocab/entity/entity.ts`.

It also changes `ActorIcon.svelte` to be `EntityIcon.svelte`, since we can assume consistent semantics for any ActivityStreams object -- the interface is basically just a `name` and `icon`, which many things can have. Additional type-specific (like "all actors should look like this") can be implemented here, either directly in the component or through composition externally.


greatbacon:
Ah, I think I had something a little different in mind when we had chatted about this? I was envision the shape of the color changing (i.e. circles of color for Personas & squares for Communities) but still leaving the first letter in the middle.

ryanatkn:
Ohh right my mistake, I forgot that it was border radius we decided to differentiate on. I implemented it now -- I think it looks pretty good!

greatbacon:
Sweet, I agree that it looks nice! Two more thoughts; is a Persona's Home Community a `Community` or a `Persona`? Since right now it's shaped like a `Persona`, but I expected it to be a `Community` shape.

Either way, I also noticed that the icon at the top of the `Marquee` is displaying as a `Persona` even if it's a standard `Community`.

greatbacon (approved):

PR #146, opened 2021-09-29
Fix readable spaces

greatbacon:
Pair programming PR that fixes the type of the selectedSpace store


ryanatkn (approved):
# :space_invader:

PR #147, opened 2021-09-29
Add timestamps to accounts

greatbacon:
This PR adds a created & updated field to the Accounts model in the DB.

It also modified some of the schema & validation stuff to prevent the `password` attribute from leaking in DB queries.

@ryanatkn I have an issue though; I mapped things through in the Typebox Schema validation using string, the underlying object seems to be a `Date` type though & that doesn't exist in Typebox. What should I do?


greatbacon:
Hey @ryanatkn just wanted to follow up on this PR. Any thoughts on the issues with Types?

ryanatkn:
I was also hitting Typebox limitations due to its constraint of generating good runtime types. This constraint makes it unwieldy in some cases to generate arbitrary JSON schema, so in this PR I removed it as a dependency: https://github.com/feltcoop/felt-server/pull/152

We'll want to update this PR after merging #154

greatbacon:
Closing in favor of #155

PR #149, opened 2021-09-30
add generic dispatch function to client api and ui

ryanatkn:
This is a proof of concept adding a generic dispatch function for events on the client.

It changes one component call from:

```ts
const result = await api.createCommunity({name, persona_id: $selectedPersonaId!});
```

to

```ts
const result = await dispatch('create_community', {name, persona_id: $selectedPersonaId!});
```

This is the same event name and params that the server's services use. The `dispatch` function is a single code path that can handle all client events/commands. The service events are the same as the ones the client uses, and the client will add its own client-only events as well.

Before, to implement an event on the client, we had 2 bespoke methods, one on the `api` (`createCommunity`) and one on the `ui` (`addCommunity`). After this change, the `api` is generic and the `ui` has a handler named the same as the event, `create_community`, which is called by `dispatch`.

To support this, I added the parameter `findService` and method `has(serviceName)` to the API clients, so we can automatically detect if a particular event has a server-side implementation -- if not, it's a client-side only event.

The API is still in flux. The `create_community` usecase isn't enough to flesh out the API. See #150 for the full implementation. The system should support client-side-only events (like toggling a menu), synchronous calls (not forced into a promise wrapper), arbitrary return values (like a reactive store containing posts matching a query), and composition (events dispatching events).

In the near term this will unlock:

- generated documentation of all events in the system, a reference of all of the things you can "do" in the UI and those inputs/outputs
- logging and introspecting every event that flows through the client through a single code path (in many ways makes it easier to observe, understand, debug, and modify the system)
- simple automated tests (we do this with services already, and now we could do it on the client too)
- making the API client generic, deleting its bespoke methods
- aligning and integrating the client with the existing server-side services system, reducing the API surface area so there's less to keep in our heads and there's no translation between layers -- the `create_community` event is defined once and its name/inputs/outputs are reused everywhere that makes sense, both client and server

and in the medium/long term:

- automatically generating forms and GUI for creating/dispatching events (enabling low-level "pop the hood" controls that get the same capabilities as the rest of the UI, implemented generically)
- more powerful automated tests (there's a lot here -- first we'll get the low-hanging fruit in the near term, and long-term the schemas enable things like property-based, fuzz, and procedurally generated tests)
- recording/crafting event streams for client automation (scripts/macros)
- GUI-editable scripts that can be shared securely (this isn't without its security problems, but it unlocks scripting the system without arbitrary code execution)

Further work is in #150 to handle all of our usecases -- this one only implements `create_community` as a proof of concept.


src/lib/ui/ApiClient.ts
```diff
@@ -1,10 +1,10 @@
 import type {ApiResult} from '$lib/server/api';
 
 export interface ApiClient<
-	TParamsMap extends Record<string, any> = any, // TODO default type?
-	TResultMap extends Record<string, any> = any, // TODO default type?
+	TParamsMap extends Record<string, any> = any, // TODO default and value types?
+	TResultMap extends Record<string, any> = any, // TODO default and value types?
 > {
-	// TODO `name`?
+	invokes: (name: string) => boolean;
```
ryanatkn, line 7: Placeholder name -- maybe `knows` or `handles` or `has` or something? Maybe suffixed with `Event`?

greatbacon (approved):

PR #155, opened 2021-10-04
Update account and crew with created/updated timestamps

greatbacon:
Second shot at adding timestamps, now with refactor goodness missing from #147

### DIFF
--- a/src/lib/vocab/account/account.ts
+++ b/src/lib/vocab/account/account.ts
@@ -2,6 +2,8 @@ export interface Account {
 	account_id: number;
 	name: string;
 	password: string;
+	created: Date;
+	updated?: Date;
 }
 export const AccountSchema = {
 	$id: 'https://felt.social/vocab/Account.json',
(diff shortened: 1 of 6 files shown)


greatbacon:
Hey @ryanatkn looks like there are still schema/type issues. If I define the `created` in the account schema as a `string` I still get the runtime test failure where it collides with `Date`. But then if I try to define it as a `Date` in the schema it complains about the type.  Any thoughts?

src/lib/vocab/account/account.ts
```diff
@@ -2,6 +2,8 @@ export interface Account {
 	account_id: number;
 	name: string;
 	password: string;
+	created: Date;
+	updated?: Date;
```
ryanatkn, line 6: One temporary issue is that these will be strings on the client -- we'll just `ajv` to parse/serialize using the schemas: https://ajv.js.org/guide/getting-started.html#parsing-and-serializing-json

ryanatkn: Also see this comment, I think `updated` should just be nullable not optional -- https://github.com/feltcoop/felt-server/pull/158/files#r722839862

PR #157, opened 2021-10-05
add initial vocab docs

ryanatkn:
Extends #156 with docs for the vocab schemas. Adds the `ui/SchemaInfo.svelte` component for displaying basic schema info. 

Also adds the `$lib/vocab/schema` directory and relocates the `schemas` module here from `$lib/vocab/schemas.ts`. I added tests to make sure the `$id` property of each vocab schema has the expected prefix. (currently, `'https://felt.social/vocab'`)

### DIFF
--- a/src/lib/ui/MainNav.svelte
+++ b/src/lib/ui/MainNav.svelte
@@ -22,6 +22,7 @@
 			selectedPersona: selectedPersonaStore,
 			selectedCommunity: selectedCommunityStore,
 		},
+		devmode,
 	} = getApp();
 
 	$: selectedPersona = $selectedPersonaStore!; // TODO type?
@@ -71,6 +72,9 @@
 		{:else if $mainNavView === 'account'}
 			<Markup>
 				<AccountForm guest={$session.guest} />
+				{#if $devmode}
+					<a class="menu-link" href="/docs">/docs</a>
+				{/if}
 			</Markup>
 			<SocketConnection />
 		{/if}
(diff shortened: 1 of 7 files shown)


src/lib/vocab/entity/entities.ts
```diff
@@ -1,18 +1,15 @@
...
 
-// TODO params schemas are duplicated from events -- should use them?
-// or is there no such thing as these params schemas? always contextual to an event?
-
-// TODO what about other schemas like for events?
-
 // TODO generate this file?
 
-export const schemas = [
+export const entities: AnySchema[] = [
```
ryanatkn, line 12: I'm not sure about this file rename. The module as implemented looks right, but separating the entity schemas from event schemas is questionable. We may actually want `vocab/schemas.ts` as the index for all schemas. It might depend on how exactly we define events -- are they also entities? What then are non-event vocab terms called?

src/lib/vocab/util.ts
```diff
@@ -0,0 +1,11 @@
+import {stripEnd, stripStart} from '@feltcoop/felt/util/string.js';
+
+export const ID_PREFIX = 'https://felt.social';
+export const ID_VOCAB_PREFIX = ID_PREFIX + '/vocab/';
+export const ID_SUFFIX = '.json';
+
+const names: Record<string, string | undefined> = {};
+
+// TODO maybe cache `title` on each schema?
+export const toSchemaName = ($id: string): string =>
```
ryanatkn, line 10: We probably want to put this on each schema as the `title` attribute. It's the last segment of the `$id` url minus the `.json` suffix.

PR #158, opened 2021-10-05
Add personas & communities timestamp

greatbacon:
This adds created & updated attributes to the personas & communities tables & vocab interfaces.

This PR also added the ajv-formats addon for our schema validation & moved the deploy task back to root to overwrite the default task.


src/lib/util/ajv.ts
```diff
@@ -9,6 +10,7 @@ let ajvInstance: Ajv | null = null;
 export const ajv = (): Ajv => {
 	if (ajvInstance) return ajvInstance;
 	ajvInstance = new Ajv();
+	addFormats(ajvInstance);
```
ryanatkn, line 13: # :ok_hand:

src/lib/vocab/persona/persona.ts
```diff
@@ -4,6 +4,8 @@ export interface Persona {
 	name: string;
 	icon?: string;
 	community_ids: number[];
+	created: Date;
+	updated?: Date;
```
ryanatkn, line 8: I think this should be non-optional, so `updated: Date | null;`, also adding to `required` in the schema.

src/lib/ui/LogoutForm.svelte
```diff
@@ -30,6 +32,9 @@
 	};
 </script>
 
+{#if $selectedPersona}
+	<div>This persona was created {get($selectedPersona).created}</div>
```
ryanatkn, line 36: happy not-cake-something-else day?

ryanatkn: This isn't a bug because `persona.created` will never change, but `get` isn't reactive so the more idiomatic way to do this would be to change this expression:

```svelte
{get($selectedPersona)}
```

to

```svelte
$: selectedPersonaValue = $selectedPersona; // or named whatever
```

Aside: we talked renaming selection stores so this would be `selectedPersona = $personaSelection`. Probably makes sense to do a PR that does them all at once. (also allow removing the `Store` suffixes you see in other components)

greatbacon: Happy 🌱  day!

greatbacon: Or 🎉  day!

PR #174, opened 2021-11-01
Add a rough hack/first pass at getting the event broadcast architectu…

greatbacon:
…re in line

From pair-programming this afternoon


ryanatkn (approved):
# :axe:  

(best hack emoji I could think of)

PR #176, opened 2021-11-05
increase nginx proxy timeout

ryanatkn:
Every so often, the websocket disconnects. It seems likely caused by nginx having a 60 second timeout after no activity.

This change increases its timeout to 6 minutes per this answer: https://stackoverflow.com/a/57706152

And the client will send a `ping` event every 5 minutes to keep the connection alive. The linked answer demonstrates 24 hours but that means the server would hold onto dropped connections far too long.

- [x] decide nginx timeout duration: 6 minutes
- [x] add 5 minute interval client-initiated heartbeat

Future:

- move logic to websocket store

### DIFF
--- a/src/infra/felt-server.conf
+++ b/src/infra/felt-server.conf
@@ -12,8 +12,8 @@ server {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $host;
     proxy_pass http://localhost:3000;
-    proxy_read_timeout 360s;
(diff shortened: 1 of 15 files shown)


greatbacon (dismissed):
Maybe at some point we'll want to push it down to maybe like an hour or 30 minutes at some point, but I think this works for now.

ryanatkn:
I'm going to do a followup to refactor, not happy with the modules and structure.

greatbacon (approved):

PR #177, opened 2021-11-08
Update deploy docs

greatbacon:
This PR updates documentation around the deployment process as well as fine tuning & testing of deployment scripts in a new cloud context.


src/infra/setup.task.ts
```diff
@@ -1,24 +1,31 @@
...
+			apt install unzip;			
+		  curl -fsSL https://fnm.vercel.app/install | bash;`,
 		]);
 		//Splitting these tasks here let's fnm get picked up from the bash profile
-		await spawn('ssh', [deployLogin, `fnm install 16;`]);
+		await spawn('ssh', [
+			deployLogin,
+			`/root/.fnm/fnm install 16;
+			apt install npm;
+			sudo npm install pm2@latest -g`,
```
ryanatkn, line 27: Is this `sudo` necessary here? If so I think it'll help to leave a TODO here like `// TODO remove sudo after changing deployment to run as user instead of root`. We'll want to do this for security at some point before the beta.

src/lib/config.js
```diff
@@ -8,9 +8,10 @@
 // This probably means supporting `import.meta.env` or `$app/env` in Gro.
 const dev = import.meta.env?.DEV ?? process.env.NODE_ENV !== 'production';
 
-export const DEPLOY_SERVER_HOST = 'staging.felt.dev';
-export const DEPLOY_IP = '96.126.116.174';
-export const DEPLOY_USER = 'root';
+export const DEPLOY_SERVER_HOST = '<DEPLOY_URL_GOES_HERE>';
+export const DEPLOY_IP = '<DEPLOY_IP_GOES_HERE>';
+export const DEPLOY_USER = '<DEPLOY_USER_GOES_HERE>';
+export const EMAIL_ADDRESS = '<DEPLOY_EMAIL_GOES_HERE>';
```
ryanatkn, line 15: Quick fix is to do

```ts
export const DEPLOY_SERVER_HOST = '<DEPLOY_URL_GOES_HERE>';
```

to:

```ts
export const DEPLOY_SERVER_HOST = process.env.DEPLOY_SERVER_HOST;
```

more correctly, but only works in `.ts` files, so we can't do this in `src/config.js` right now:

```ts
export const DEPLOY_SERVER_HOST = fromEnv('DEPLOY_SERVER_HOST');
```

package.json
```diff
@@ -12,7 +12,8 @@
     "dev": "gro dev",
     "build": "gro build",
     "test": "gro test",
-    "deploy": "gro deploy"
+    "deploy": "gro deploy",
+    "start": "node dist/server/lib/server/server.js"
```
ryanatkn, line 16: :+1:

ryanatkn (approved):

src/infra/setup.task.ts
greatbacon: great question! I'll do a little testing

greatbacon: Alright, sudo removed

src/deploy.task.ts
```diff
@@ -1,19 +1,21 @@
...
-import {DEPLOY_IP, DEPLOY_USER} from '$lib/config';
-
-const deployLogin = `${DEPLOY_USER}@${DEPLOY_IP}`;
+import {fromEnv} from '$lib/server/env';
 
 export const task: Task = {
 	summary: 'deploy felt server to prod',
 	dev: false,
 	run: async ({invokeTask}) => {
+		const DEPLOY_IP = fromEnv('DEPLOY_IP');
```
greatbacon, line 10: Hey @ryanatkn couple of questions. I thought I had this set up properly, but when I go to run `gro deploy` it is running in dev mode (not production). Any thoughts?

src/lib/config.js
```diff
@@ -7,10 +7,7 @@
 // TODO Replace this at buildtime with a constant so it can be optimized.
 // This probably means supporting `import.meta.env` or `$app/env` in Gro.
 const dev = import.meta.env?.DEV ?? process.env.NODE_ENV !== 'production';
-
-export const DEPLOY_SERVER_HOST = 'staging.felt.dev';
-export const DEPLOY_IP = '96.126.116.174';
-export const DEPLOY_USER = 'root';
+export const DEPLOY_SERVER_HOST = process.env.DEPLOY_SERVER_HOST;
```
greatbacon, line 10: @ryanatkn Other thing is that when I go to run locally now, when the code gets to this on page load I get a `Uncaught (in promise) ReferenceError: process is not defined` any thoughts?

greatbacon: Nvm, got this one sorted out

src/deploy.task.ts
ryanatkn: Sounds like a Gro bug, the `dev: false,` line here should set production mode if `process.env.NODE_ENV` isn't set. I'll take a look.

PR #182, opened 2021-12-03
Delete spaces

greatbacon:
This PR adds server side logic for deleting a space as well as a first pass at front end UI & state management.

Still a bit of a WIP, I have one particular issue I need help with, which I will call out in line.

### DIFF
--- a/src/lib/ui/SpaceDelete.svelte
+++ b/src/lib/ui/SpaceDelete.svelte
@@ -4,14 +4,12 @@
 
 	import {getApp} from '$lib/ui/app';
 	import type {Space} from '$lib/vocab/space/space';
-	import type {Community} from '$lib/vocab/community/community';
 
 	const {
 		api: {dispatch},
 	} = getApp();
 
 	export let space: Readable<Space>;
-	export let community: Readable<Community>;
 
 	let opened = false;
 	let errorMessage: string | undefined;
@@ -20,7 +18,6 @@
 		errorMessage = '';
 		const result = await dispatch('delete_space', {
 			space_id: $space.space_id,
-			community_id: $community.community_id,
 		});
 		if (result.ok) {
 			opened = false;
--- a/src/lib/ui/SpaceDelete.svelte
+++ b/src/lib/ui/SpaceDelete.svelte
@@ -1,0 +1,67 @@
+<script lang="ts">
+	import Dialog from '@feltcoop/felt/ui/Dialog.svelte';
+	import type {Readable} from 'svelte/store';
+
+	import {getApp} from '$lib/ui/app';
+	import type {Space} from '$lib/vocab/space/space';
+
+	const {
+		api: {dispatch},
+	} = getApp();
+
+	export let selectedSpace: Readable<Space>;
+
+	let opened = false;
+	let errorMessage: string | undefined;
+
+	const deleteSpace = async () => {
+		errorMessage = '';
+		const result = await dispatch('delete_space', {space_id: $selectedSpace.space_id});
+		if (result.ok) {
+			opened = false;
+		} else {
+			errorMessage = result.reason;
+		}
+	};
+
+	const onKeydown = async (e: KeyboardEvent) => {
+		if (e.key === 'Enter') {
+			e.preventDefault();
+			await deleteSpace();
+		}
+	};
+</script>
+
+<button
+	aria-label="Delete Space"
+	type="button"
+	class="button-emoji"
+	on:click={() => (opened = true)}
+>
+	🗑️
+</button>
+{#if opened}
+	<Dialog on:close={() => (opened = false)}>
+		<div class="markup">
+			<h1>Delete {$selectedSpace.name} space?</h1>
+			<form>
+				<div class:error={!!errorMessage}>{errorMessage || ''}</div>
+				<button type="button" on:click={deleteSpace} on:keydown={onKeydown}> Delete space </button>
+			</form>
+		</div>
+	</Dialog>
+{/if}
+
+<style>
+	.error {
+		font-weight: bold;
+		color: rgb(73, 84, 153);
+	}
+	.button-emoji {
+		background: none;
+		border: none;
+		cursor: pointer;
+		margin: 0;
+		word-wrap: break-word;
+	}
+</style>
--- a/src/lib/ui/SpaceDelete.svelte
+++ b/src/lib/ui/SpaceDelete.svelte
@@ -10,7 +10,7 @@
 		api: {dispatch},
 	} = getApp();
 
-	export let selectedSpace: Readable<Space>;
+	export let space: Readable<Space>;
 	export let community: Readable<Community>;
 
 	let opened = false;
@@ -19,7 +19,7 @@
 	const deleteSpace = async () => {
 		errorMessage = '';
 		const result = await dispatch('delete_space', {
-			space_id: $selectedSpace.space_id,
+			space_id: $space.space_id,
 			community_id: $community.community_id,
 		});
 		if (result.ok) {
@@ -48,7 +48,7 @@
 {#if opened}
 	<Dialog on:close={() => (opened = false)}>
 		<div class="markup">
-			<h1>Delete {$selectedSpace.name} space?</h1>
+			<h1>Delete {$space.name} space?</h1>
 			<form>
 				<div class:error={!!errorMessage}>{errorMessage || ''}</div>
 				<button type="button" on:click={deleteSpace} on:keydown={onKeydown}> Delete space </button>
--- a/src/lib/ui/SpaceDelete.svelte
+++ b/src/lib/ui/SpaceDelete.svelte
@@ -4,19 +4,24 @@
 
 	import {getApp} from '$lib/ui/app';
 	import type {Space} from '$lib/vocab/space/space';
+	import type {Community} from '$lib/vocab/community/community';
 
 	const {
 		api: {dispatch},
 	} = getApp();
 
 	export let selectedSpace: Readable<Space>;
+	export let community: Readable<Community>;
 
 	let opened = false;
 	let errorMessage: string | undefined;
 
 	const deleteSpace = async () => {
 		errorMessage = '';
-		const result = await dispatch('delete_space', {space_id: $selectedSpace.space_id});
+		const result = await dispatch('delete_space', {
+			space_id: $selectedSpace.space_id,
+			community_id: $community.community_id,
+		});
 		if (result.ok) {
 			opened = false;
 		} else {
--- a/src/lib/ui/ui.ts
+++ b/src/lib/ui/ui.ts
@@ -408,18 +408,23 @@ export const toUi = (session: Writable<ClientSession>, initialMobile: boolean): 
 			const result = await invoke();
 			if (!result.ok) return result;
 			//update state here
-			const {space_id, community_id} = params;
-			const community = get(communities).find((c) => get(c).community_id === community_id)!;
-			community.update(($community) => ({
-				...$community,
-				// TODO clean this up as part of the data normalization efforts
-				spaces: $community.spaces.filter((space) => space.space_id !== space_id),
-			}));
-			spaces.update(($spaces) => $spaces.filter((space) => get(space).space_id !== space_id));
-			dispatch('select_space', {
-				community_id: community_id,
-				space_id: get(community).spaces[0].space_id,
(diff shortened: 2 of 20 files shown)


src/lib/ui/ui.ts
```diff
@@ -400,9 +400,25 @@ export const toUi = (session: Writable<ClientSession>, initialMobile: boolean):
...
+			//update state here
+			const {space_id, community_id} = params;
+			const community = get(communities).find((c) => get(c).community_id === community_id)!;
+			community.update(($community) => ({
+				...$community,
+				// TODO clean this up as part of the data normalization efforts
+				spaces: $community.spaces.filter((space) => space.space_id != space_id),
+			}));
+			spaces.update(($spaces) => $spaces.filter((space) => get(space).space_id != space_id));
+			select_space({community_id: community_id, space_id: get(community).spaces[0]});
```
greatbacon, line 419: So (I think) I've got the logic sorted out for removing a deleted space from the community & space stores. I also want to update the selected space so that the UI doesn't hang on a deleted space & instead moves to a still extant space. I'm trying that there with the `select_space` line, but I guess I can't access that from within a sibling call? Any thoughts here @ryanatkn ?

src/lib/ui/SpaceDelete.svelte
```diff
@@ -0,0 +1,72 @@
...
+
+	import {getApp} from '$lib/ui/app';
+	import type {Space} from '$lib/vocab/space/space';
+	import type {Community} from '$lib/vocab/community/community';
+
+	const {
+		api: {dispatch},
+	} = getApp();
+
+	export let selectedSpace: Readable<Space>;
```
ryanatkn, line 13: Small suggestion, I think this should be named `space` instead of `selectedSpace`.

src/lib/ui/ui.ts
ryanatkn: The `ui` local var can be used to reach the sibling methods, so `ui.select_space`. I think the right logic is to loop through each of `selectedSpaceIdByCommunity` and call `ui.select_space` for any of the `space_id`s matching the one in the params. The loop is needed because spaces do not belong to a single community.

src/lib/ui/ui.ts
```diff
@@ -400,9 +400,25 @@ export const toUi = (session: Writable<ClientSession>, initialMobile: boolean):
...
+		delete_space: async ({params, invoke}) => {
+			const result = await invoke();
+			if (!result.ok) return result;
+			//update state here
+			const {space_id, community_id} = params;
+			const community = get(communities).find((c) => get(c).community_id === community_id)!;
+			community.update(($community) => ({
+				...$community,
+				// TODO clean this up as part of the data normalization efforts
+				spaces: $community.spaces.filter((space) => space.space_id != space_id),
```
ryanatkn, line 416: Minor nitpick - we probably should add `eslint` soon to get these kinds of things automatically as we're coding -- there's 4 instances of using `!=` in the PR instead of `!==`.

src/lib/ui/SpaceDelete.svelte
```diff
@@ -0,0 +1,72 @@
...
+	export let community: Readable<Community>;
+
+	let opened = false;
+	let errorMessage: string | undefined;
+
+	const deleteSpace = async () => {
+		errorMessage = '';
```
ryanatkn, line 20: I think it'd be a fun refactor together to improve this code and make it more reusable, when we have an appetite for improving some UI patterns. Just off the top of my head:

- extract a generic component for the modal trigger, making the delete space form usable in other layouts
- use the `PendingButton` 
- standardize form behavior
  - async state (to do pending button automatically)
  -  error messages
  - using `Enter` to proceed generically through a form

src/lib/ui/SpaceDelete.svelte
```diff
@@ -0,0 +1,72 @@
...
+	type="button"
+	class="button-emoji"
+	on:click={() => (opened = true)}
+>
+	🗑️
+</button>
+{#if opened}
+	<Dialog on:close={() => (opened = false)}>
+		<div class="markup">
+			<h1>Delete {$selectedSpace.name} space?</h1>
```
ryanatkn, line 51: This might be a good candidate for mounting a generic component like `SpaceInfo` -- it's not very useful yet but I think it'll be good to use it in situations like these to generate ideas. The `h1` could maybe then be generic and slightly clearer, "Delete this space?", but feel free to ignore this comment.

src/lib/vocab/space/spaceRepo.ts
```diff
@@ -80,4 +80,33 @@ export const spaceRepo = (db: Database) => ({
...
+
+		const data = await db.sql<any[]>`
+      DELETE FROM spaces WHERE ${space_id}=space_id
+    `;
+
+		if (data.count != 1) {
+			return {
+				ok: false,
+				type: 'deletion_error',
+				reason: `There was an issue deleting community_space: ${space_id}`,
```
ryanatkn, line 106: copypaste typo, `deleting community_space` should be `deleting space`

ryanatkn (approved):

src/lib/vocab/space/spaceRepo.ts
```diff
@@ -80,4 +80,33 @@ export const spaceRepo = (db: Database) => ({
...
 		return {ok: true, value: spaces};
 	},
+	deleteById: async (
+		space_id: number,
+	): Promise<Result<{value: any[]}, {type: 'deletion_error'} & ErrorResponse>> => {
+		const data_cs = await db.sql<any[]>`
+      DELETE FROM community_spaces WHERE ${space_id}=space_id
+    `;
+
+		if (data_cs.count != 1) {
```
ryanatkn, line 90: To support multiples i think this should be:
```suggestion
		if (data_cs.count === 0) {
```

src/lib/vocab/space/spaceRepo.ts
```diff
@@ -80,4 +80,33 @@ export const spaceRepo = (db: Database) => ({
...
+			return {
+				ok: false,
+				type: 'deletion_error',
+				reason: `There was an issue deleting community_space: ${space_id}`,
+			};
+		}
+
+		const data = await db.sql<any[]>`
```
ryanatkn, line 98: To handle spaces across communities, I think the delete space SQL should be executed only if there's 0 `community_spaces` remaining for this space.

Side note: we'll see more of this kind of complexity going forward because we handle spaces and communities as M:N. Like everything else it's open for reconsideration, but the benefits appear substantial for exploring cross-community governance.

ryanatkn:
This may be something to change in a followup PR:

There's a minor inconsistency where the `params` for `delete_space` include `community_id` and `space_id`, but `community_id` is not used on the server. Essentially that means it's a hard delete; in the repo method everything for the space is deleted, so if you're invited to collaborate on a document, and when you "delete" it from your workspace it deletes the original for the other person, which is probably a bad surprise unless it's adorned with red warning text.

I think a "soft" or "scoped" delete is a better default; "delete" should probably be more like "leave". We could provide the hard "delete" as a separate event with its own permissions.

In a soft delete, I think the logic should be to delete the `space` only when the `community_space` count is 0.

ryanatkn:
Fixed tests here: https://github.com/feltcoop/felt-server/pull/183

ryanatkn (changes requested):
I had [one small PR](https://github.com/feltcoop/felt-server/pull/183) to fix tests, and one main change proposal, and the rest is minor and/or doesn't need immediate attention.

The proposed change: I think it'd be better to not return the `space_id` in the response data. The `200` should be enough I think? I also left a related comment about "hard" vs "soft" deletes (aka global or community-scoped) -- the `response` may be different for each of those, but I don't think either needs a `space_id` on it. Specifically I think the `delete_space` response schema should be `type: 'null'`.

src/lib/ui/ui.ts
greatbacon: Okay yeah, I agree with doing more event dispatching. I think opting for event design and response that allows for idempotent or NOP execution of events over adding an additional layer for tracking calling events seems like a more prudent long term solution. Thanks!

src/lib/vocab/space/spaceRepo.ts
greatbacon: This can only delete 1 space at a time though? The input param is a single number, not an array.

src/lib/ui/SpaceDelete.svelte
greatbacon: I think I'll save any UI convos for a future refactor (per above comment).

greatbacon:
> I had [one small PR](https://github.com/feltcoop/felt-server/pull/183) to fix tests, and one main change proposal, and the rest is minor and/or doesn't need immediate attention.
> 
> The proposed change: I think it'd be better to not return the `space_id` in the response data. The `200` should be enough I think? I also left a related comment about "hard" vs "soft" deletes (aka global or community-scoped) -- the `response` may be different for each of those, but I don't think either needs a `space_id` on it. Specifically I think the `delete_space` response schema should be `type: 'null'`. I could see hard deletes possibly returning the count of community ids that were deleted, but returning a list of the community ids might be leaking sensitive info.

Yeah, that makes sense. I started at the DB layer and worked my way up. Figured returning the id of the deleted object would be useful for UI stuff but it's already passed as the param at the top. I'll clear that out.

Re: "hard" & "soft" deletes. I think keeping this as the admin/owner "hard" delete works for the short term, and I'll add an issue to implement a "soft" leave option which can include the cleanup logic to delete the space if no one is actually in it anymore.

ryanatkn (approved):

src/lib/vocab/space/space.events.ts
```diff
@@ -88,4 +88,33 @@ export const read_spaces: ServiceEventInfo = {
...
+	},
+	response: {
+		$id: 'https://felt.social/vocab/delete_space_response.json',
+		type: 'object',
+		properties: {
+			space_id: {type: 'number'},
+		},
+		required: ['space_id'],
+		additionalProperties: false,
+	},
```
ryanatkn, line 112:
```suggestion
	response: {
		$id: 'https://felt.social/vocab/delete_space_response.json',
		type: 'null',
	},
```

I think this is right, and then run `gro gen`.

src/lib/ui/ui.ts
```diff
@@ -400,9 +400,28 @@ export const toUi = (session: Writable<ClientSession>, initialMobile: boolean):
...
+			//update state here
+			const {space_id, community_id} = params;
+			const community = get(communities).find((c) => get(c).community_id === community_id)!;
+			community.update(($community) => ({
+				...$community,
+				// TODO clean this up as part of the data normalization efforts
+				spaces: $community.spaces.filter((space) => space.space_id !== space_id),
+			}));
+			spaces.update(($spaces) => $spaces.filter((space) => get(space).space_id !== space_id));
+			dispatch('select_space', {
```
ryanatkn, line 419: This should be wrapped in an `if` that checks if the deleted `space_id` matches the selected one for each community of `selectedSpaceIdByCommunity`.

ryanatkn:
Sounds like a plan, one change then is to remove `community_id` from the params. I think the only reason to include it would be for authorization purposes but that's a future concern and would only be an optimization, if used at all.

src/lib/ui/ui.ts
```diff
@@ -400,9 +400,28 @@ export const toUi = (session: Writable<ClientSession>, initialMobile: boolean):
...
+		delete_space: async ({params, invoke, dispatch}) => {
+			const result = await invoke();
+			if (!result.ok) return result;
+			//update state here
+			const {space_id, community_id} = params;
+			const community = get(communities).find((c) => get(c).community_id === community_id)!;
+			community.update(($community) => ({
+				...$community,
+				// TODO clean this up as part of the data normalization efforts
+				spaces: $community.spaces.filter((space) => space.space_id !== space_id),
```
ryanatkn, line 416: This logic needs to run for each community, not just the given one.

ryanatkn (changes requested):
Sounds like a plan, one change then to finish the hard delete functionality is to remove community_id from the params. I think the only reason to include it would be for authorization purposes but that's a future concern and would only be an optimization, if used at all.

src/lib/vocab/space/spaceRepo.ts
ryanatkn: It's the `community_spaces` table, so it'll error if more than one community has the space.

ryanatkn (approved):

PR #184, opened 2021-12-06
Add githash to UI

greatbacon:
So this is a utility PR that attempts to add the trimmed (i.e. 7 char long) git hash to a spot in our UI for debugging & deployment confirmations.

It's still kind of a work in progress but I think I have an ok mechanism in place using the .env files & stuff.

However, I've got an issue now where when the deployment task runs to modify the .env.production file, the deployment artififact has already been built with whatever the previous value is.

Is there a way to instead wrap the build task with this git value change or something instead of putting it in the deployment task?


src/infra/serverDeploy.task.ts
```diff
@@ -1,21 +1,44 @@
...
+				return console.log(err);
+			}
+			var result = data.replace(/VITE_GIT_VERSION=.*/g, `VITE_GIT_VERSION=${gitVersion}`);
+			fs.writeFile(ENV_PROD, result, 'utf8', function (err) {
+				if (err) return console.log(err);
+			});
+		});
+
+		//build the actual tar deployment artifact
+		await invokeTask('clean', {_: ['--sveltekit']});
```
ryanatkn, line 39: The `build` task on the line below runs the clean task by default, so this shouldn't be needed. As a side note, the syntax is off here, calling `invokeTask` doesn't parse any args, so instead of this:

```ts
await invokeTask('clean', {_: ['--sveltekit']});
```

should be:

```ts
await invokeTask('clean', {_: [], sveltekit: true});
```

The `_: []` is unfortunate, and I think I'll make it optional in Gro.


or if you want to forward args:


```ts
await invokeTask('clean', {...args, sveltekit: true});
```

src/infra/serverDeploy.task.ts
```diff
@@ -1,21 +1,44 @@
...
+
+		//set git version in the .env.production file
+		const branch = fs.readFileSync('.git/HEAD').toString().trim();
+		const gitVersion = fs
+			.readFileSync('.git/' + branch.substring(5))
+			.toString()
+			.trim()
+			.substring(0, 7);
+
+		fs.readFile(ENV_PROD, 'utf8', function (err, data) {
```
ryanatkn, line 28: This may be causing the issue where your build artifact uses the previous value. The `readFile` and `writeFile` APIs are asynchronous but the control flow in the task isn't `await`ing them.

To fix I'd recommend deleting the `fs` import and using the promise-enabled `fs` on the task context, and using `await` when appropriate. You'll be able to remove this callback nesting.

A slight downside is that the standard Gro `fs` is all asynchronous for portability reasons. So this line:

```ts
run: async ({invokeTask}) => {
		const branch = fs.readFileSync('.git/HEAD').toString().trim();
```

becomes:

```
run: async ({invokeTask, fs}) => {
		const branch = fs.readFileSync('.git/HEAD').toString().trim();
```

The `fs` provided by Gro has both promises and better semantics than the Node `fs`, and it keeps our code more portable. Here's the docs: https://github.com/jprichardson/node-fs-extra

src/infra/serverDeploy.task.ts
```diff
@@ -1,21 +1,44 @@
...
 		const DEPLOY_USER = fromEnv('DEPLOY_USER');
-
 		const deployLogin = `${DEPLOY_USER}@${DEPLOY_IP}`;
-		await invokeTask('clean');
+		const ENV_PROD = '.env.production';
+
+		//set git version in the .env.production file
+		const branch = fs.readFileSync('.git/HEAD').toString().trim();
+		const gitVersion = fs
+			.readFileSync('.git/' + branch.substring(5))
```
ryanatkn, line 23: Does  `substring(5)` belong on the line that defines `branch`?

src/infra/.env.development.default
```diff
@@ -1,2 +1,3 @@
 COOKIE_KEYS=insecureKey1__insecureKey2__insecureKey3
-VITE_DEPLOY_SERVER_HOST='localhost'
\ No newline at end of file
+VITE_DEPLOY_SERVER_HOST=localhost
+VITE_GIT_VERSION=development
```
ryanatkn, line 3: Is there a better name for this?

Instead of "git version"

"build hash"?

"git build hash"?

"git build"?

src/infra/serverDeploy.task.ts
```diff
@@ -50,7 +73,7 @@ export const task: Task = {
 		]);
 		//TEMP: move .env files into root
 		await spawn('scp', [`src/infra/.env.default`, `${deployLogin}:${currentDeploy}/.env`]);
-		await spawn('scp', [`.env.production`, `${deployLogin}:${currentDeploy}/.env.production`]);
+		await spawn('scp', [`${ENV_PROD}`, `${deployLogin}:${currentDeploy}/.env.production`]);
```
ryanatkn, line 76:
```suggestion
		await spawn('scp', [ENV_PROD, `${deployLogin}:${currentDeploy}/${ENV_PROD}`]);
```

src/infra/serverDeploy.task.ts
```diff
@@ -1,21 +1,44 @@
...
 		//TODO gro dev workaround
 		process.env.NODE_ENV = 'production';
 		const {fromEnv} = await import('$lib/server/env');
 
 		const DEPLOY_IP = fromEnv('DEPLOY_IP');
 		const DEPLOY_USER = fromEnv('DEPLOY_USER');
-
 		const deployLogin = `${DEPLOY_USER}@${DEPLOY_IP}`;
-		await invokeTask('clean');
+		const ENV_PROD = '.env.production';
```
ryanatkn, line 18: Might want to move this to `$lib/server/ent.ts` and use it in the 2 spots there.

ryanatkn (changes requested):
Cool idea, had some minor notes and I think I found the bug.

Related -- what about tracking and displaying the build time in the same way?

src/infra/serverDeploy.task.ts
greatbacon: I think creating a dependency on the server code for our deploy task has the possibility to create more headaches than this refactor would solve.

greatbacon:
@ryanatkn So I implemented the changes but the problem still seems to be persisting. I wonder if it has to do with the .env.production values already being loaded into memory through the lifecycle of the task. So the file is changed before the build, but the unchanged value have already been loaded into memory and are carried through?

src/infra/serverDeploy.task.ts
ryanatkn: Didn't see this before, it already has a dependency above on the line `const {fromEnv} = await import('$lib/server/env');`, and I think constants like these are a good example of the benefit of writing tasks in the same language as our apps. If the module in question started any infrastructure like the actual server I agree it shouldn't be imported.

ryanatkn:
That makes sense, two possible fixes come to mind.

The less disruptive one is to add a line that sets the `VITE_GIT_HASH` on the `import.meta.env` (and `process.env`? not sure). That may not work depending on Vite internals.

Another would be to change the line `await invokeTask('build');` to `await spawn('npx', ['gro', 'build']);` (imported as `import {spawn} from '@feltcoop/felt/util/process.js';`)

ryanatkn (approved):
Looks great. Just in case you didn't see this comment, was thinking it's worth reusing the constant in this case: https://github.com/feltcoop/felt-server/pull/184/files#r764291127

src/infra/serverDeploy.task.ts
```diff
@@ -5,17 +5,30 @@ import {DIST_DIRNAME} from '@feltcoop/gro/dist/paths.js';
...
+
+		//set git version in the .env.production file
+		const branch = (await fs.readFile('.git/HEAD', 'utf8')).trim().substring(5);
+		const gitVersion = (await fs.readFile('.git/' + branch, 'utf8')).trim().substring(0, 7);
+
+		const data = await fs.readFile(ENV_PROD, 'utf8');
+		const result = data.replace(/VITE_GIT_HASH=.*/g, `VITE_GIT_HASH=${gitVersion}`);
+		await fs.writeFile(ENV_PROD, result, 'utf8');
+
+		process.env.VITE_GIT_HASH = gitVersion;
```
ryanatkn, line 27: I think there's a cleaner fix -- can you try deleting this line and moving `const {fromEnv} = await import('$lib/server/env');` to its location? The idea is that if you write out to the production env file first, if there are no imports elsewhere, then importing `fromEnv` here should have the fresh values.

src/infra/felt-server.conf
```diff
@@ -1,5 +1,5 @@
 server {
-  server_name localhost;
+  server_name undefined;
```
ryanatkn, line 2: This is a problem right?

greatbacon:
> Looks great. Just in case you didn't see this comment, was thinking it's worth reusing the constant in this case: https://github.com/feltcoop/felt-server/pull/184/files#r764291127

I saw that, but I didn't really like the idea of making our tasks have a direct dependency on server code. It felt like breaking the separation of concerns.

src/infra/serverDeploy.task.ts
greatbacon: Yeah, but that import isn't really a long term solution right? It's just a hack that's in place while some underlying gro infrastructure is fixed isn't it? Or am I misremembering?

PR #186, opened 2021-12-08
disallow creating memberships to personal communities

ryanatkn:
This changes personal communities to disallow invites. On the frontend it removes the invite button. It also adds a test that ensures memberships cannot be duplicates; the table constraint already enforced this.

- [x] write docs
- [x] implement server
- [x] test server impl
- [x] hide the UI button
- [x] add `Community.type` as `'personal' | 'standard'`
- [x] add `Persona.type` as `'account' | 'community'`
- [x] add `persona.community_id`
- [x] finalize API for `lib/db/testMigration` 
- [x] use enum type in the db: https://www.postgresql.org/docs/14/datatype-enum.html
- [x] branching logic in repo for now

When the community isn't found when creating a membership, a 400 is returned, not 404, I was unsure so I looked it up -- https://stackoverflow.com/questions/44915255/is-it-ok-return-http-status-404-in-the-post

Future:

- follow up with #194 to handle the interaction with personal communities


src/lib/vocab/membership/membershipRepo.ts
```diff
@@ -3,12 +3,30 @@ import type {Result} from '@feltcoop/felt';
...
+	}: CreateMembershipParams): Promise<Result<{value: Membership}, ErrorResponse>> => {
+		// TODO use `unwrap` and propagate errors
+
+		// Persona home communities disallow memberships as a hard rule.
+		// They're currently linked by name; this could be changed,
+		// e.g. adding `persona.community_id` and `community.persona_id`
+		const existingCommunityResult = await db.repos.community.findById(community_id);
+		if (!existingCommunityResult.ok) return existingCommunityResult;
+		const existingCommunity = existingCommunityResult.value;
+		if (existingCommunity) {
```
greatbacon, line 21: I feel like this is business logic that should live a level higher in the service and not a repo level.

Unrelated: this also seems like an unwieldy way to check for a "home" community. Since we're starting to formalize it as a special class of community, should we add like an `properties` attribute to Communities and have `home:true` set inside it, or something like that?

src/lib/util/error.ts
```diff
@@ -1,3 +1,3 @@
 export interface ErrorResponse {
-	reason: string;
+	reason: string; // TODO maybe rename this to `message`?
```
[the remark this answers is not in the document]
ryanatkn: Cool same I'll add it to the backlog.

src/lib/vocab/membership/membershipRepo.ts
ryanatkn: Yeah unwieldy is right. I think encoding this more explicitly is a good call for this PR -- I'll take some time to think about it.

src/lib/ui/CommunityInput.svelte
```diff
@@ -25,14 +25,14 @@
 		const result = await dispatch('CreateCommunity', {name, persona_id: $persona.persona_id});
 		pending = false;
 		errorMessage = result.ok ? null : result.message;
+		name = '';
+		done?.();
```
ryanatkn, line 29: bugfix that I found when testing the new code

src/lib/vocab/persona/personaRepo.ts
```diff
@@ -7,39 +7,57 @@ import type {Community} from '$lib/vocab/community/community.js';
...
+					WHERE persona_id = ${persona.persona_id}
+			`;
+			persona.community_id = community.community_id;
+			persona.community_ids = [community.community_id];
+			// TODO this is also a yucky hack
+			community.memberPersonas = [persona];
+			return {ok: true, value: {persona, community}};
+		} else {
+			// TODO this is a hack that can be removed when this code is moved into the service layer
+			return {ok: true, value: {persona, community: null as any}};
```
ryanatkn, line 52: this code is going to be **a lot** better once we move logic into the services

greatbacon (approved):
LGTM! Just had one little wording change in the docs

PR #188, opened 2021-12-08
add contextmenu

ryanatkn:
Implements the first usecase of the context menu -- it removes the existing tab interface at the top of the space nav, replacing it with a tooltip context menu. I currently have it coded so both right and left clicking toggle its visibility; this seems ok for now since we don't have any better primary action. It also adds numerous placeholder entity markers on DOM elements to demonstrate customizing the slot contents of the contextmenu.

Originally used `svelte-popperjs` but I realized positioning the context menu is the least interesting problem to solve here, so I'm skipping that for now and focusing on the right API.

Adds our first subdirectory to `$lib/ui`, `$lib/ui/contextmenu`, to contain all of the new modules. Here's [the MDN entry for `contextmenu`](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) (btw I made our stuff match the casing used by the platform, so it's all one lowercase word). We'll need to implement something else for iOS it seems.

Todo:

- [x] implement with hardcoded content
- [x] add `contextmenu` to `ui`
- [x] fix interaction with dialog

Future:

- implement functionality for the items in the context menu (events)
- change from `personas.name` to `persona.persona_id` in `data-entity` (community too) (or punt?)
- implement entity ids that are unique instead of reusing the serial ids for db collections, along with a global client lookup map
- implement something for `file:`, currently displays a "TODO" because we don't have a simple lookup map
- fix for iOS
- close with the `Escape` key like the modal?
- arbitrary content (blocks with events for buttons?)
- add menu events (maybe `toggle_context_menu`) on arbitrary entities in the DOM tree, and show the whole hierarchy in the menu
- figure out the distinction between context menu and tooltip behavior, and which should be positioned at the mouse, and which should be positioned around the target elements
- fix so it keeps itself onscreen instead of overflowing offscreen (see `svelte-popperjs`)

Sadly it looks like the browsers don't respect a synthetically dispatched `contextmenu` event, and you can't "hold" or pause the real ones, so that means we can't have a contextmenu item like "open browser contextmenu" that does the default browser contextmenu. :(


ryanatkn:
It has some rough edges but it's ready-ish. I think we should walk through some parts together but feel free to do w/e

src/lib/ui/LogoutForm.svelte
```diff
@@ -33,7 +33,9 @@
 	};
 </script>
 
-<div>This account was created {account.created}</div>
+{#if account}
```
ryanatkn, line 36: the context menu being open for a moment while the session turns into a guest causes an error, hence this new check -- there may be a better way to fix this, via more structured event dispatching perhaps

greatbacon (approved):

PR #191, opened 2021-12-10
fix production task env

ryanatkn:
Fixes the env usage (fixing [these hacks](https://github.com/feltcoop/felt-server/pull/177/files/b9eebf69628870a8673209159ce6cc854a1b1015..38808b937001bbc10d5a671f74c71ef6ceff5a6b)) for production tasks by upgrading Gro. (be sure to `npm i`) Removes the hacks we had in place and should hopefully work as expected! I *think* things should work but I could very well be wrong, and I may have forgotten to make some related changes.

Also uses `spawn('npx', ['gro', 'build'])` instead of `invokeTask('build')` to more robustly ensure the environment is correctly set up. The code was somewhat hackily relying on `fromEnv` not being called before build, which is not an assumption we want to make. (this is my mistake, was my suggestion, and a lesson I learned refactoring `gro publish` and `gro deploy` internals)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "ws": "^7.5.4"
       },
       "devDependencies": {
-        "@feltcoop/gro": "^0.45.1",
+        "@feltcoop/gro": "^0.46.0",
         "@sveltejs/adapter-node": "^1.0.0-next.55",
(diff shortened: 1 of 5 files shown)


greatbacon (approved):
LGTM as far as I can tell! Guess we'll just have to test in prod 😜

PR #192, opened 2021-12-10
upgrade deps

ryanatkn:
Upgrades all of our dependencies. The main change is `ws@8`, has some breaking changes. Duplicates some minor things from #191 

At first I saw a weird delay, maybe to do with upgrading `ws`, but I can't reproduce, so I think it's gtg.


greatbacon (approved):
Thanks for keeping our code clean & up to date 🧹

PR #193, opened 2021-12-10
Refactor space to community relation to be many to 1

greatbacon:
This PR includes a DB migration that adds a new `community_id` to the `spaces` table, imports the relevant data from the `community_spaces` table to the new column, and then deletes the relational table. I was originally going to leave the table in place for a little bit while we tested things, but the new delete spaces feature broke while it was still in place and I figured the risk of just outright deleting it was low at this point.


ryanatkn (approved):
Beautiful. There's some `ui.ts` code we can simplify now.

PR #195, opened 2021-12-14
Refactor `SpaceType` into `ViewType`

greatbacon:
Quick PR that refactors the `SpaceType` language in spaces into `ViewType` per a previous conversation

### DIFF
--- a/src/lib/vocab/space/space.ts
+++ b/src/lib/vocab/space/space.ts
@@ -34,7 +34,7 @@ export const SpaceSchema = {
 	additionalProperties: false,
 };
 
-export enum SpaceType {
+export enum ViewType {
 	Home = 'Home',
 	Room = 'Room',
 	Board = 'Board',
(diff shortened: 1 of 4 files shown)


ryanatkn (approved):
:+1:

src/lib/vocab/space/space.ts
```diff
@@ -34,7 +34,7 @@ export const SpaceSchema = {
 	additionalProperties: false,
 };
 
-export enum SpaceType {
+export enum ViewType {
```
ryanatkn, line 37: Wondering if we want to move this stuff to a different module? Is `View` a vocab item now? Maybe just leave a TODO?

src/lib/vocab/space/space.ts
```diff
@@ -43,11 +43,11 @@ export enum SpaceType {
 	Voice = 'Voice',
 	Iframe = 'Iframe',
 }
-export const spaceTypes: SpaceType[] = Object.keys(SpaceType) as SpaceType[];
+export const viewTypes: ViewType[] = Object.keys(ViewType) as ViewType[];
```
ryanatkn, line 46: Wondering if this stuff belongs elsewhere. I'm not sure `View` is a vocab item though. Thinking maybe `$lib/ui/views` could be a new directory, with the data+types from this module and each of the components. Though I don't love groupings by type, maybe it's good in this case? Maybe just leave a TODO?

greatbacon: I think for now keeping it attached to Spaces feels like a good short term path. `view` feels like a sub-type (sub-vocab? attached vocab?) term for utilization by Spaces. I think in the interests of security having it be a concept and list that is owned by server and interpreted by the client makes the most sense.

PR #201, opened 2021-12-17
remove service params usage from repos

ryanatkn:
Changes repo methods to take individual arguments instead of the same params objects as services, making the separation between the two clear. Repos should know nothing about services and events.


greatbacon (approved):
🧹

PR #202, opened 2021-12-17
repo cleanup

ryanatkn:
Cleans up the repos a bit:

- prefer `ALLCAPS` for SQL
- use tabs instead of spaces in SQL template literals (fixes rendering on GitHub and other places)
- format some SQL queries across multiple lines


src/lib/vocab/entity/entityRepo.ts
```diff
@@ -10,26 +10,21 @@ export const entityRepo = (db: Database) => ({
...
 	// TODO maybe `EntityQuery`?
 	filterBySpace: async (space_id: number): Promise<Result<{value: Entity[]}>> => {
 		console.log(`[db] preparing to query for space entities: ${space_id}`);
 		const data = await db.sql<Entity[]>`
-      SELECT f.entity_id, f.content, f.actor_id, f.space_id, f.created, f.updated 
-			FROM entities f WHERE f.space_id= ${space_id}
-    `;
+			SELECT entity_id, content, actor_id, space_id, created, updated 
+			FROM entities WHERE space_id= ${space_id}
+		`;
```
ryanatkn, line 26: I wasn't 100% sure about these changes, if leaving the `f.` might be better for the future. I did this one but skipped a couple others that I saw because of the uncertainty.

greatbacon: I think it makes sense here. You really only want/need that sort of prefacing when doing complex queries anyway.

greatbacon (approved):

PR #204, opened 2021-12-17
fix navigation after deleting selected space

ryanatkn:
Deleting the selected space previously triggered selecting a new space, but that causes the URL to get out of sync. Instead, the current URL is the source of truth for the selected space, so this changes the logic to call SvelteKit's `goto` instead, which then updates the space. (see the main `__layout.svelte` for how)

Was broken because `$app/navigation` can't be imported by our test environment. Fixed it here: https://github.com/feltcoop/gro/pull/294

Upgrades SvelteKit a couple versions to reach parity with Gro, but does not upgrade to the latest, because adapter-node has breaking changes.


src/lib/ui/ui.ts
```diff
@@ -468,10 +469,11 @@ export const toUi = (session: Writable<ClientSession>, initialMobile: boolean):
...
 
-				if (space_id === get(selectedSpaceIdByCommunity)[get(community).community_id])
-					dispatch('select_space', {
-						community_id: get(community).community_id,
-						space_id: get(community).spaces[0].space_id,
+				// TODO maybe make a nav helper or event?
+				const $community = get(community);
+				if (space_id === get(selectedSpaceIdByCommunity)[$community.community_id])
+					goto('/' + $community.name + $community.spaces[0].url + location.search, {
+						replaceState: true,
```
ryanatkn, line 476: Including `replaceState` makes it so the current invalid URL is removed from nav history, so the back/forward buttons skip if like it never existed.

ryanatkn:
Currently breaking because `$app/navigation` can't be imported by our server code. Need to think about the right fix.

greatbacon (approved):

PR #206, opened 2021-12-17
rename selection identifiers

ryanatkn:
Previously we had things like the `selectedPersona` that was a `Readable<Readable<Persona> | null>` -- i.e. it's a nullable reference to a persona store. The problem is that we ambiguously referred to both this object and the inner store as the `selectedPersona`.

This changes the naming convention to make the outer stores suffixed with `Selection`, e.g. `personaSelection`. We continue to call the inner nullable value the `selectedPersona`.

I don't think this is a 100% clear choice -- it's somewhat awkward that it's a suffix in one case and a prefix in the other -- but it's much clearer, and we decided the `Selection` suffix is a decent compromise. Maybe there's a better convention to be found.


greatbacon (approved):
`selection` in the ui store & `selected` for the in component dereferenced form feels good to me

PR #209, opened 2022-01-05
add dialog system

ryanatkn:
Adds the simplest dialog system I could think of. The motivation is to be able to open dialogs without requiring the triggering component to remain mounted. Currently we have a contextmenu button that tries to open a dialog, but it fails because when the contextmenu closes, so does its child dialog component. 

As a side effect, we'll now be able to drive dialogs throughout the whole system programmatically like any other event. (e.g. so we could use it to build a tutorial system and user macros/automations)

Uses two new events, `'OpenDialog'` and `'CloseDialog'`. I implemented it for only one existing dialog, the `ManageMembershipsForm`, because it's currently broken. 

Might want to rename the `DialogState` interface. I kept it intentionally simple so there's no wrapper store or class, but I think we'll need features that increase complexity.

Next steps:

- change the other dialogs to use this system (will inform more of the following questions)
- API for `'CloseDialog'` -- should it send the entire object? An ID?
- types for component names
- how to handle return values (Picker usecase)
- what JSON data structure should we be using? Svelte AST or JSON-LD? dynamically load components? move from an import to context?
- should it interact with the URL nav at all? currently it is completely separate

### DIFF
--- a/src/lib/app/components.ts
+++ b/src/lib/app/components.ts
@@ -1,0 +1,11 @@
+import ManageMembershipForm from '$lib/ui/ManageMembershipForm.svelte';
+import type {SvelteComponent} from 'svelte';
+
+// The collection of components that can be dynamically mounted by the app.
+
+// TODO refactor this to load these components on demand,
+// instead of preloading the entire component library
+
+export const components: {[key: string]: typeof SvelteComponent} = {
+	ManageMembershipForm,
+};
--- a/src/lib/app/ContextmenuSlot.svelte
+++ b/src/lib/app/ContextmenuSlot.svelte
@@ -108,12 +108,6 @@
 		width: 100%;
 	}
 
-	.contextmenu-slot .markup {
-		display: flex;
-		flex-direction: column;
-		align-items: center;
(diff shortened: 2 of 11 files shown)


src/lib/app/components.ts
```diff
@@ -0,0 +1,11 @@
...
+import type {SvelteComponent} from 'svelte';
+
+// The collection of components that can be dynamically mounted by the app.
+
+// TODO refactor this to load these components on demand,
+// instead of preloading the entire component library
+
+export const components: {[key: string]: typeof SvelteComponent} = {
+	ManageMembershipForm,
+};
```
ryanatkn, line 11: This is the biggest negative I see right now in the PR. Any components that we want to mount programmatically (e.g. through the new `'OpenDialog'` event) must be loaded app-wide, so we don't get code splitting, bloating the initial bundle.

The fix is probably to load components dynamically at runtime. This is related to our other conversation about having content defined as static ASTs versus dynamically loadable JSON-LD.

greatbacon (approved):

PR #215, opened 2022-01-11
add view vocab term

ryanatkn:
This is a refactor that creates `$lib/vocab/view` and moves the view definitions that were previously in `$lib/vocab/space/space.ts`. It also moves the `viewComponents` to `$lib/ui/view/components.ts`, which mirror the pattern we have with dialogs and the contextmenu, and then includes them in the main `$lib/app/components` lookup map.

It changes the `components` usage to use Svelte context instead of props. In the future, we'll use this to lazy-load components. Without this change, we would have had to thread `components` down through every route, which we know isn't a long-term solution; this allows us to localize the "hacks" with an interface that can evolve to the correct design.

Also moves all of the view components to `$lib/ui/view/`. The related components like `ForumItem.svelte` remain in `$lib/ui/`. The idea here is that views conform to a specific interface, and everything else is technically decoupled, and can be broadly reused, even if the name suggests otherwise. So any new views or components may use `ForumItem.svelte`; I think it makes **slightly more** sense to keep them separate, not co-located.


greatbacon:
In my mental model, `vocab` was a directory reserved primarily for server side components. It feels off to me to move svelte components into this directory. Would a `src` level directory or a top level directory in `ui` maybe make more sense?

ryanatkn:
That's a really good point, I'm of two minds on this. Is a vocabulary term:

1. a word for something that has a consistent meaning throughout the fullstack app?

or 

2. a word for something that has a database representation, repo, and services?

In the first case, the lowest common denominator might be anything with a type/schema and consistent fullstack meaning. I think it's clear "contextmenu" is a client-only UI concern and not vocabulary, but "view" is blurry because our server stores them, services will take them as inputs, and the server needs to validate them with schemas.

In the second case, the lines blur differently: Isn't tying "the vocabulary terms Felt defines" with database representation a conflation of implementation details with meaning? Tables and repos could be fully swapped out for a different backend, so maybe the important piece is the services? But then what if we have an object that is used as inputs to many services, but doesn't have its own services? This may be a bad example: a `Permission` object that's always related to  `Role` service events, with no standalone permission events.

If this doesn't resonate I'd support moving it to `$lib/ui/view` and we can always revisit if/when contradictions or frictions appear.

Currently `$lib/vocab/event` is the other oddball, it could go into `$lib/ui` too.

greatbacon:
Yeah, I think putting the types/schemas in `vocab` still makes sense It was more the strictly frontend components landing in there that gave me pause. If only because I think that would make it the first set of `.svelte` files to be outside of the `ui` directory & that seems like a confusing organizational choice.

So I think the third option you outlined feels the most right to me!

ryanatkn:
Ok cool yeah that makes a lot of sense to me. Moved the views and the `components.ts` to the new dir `$lib/ui/view/`, and kept `$lib/vocab/view/view.ts` in place.

greatbacon (approved):

PR #216, opened 2022-01-12
add entity explorer view

ryanatkn:
Adds an `EntityExplorer.svelte` view component for generically viewing a `Space`, and adds the `'ViewSpace'` event to control it. Dispatching the event with a null `view` value reverts to the community default view. (this is how the `EntityExplorer` closes itself) Adds the new `SpaceContextmenu` entry `"View with EntityExplorer"`.

The `EntityExplorer` is just a stub that puts JSON onscreen. It uses `EntityItems` and `EntityItem`, copying the pattern we use for other view types like forum.

Also removes the `ViewType` in favor of a plain string. Didn't buy us much, mainly caused type annoyances, and won't be useful in a world where views are dynamically async created/loaded. (b/c people will be able to create their own ad-hoc views, there's no enumeration of them available)

Adds a new `mutable` store helper to efficiently handle mutable values like the `WeakMap` used to store the client view overrides. We need it because we use the Svelte `immutable` compiler flag everywhere, and we don't want to get in the habit of disabling it. Might make sense to upstream to Felt before merging, since I'm pretty confident in the API. This is a useful building block that we can use to improve some of our other client-side data modeling.

Future:

- contextmenu entry will be a submenu with `"View space with...` or something, not hardcoded to just the `EntityExplorer`


src/lib/ui/ui.events.ts
```diff
@@ -85,7 +85,7 @@ export const events: EventInfo[] = [
 			$id: 'https://felt.social/vocab/SelectCommunityParams.json',
 			type: 'object',
 			properties: {
-				community_id: {anyOf: [{type: 'number'}, {type: 'null'}]},
+				community_id: {type: ['number', 'null']},
```
ryanatkn, line 88: This is a no-op, using the shorthand that I saw you use elsewhere.

src/lib/vocab/view/components.ts
```diff
@@ -0,0 +1,21 @@
...
+import Room from '$lib/ui/Room.svelte';
+import Board from '$lib/ui/Board.svelte';
+import Forum from '$lib/ui/Forum.svelte';
+import Notes from '$lib/ui/Notes.svelte';
+import Iframe from '$lib/ui/Iframe.svelte';
+import Voice from '$lib/ui/Voice.svelte';
+import EntityExplorer from '$lib/ui/EntityExplorer.svelte';
+
+export const viewComponents: Record<string, typeof SvelteComponent> = {
+	Home,
```
greatbacon, line 13: Is this file a dup from the other PR?

greatbacon (approved):
Looks good to me! Just had one question about a possible dupe file that I thought was in `ui`, could be misrembering lol

src/lib/vocab/view/components.ts
ryanatkn: Whoops yeah good catch, merge error.

ryanatkn:
Ya my mistake, I messed up the merge.

PR #217, opened 2022-01-12
add submenus to the contextmenu

ryanatkn:
This adds "submenu" support to the contextmenu, as [Apple's Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/controls/context-menus/) refers to them.

- [x] basic behavior
- [x] keyboard nav: I'm trying to follow these guidelines: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html
- [x] programmatic activation
- [x] remove test menu code

questions

- should `Tab` close the contextmenu like Firefox, or be a no-op like Chrome?
- should the items be implemented as `a` elements like [the menu button guidelines](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html)? Or buttons? I didn't go with buttons because the behavior of the contextmenus in Firefox/Chrome do not use tab `Tab` key. I'm currently thinking in terms of programmatic actions, not navigable links, but maybe we can use the hash for this? gets complicated -- an `a` without `href` is an a11y warning in Svelte, and we're also not using `focus` currently, but we could with `a`

future:

- positioning onscreen without overflowing the edges
- fix the precision problem where the menu closes the instant your mouse leaves the element by adding a delay to hiding the menus and/or a larger transparent hitbox
- improve a11y (including "type characters to select items" as described here: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions-active-descendant.html)
- keyboard command to open contextmenu on focused element (and mouse position?)
- better styling
  - smaller icons for users/communities
  - differentiate active selection


greatbacon (approved):
LGTM

PR #218, opened 2022-01-18
fix text overflow in messages

ryanatkn:
Fixes text overflowing post content areas like with long links and other long unbroken strings of text.

Adds one extra property because of Firefox Android support for `white-space: break-spaces` -- https://caniuse.com/mdn-css_properties_white-space_break-spaces

Also changes the layout of the chat room messages to put the user icon to the left of the user name and text, a better use of vertical space and aligns things more nicely.

Also fixes the contextmenu shift key as we discussed.

The `Avatar` component is starting to look a bit stretched and awkward. We may want to split out `AvatarName` and `AvatarIcon` or something.

### DIFF
--- a/src/lib/app/contextmenu/LinkContextmenu.svelte
+++ b/src/lib/app/contextmenu/LinkContextmenu.svelte
@@ -24,7 +24,7 @@
 <!-- TODO if it's an external link, add target="_blank" -->
 <a href={value}>
 	<span class="icon">🔗</span>
-	{formatUrl(value)}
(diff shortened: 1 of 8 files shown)


greatbacon (approved):
Nice! Tested the mac fix & the name shifted avatars looks a lot smoother to me too.

ryanatkn:
Awesome, I made one small adjustment, extracting a `.formatted` class into our main styles to avoid the copypasta.

PR #219, opened 2022-01-19
Refactor entity content to data

greatbacon:
Okay, so this is step 1 from the conclusion of our threads & forums discussion from the last week.

This converts the `content` column inside our Entities into the `data` column, also changing it's type from String to JSONB.

It also has some loose framework around typing the data at the view layer.

### DIFF
--- a/src/lib/app/eventTypes.ts
+++ b/src/lib/app/eventTypes.ts
@@ -169,9 +169,7 @@ export type DeleteSpaceResponseResult = ApiResult<DeleteSpaceResponse>;
 export interface CreateEntityParams {
 	actor_id: number;
 	space_id: number;
-	data: {
-		[k: string]: unknown;
-	};
+	data: EntityData;
 }
 export interface CreateEntityResponse {
 	entity: Entity;
--- a/src/lib/app/eventTypes.ts
+++ b/src/lib/app/eventTypes.ts
@@ -169,7 +169,9 @@ export type DeleteSpaceResponseResult = ApiResult<DeleteSpaceResponse>;
 export interface CreateEntityParams {
 	actor_id: number;
 	space_id: number;
-	content: string;
+	data: {
+		[k: string]: unknown;
+	};
 }
 export interface CreateEntityResponse {
 	entity: Entity;
--- a/src/lib/app/eventTypes.ts
+++ b/src/lib/app/eventTypes.ts
@@ -9,6 +9,7 @@ import type {Persona} from '$lib/vocab/persona/persona';
 import type {Membership} from '$lib/vocab/membership/membership';
 import type {Space} from '$lib/vocab/space/space';
 import type {Entity} from '$lib/vocab/entity/entity';
+import type {EntityData} from '$lib/vocab/entity/entityData';
 import type {ViewData} from '$lib/vocab/view/view';
 import type {DispatchContext} from '$lib/app/dispatch';
 
--- a/src/lib/db/migrations/00009-refactor-entity-content-to-data.js
+++ b/src/lib/db/migrations/00009-refactor-entity-content-to-data.js
@@ -2,7 +2,7 @@
 export const up = async (sql) => {
 	await sql`
 	 UPDATE entities
-	 SET content = concat('{"type":"Note","partOf":null,"title","","content":"',content,'"}');
+	 SET content = concat('{"type":"Note","content":"',content,'"}');
 	 `;
 
 	await sql`
--- a/src/lib/db/migrations/00009-refactor-entity-content-to-data.js
+++ b/src/lib/db/migrations/00009-refactor-entity-content-to-data.js
@@ -2,7 +2,7 @@
 export const up = async (sql) => {
 	await sql`
 	 UPDATE entities
-	 SET content = concat('{"type":"Post","partOf":null,"content":"',content,'"}');
+	 SET content = concat('{"type":"Note","partOf":null,"title","","content":"',content,'"}');
 	 `;
 
 	await sql`
--- a/src/lib/db/migrations/00009-refactor-entity-content-to-data.js
+++ b/src/lib/db/migrations/00009-refactor-entity-content-to-data.js
@@ -1,0 +1,14 @@
+/** @param {import('postgres').Sql<any>} sql */
+export const up = async (sql) => {
+	await sql`
+	 UPDATE entities
+	 SET content = concat('{"type":"Post","partOf":null,"content":"',content,'"}');
+	 `;
+
+	await sql`
+	ALTER TABLE entities
+	ALTER COLUMN content TYPE jsonb USING content::jsonb;
+	`;
+
+	await sql`ALTER TABLE entities RENAME COLUMN content TO data;`;
+};
--- a/src/lib/db/repos.test.ts
+++ b/src/lib/db/repos.test.ts
@@ -99,11 +99,13 @@ test__repos('create, change, and delete some data from repos', async ({server}) 
 
 	const entityContent1 = 'this is entity 1';
 	const entityContent2 = 'entity: 2';
+	const data1 = {type: 'Post', partOf: null, content: entityContent1};
+	const data2 = {type: 'Post', partOf: null, content: entityContent2};
 	const entity1 = await unwrapEntity(
-		server.db.repos.entity.create(persona.persona_id, space.space_id, entityContent1),
+		server.db.repos.entity.create(persona.persona_id, space.space_id, data1),
 	);
 	const entity2 = await unwrapEntity(
-		server.db.repos.entity.create(persona.persona_id, space.space_id, entityContent2),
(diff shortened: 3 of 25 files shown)


src/lib/vocab/entity/entity.events.ts
```diff
@@ -9,9 +9,9 @@ export const CreateEntity: ServiceEventInfo = {
 		properties: {
 			actor_id: {type: 'number'},
 			space_id: {type: 'number'},
-			content: {type: 'string'},
+			data: {type: 'object'},
```
greatbacon, line 12: I'm not 100% on if my choice to type data as `Object` in the model & schema was the correct one. Any thoughts?

ryanatkn: Object is right, and eventually it'll be the schema for the data type. I'd suggest changing it to this:

```ts
data: {type: 'object', tsType: 'EntityData'},
```

and then renaming the new `Post` interface from `./entityDataType.ts` to `EntityData` and relocating it to the bottom of this file, and then changing `data: Object;` in the `Entity` interface at the top of the file to `data: EntityData;`.

That way when we generate vocab types from schemas (after finishing the flattening normalization work), it'll be properly synced up.

src/lib/vocab/entity/entityRepo.ts
```diff
@@ -7,21 +7,21 @@ export const entityRepo = (db: Database) => ({
 	create: async (
 		actor_id: number,
 		space_id: number,
-		content: string,
+		data: Object,
```
ryanatkn, line 10: Assuming the first comment I made goes through, this type should be `EntityData`.

src/lib/vocab/entity/entityRepo.ts
```diff
@@ -7,21 +7,21 @@ export const entityRepo = (db: Database) => ({
...
 	): Promise<Result<{value: Entity}>> => {
-		const data = await db.sql<Entity[]>`
-			INSERT INTO entities (actor_id, space_id, content) VALUES (
-				${actor_id},${space_id},${content}
+		const response = await db.sql<Entity[]>`
```
ryanatkn, line 12: I kept looking at these a bit ago, wondering what the best name for the query return value is, so we can be consistent across repos.

Response is ok, I just don't love how it overlaps with request/response in services and server/client modules that deal with network stuff.

Result is no good, because that's what we call the success/failure value/error wrapper returned everywhere when things can fail.

Data is no good because it's too ambiguous and now we have actual properties/columns named `data`.

Looking at [the driver docs](https://github.com/porsager/postgres), it names things according to what's returned from the database, so in this case it'd be:

```ts
const [entities] = ...
```

idk

src/lib/db/migrations/00009-refactor-entity-content-to-data.js
```diff
@@ -0,0 +1,14 @@
+/** @param {import('postgres').Sql<any>} sql */
+export const up = async (sql) => {
+	await sql`
+	 UPDATE entities
+	 SET content = concat('{"type":"Post","partOf":null,"content":"',content,'"}');
```
ryanatkn, line 5: I think we should default to type `Object` for now or even just omit it. Null values like `partOf` can be omitted too.

src/lib/db/repos.test.ts
```diff
@@ -99,11 +99,13 @@ test__repos('create, change, and delete some data from repos', async ({server})
 
 	const entityContent1 = 'this is entity 1';
 	const entityContent2 = 'entity: 2';
+	const data1 = {type: 'Post', partOf: null, content: entityContent1};
+	const data2 = {type: 'Post', partOf: null, content: entityContent2};
```
ryanatkn, line 103: Should match whatever is chosen in the migration, so either:

```ts
const data1 = {type: 'Object', content: entityContent1};
```

or
 
```ts
const data1 = {content: entityContent1};
```

src/lib/ui/NotesItem.svelte
```diff
@@ -2,13 +2,15 @@
...
 	import type {Entity} from '$lib/vocab/entity/entity';
+	import type {Post} from '$lib/vocab/entity/entityDataType';
 	import {getApp} from '$lib/ui/app';
 
 	const {
 		ui: {contextmenu, findPersonaById},
 	} = getApp();
 
 	export let entity: Readable<Entity>;
+	$: data = $entity.data as Post;
```
ryanatkn, line 13: Just musing about what the types of this could ultimately be, not to change now; one option is something like `EntityWith<'content' | 'someOtherProperty'>` that is a generic that says, "I want an `Entity` that must have these properties". Then parent components would have to handle that case to typecheck, like:

```svelte
{#if 'content' in entity'}
  <NotesItem {entity} />
{:else}
  some fallback to display a generic entity, or just ignore
{/if}

```

src/lib/vocab/entity/entityDataType.ts
```diff
@@ -0,0 +1,5 @@
+export interface Post {
+	type: String;
```
ryanatkn, line 2:
```suggestion
	type: string;
```

src/lib/vocab/entity/entityDataType.ts
```diff
@@ -0,0 +1,5 @@
...
+	partOf?: number;
+	content: String;
+}
```
ryanatkn, line 5: these need to be `string` lowercase

src/lib/vocab/random.ts
```diff
@@ -22,7 +22,7 @@ export const randomPersonaName = randomString;
 export const randomCommunnityName = randomString;
 export const randomSpaceUrl = randomString;
 export const randomSpaceName = randomString;
-export const randomContent = randomString;
+export const randomData = (): any => ({type: 'Post', partOf: null, content: randomString()});
```
ryanatkn, line 25: type should be `EntityData`

src/lib/vocab/random.ts
```diff
@@ -47,15 +47,15 @@ export const randomCommunityParams = (persona_id: number): CreateCommunityParams
 };
 export const randomSpaceParams = (community_id: number): CreateSpaceParams => ({
 	community_id,
-	content: randomContent(),
+	content: JSON.stringify(randomData()),
```
ryanatkn, line 50:
```suggestion
	content: randomString(),
```

This is clearer, since `randomData` is different than the JSON content here, which will be migrated to `view` soon.

src/lib/ui/BoardItem.svelte
```diff
@@ -12,6 +13,7 @@
 	} = getApp();
 
 	export let entity: Readable<Entity>;
+	$: data = $entity.data as Post;
```
ryanatkn, line 16: After changing the types like mentioned elsewhere, you can remove this line in the 4 files and use `$entity.data.content` directly.

ryanatkn:
Looks great, most of my comments were about the types.

src/lib/db/migrations/00009-refactor-entity-content-to-data.js
greatbacon: Fine, I'll call it `Note`, but we just spent a week and a half going through defining our entity data models, and I wan't to make sure we're leaving breadcrumbs for ourselves in the future so that we don't forget and go through this all over again.

src/lib/vocab/entity/entity.events.ts
greatbacon: So as we continue to add more Data types, are we just going to expand this EntityData with more attributes to match all of them? Do the new types just extend this generic EntityData? Where do we go from here when we want actual concrete types?

src/lib/db/migrations/00009-refactor-entity-content-to-data.js
ryanatkn: We talked about this -- `Note` makes the most sense.

src/lib/vocab/entity/entity.events.ts
ryanatkn: We talked through this and got the entity type pattern set.

ryanatkn:
The type error can be fixed by adding this line to the `eventTypes.gen.ts` file:

```
```

src/lib/app/eventTypes.ts
```diff
@@ -169,7 +169,7 @@ export type DeleteSpaceResponseResult = ApiResult<DeleteSpaceResponse>;
 export interface CreateEntityParams {
 	actor_id: number;
 	space_id: number;
-	content: string;
+	data: EntityData;
```
ryanatkn, line 172: The type error can be fixed by adding this line to the `eventTypes.gen.ts` file:

```ts
```

Although since you added schemas the type could be generated and embedded in this file without an import. We don't have a pattern for that yet though, it's not that hard, we could figure something out in this PR if you want.

ryanatkn (approved):
wooo! :clap:  I've got some minor comments but I'll do them as a followup PR, since this one is so big.

PR #226, opened 2022-01-22
change login and logout endpoints to account services

ryanatkn:
This is a long-punted refactor that turns login/logout into services. This is the last of the endpoints that isn't a service!

- adds the `LoginAccount` and `LogoutAccount` services+events and removes `LogIn` and `LogOut`, which used the now-removed workaround `"RemoteEvent"` type that made API calls manually as an escape hatch, so the `ui` handler code is much cleaner now
- adds a `session` service param to deal with the fact that login/logout need to use the `req` to modify the session
- adds the `authorize` function, removing `authorizationMiddleware`, and calls it from both the http and websocket service middleware; it uses a new optional property on events named `authorize` -- if this property is `true` or omitted, the event's service requires an authenticated `account_id`, and a `false` value allows the service to be called without logging in (currently, just the login service itself!); in the future we'll add support for additional values that further customize the authorization behavior per event
- adds the `websockets` property to events to handle the `logout` case; it can't modify the session so it's disallowed from being called via websockets, and the websocket middleware will return an appropriate error if the client mistakenly uses the websocket client instead of the http one
- the `dispatch` function now calls a function to determine which client it should use to send each event
- changes the `accountName` login param to a more generic `username`; logins often take multiple kinds of inputs, not just the account name (e.g. if `account.name` and `account.email` are distinct in the future, so you could sign up with just a name and later add an email for instances that are email-optional), so I think this is a better future-proof name, so it's not tightly coupled to `account.name`


src/lib/server/ApiServer.ts
```diff
@@ -81,21 +77,15 @@ export class ApiServer {
 				});
 				next();
 			})
-			.use(toCookieSessionMiddleware())
-			.use(toAuthenticationMiddleware(this))
-			// API
-			.post('/api/v1/login', toLoginMiddleware(this)) // TODO wait shouldn't this fail in Polka's system??
```
greatbacon, line 87: 🧹

src/lib/server/ServiceEffects.ts
```diff
@@ -0,0 +1,33 @@
...
+ * while the service processes a request.
+ */
+export class ServiceEffects implements IServiceEffects {
+	constructor(private readonly req: ApiServerRequest | WebsocketServerRequest) {}
+
+	login(account_id: number) {
+		if (this.req.session) {
+			this.req.session.account_id = account_id;
+		} else {
+			// TODO how to handle this? in what cases can this happen? can it happen if our auth middleware is configured correctly?
```
greatbacon, line 25: Probably want to at least leave a warning log statement here saying what happened.

src/lib/server/service.ts
```diff
@@ -1,5 +1,8 @@
-import type {Database} from '$lib/db/Database';
-import type {ServiceEventInfo} from '$lib/vocab/event/event';
+import {type Result} from '@feltcoop/felt';
+
+import {type Database} from '$lib/db/Database';
+import {type ServiceEventInfo} from '$lib/vocab/event/event';
+import {type IServiceEffects} from './ServiceEffects';
```
greatbacon, line 5: Don't we usually use the explicit `$lib..` syntax?

src/lib/server/WebsocketServer.ts
```diff
@@ -61,10 +72,9 @@ export class WebsocketServer extends (EventEmitter as {new (): WebsocketServerEm
 		wss.on('error', (error) => {
 			console.log('[wss] error', error);
 		});
-		wss.on('headers', (headers, req) => {
-			// TODO could parse cookies from these headers if we don't connect the `WebSocketServer` to the `server` above
-			console.log('[wss] req.url headers', req.url, headers);
-		});
+		// wss.on('headers', (headers, req) => {
```
greatbacon, line 75: Any reason to not just clean out these comments?

src/lib/app/eventTypes.ts
```diff
@@ -60,18 +60,18 @@ export interface EventResponseByName {
 	Ping: PingResponse;
 }
 
-export interface LogInParams {
-	accountName: string;
+export interface LoginAccountParams {
+	username: string;
```
ryanatkn, line 64: Per our conversation this morning, this will likely be "email" in a future PR. Should I revert it to `accountName`, leave as `username`, or change to `email` now?

src/lib/ui/ui.ts
```diff
@@ -265,59 +265,18 @@ export const toUi = (
...
-				}
-			} catch (err) {
-				console.error('[LogOut] err', err); // TODO logging
-				return {
-					ok: false,
-					status: 500,
-					message: 'unknown error',
-				};
-			}
+		LogoutAccount: async ({invoke}) => {
```
greatbacon, line 275: 🤌

src/lib/app/eventTypes.ts
greatbacon: I'd just leave it as username for now

src/lib/server/WebsocketServer.ts
ryanatkn: Mainly because I don't know if we'll need to use this, and leaving it here reminds us that it exists without us looking through the docs. But now that we're talking about it, we can probably commit to memory so I'll delete it -- if we need to inspect of modify the headers on the initial websocket handshake, this is what we'd need to use.

src/lib/server/service.ts
ryanatkn: Good catch thank you, there's a handful of other places I'll update as well (searching `from './`)

src/lib/ui/LoginForm.svelte
```diff
@@ -55,12 +55,13 @@
...
 	<input
-		bind:this={accountNameEl}
-		bind:value={accountName}
+		bind:this={usernameEl}
+		bind:value={username}
 		on:keypress={onKeypress}
 		{disabled}
 		placeholder="account name"
 		use:autofocus
+		autocomplete="username"
```
ryanatkn, line 64: Added this because of a Chrome warning, "username" is a standard: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls%3A-the-autocomplete-attribute

src/lib/vocab/account/accountServices.ts
```diff
@@ -0,0 +1,83 @@
...
+import type {
+	LoginAccountParams,
+	LoginAccountResponseResult,
+	LogoutAccountParams,
+	LogoutAccountResponseResult,
+} from '$lib/app/eventTypes';
+
+export const loginAccountService: Service<LoginAccountParams, LoginAccountResponseResult> = {
+	event: LoginAccount,
+	perform: async ({repos, params, account_id, effects}) => {
```
ryanatkn, line 14: Most of the code here is copied modified from the old `loginMiddleware`

src/lib/vocab/event/event.ts
```diff
@@ -17,6 +17,7 @@ export interface ClientEventInfo {
 export interface ServiceEventInfo {
 	type: 'ServiceEvent';
 	name: string; // `snake_cased`
+	authorize?: boolean; // `true` by default
```
ryanatkn, line 20: This is overloaded to mean two things right now:

- "for endpoints like login, don't try to use the websocket client because it doesn't work for services"
- "for endpoints like logout, the websocket client can't affect the session, so don't use the websocket client for it"

I think we're going to put authorization configuration in the `authorize` property in the future, so I think it's the second case that makes this messy. We could add a second property like `websockets: false` or `httpOnly: true` to fix that one case to make things more explicit.

ryanatkn: Ok yeah I'm fairly confident in the change described in the previous comment, made it here: https://github.com/feltcoop/felt-server/pull/226/commits/3b8b46603df880c9f30ad3d0241d3602306e910f

src/lib/vocab/event/event.ts
```diff
@@ -17,7 +17,8 @@ export interface ClientEventInfo {
 export interface ServiceEventInfo {
 	type: 'ServiceEvent';
 	name: string; // `snake_cased`
-	authorize?: boolean; // `true` by default
+	authenticate?: boolean; // `true` by default -- does this service require login?
+	websockets?: boolean; // `true` by default -- can this service be called via websockets?
```
greatbacon, line 21: 👍 I like this!

src/lib/server/ServiceEffects.ts
```diff
@@ -0,0 +1,33 @@
...
+	logout(): void;
+}
+
+/**
+ * `ServiceEffects` is the API available to services beyond the database repos.
+ * Unlike repo methods, they do not compose with database transactions,
+ * so they cannot be undone automatically if something else goes wrong
+ * while the service processes a request.
+ */
+export class ServiceEffects implements IServiceEffects {
```
greatbacon, line 18: So overall this looks like a good PR. I think this naming/module/thingy is the only piece that really stands out. To me the question to ask ourselves here is really, "What is the purpose & long term vision on this component?"

I know you had mentioned this possibly owning some email logic or the like down the road. Right now though it seems like this logic (how & when it's referenced & initialized, what it does, etc.) is primarily around wrapping & manipulating the session object.

I'd maybe float a name like `SessionService` or `SystemService`? Since this really does just seem to be a service that does work, even if that work doesn't necessarily involve DB transactions.

This seems like a big convo though, so maybe it's best to do all this in person on Monday?

ryanatkn: Yeah I agree and I'm not sure what to say here, so I'll outline the problem so we can talk more about it.

The motivation was to constrain the `Service` API so they're less "do anything" and more "do stuff with the repos in transactions and a minimal non-transaction-safe API, and return a value". The benefit of defining `Services` like this is we get a consistent input/output interface with the type-and-schema-checked params/response. I named it `ServiceEffects` because it's outside of the transaction rollback-safety that services normally have. It's similar to "side effects" described by functional programming. I don't think it's a good name though, and I'm unclear whether the whole design is a good idea.

Given that services have this very tight definition, I think it's best to use a word other than `Service` to describe this session action, but I'm not as confident in constraining the interface. I'm going to use the placeholder `Api` below in these examples instead of `Service` for this reason.

One option is we pass "the whole world" to services, so they'd get the `server` and `req` object (and therefore whole `db`, not just the `repos`) and can do whatever they need. Then we could delete the whole `ServiceEffects` system. This could be paired with expecting services to import any module dependencies they may have, and rely on inputs only for stateful/per-request things like the `db` and `req`. In this case `logout` could be a function that any module can import, and it has a dependency on the `req` object that each service would receive. Emailing could work similarly, either as a module and/or using dependencies on the `server`.

A second option is we fix the name of `ServiceEffects` and constrain its interface to be scoped to just the session, something like `SessionApi`. This way we would pass `session` as a param to the services, and they could call `session.logout()`.

Another option is we fix the name and keep it a wide interface to the session and future things like email, like `SystemApi` or `ServiceApi`. So it instead of the `effects` param it would be `api.logout()` and such. In this case, emailing would be `api.email('somebody@email.net', {body: 'etc'})`. 

Another option is we emit events, but I decided against that because I think services will need explicit control flow over these things. Emailing is the first example I can think of here: it'll be an async call.

The constrained interfaces seem slightly at odds with the idea that services can import any modules they need to get work done. In the future, I think we'll want services to be modular, so custom deployments can add any services they need via simple JS modules, and they might want "the whole world" of our internal server, along with any external modules.

ryanatkn:
Ok cool I got all the changes in that we talked about -- there's just one commit there at the end with the diff: https://github.com/feltcoop/felt-server/pull/226/commits/bbc414467c1dea1d362833dae88baee8c62ba33e

greatbacon (approved):
🎉

PR #230, opened 2022-01-25
Add timestamp formatter

greatbacon:
So like most things this took a lot less time than I was expecting lol

Timestamps were only in a couple of places on the client. I removed a couple until we have better spots for managing personas & the account. 

As for the space header & RoomItem, my CSS-fu failed me and I wasn't able to get the dates to the right without padding, which I don't think is right, since we've got flex-boxes to work with instead. But I couldn't quite figure it out. 😞

### DIFF
--- a/src/lib/ui/RoomItem.svelte
+++ b/src/lib/ui/RoomItem.svelte
@@ -35,9 +35,7 @@ And then PersonaContextmenu would be only for *session* personas? `SessionPerson
 	<div class="markup formatted">
 		<div class="signature">
 			<Avatar name={toName($persona)} icon={toIcon($persona)} showIcon={false} />
-			<div class="timestamp">
-				{format(new Date($entity.created), 'Pp')}
-			</div>
+			{format(new Date($entity.created), 'Pp')}
 		</div>
 		<div>
 			{$entity.data.content}
@@ -55,11 +53,7 @@ And then PersonaContextmenu would be only for *session* personas? `SessionPerson
 	.signature {
 		display: flex;
 		align-items: center;
-	}
-	.timestamp {
-		flex: 1;
-		display: flex;
-		justify-content: flex-end;
+		justify-content: space-between;
 	}
 	.markup {
 		padding: 0 0 0 var(--spacing_md);
--- a/src/lib/ui/RoomItem.svelte
+++ b/src/lib/ui/RoomItem.svelte
@@ -57,7 +57,9 @@ And then PersonaContextmenu would be only for *session* personas? `SessionPerson
 		align-items: center;
 	}
 	.timestamp {
-		padding-left: 60%;
+		flex: 1;
+		display: flex;
+		justify-content: flex-end;
 	}
 	.markup {
 		padding: 0 0 0 var(--spacing_md);
--- a/src/lib/ui/RoomItem.svelte
+++ b/src/lib/ui/RoomItem.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import type {Readable} from 'svelte/store';
+	import {format} from 'date-fns';
 
 	import type {Entity} from '$lib/vocab/entity/entity';
 	import Avatar from '$lib/ui/Avatar.svelte';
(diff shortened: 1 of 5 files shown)


src/lib/ui/RoomItem.svelte
```diff
@@ -28,13 +29,15 @@ And then PersonaContextmenu would be only for *session* personas? `SessionPerson
...
+	<div class="signature">
 		<Avatar name={toName($persona)} icon={toIcon($persona)} showName={false} />
 	</div>
 	<div class="markup formatted">
-		<div class="timestamp">
+		<div class="signature">
 			<Avatar name={toName($persona)} icon={toIcon($persona)} showIcon={false} />
-			{$entity.created}
+			<div class="timestamp">
+				{format(new Date($entity.created), 'Pp')}
```
ryanatkn, line 39: the `created` properties are already `Date` objects, so `new Date(` is an unnecessary no-op wrapper in both cases

greatbacon: I tried that, but I think there's an issue with the JSON schema eating the underlying data. When I tried it without `new Date` I got a bunch of missing data errors on the front end.

ryanatkn: Ok yeah, something is weird here, the idea is that we're parsing the values from the schemas (JSON doesn't handle `Date` types, so the parsing turns them from strings into `Date`) but we must not be doing it in every case. The cases I checked before writing the above comment worked but there must be other cases.

ryanatkn, line 39: It might take a lot of thinking to get formatting exactly how we want it in the various cases, I don't mind merging as-is. [`formatDistanceToNow`](https://date-fns.org/v2.28.0/docs/formatDistanceToNow) looks useful for cases like these chat messages, but we may ultimately want more complicated logic like, "if it's older than a day then use `format(d, 'Pp')` like you have here.

ryanatkn (approved):
Awesome, glad it was easy. `created` is also used in `Home.svelte` if you wanted to format it. Making a note on our todo list to look into fixing date parsing.

src/lib/ui/RoomItem.svelte
ryanatkn: Seems entities are the exception, the session stuff has `Date` objects.

src/lib/ui/RoomItem.svelte
```diff
@@ -49,10 +52,13 @@ And then PersonaContextmenu would be only for *session* personas? `SessionPerson
 		/* TODO experiment with a border color instead of bg */
 		background-color: hsl(var(--hue), var(--bg_saturation), calc(var(--bg_color_lightness)));
 	}
-	.timestamp {
+	.signature {
 		display: flex;
 		align-items: center;
 	}
+	.timestamp {
+		padding-left: 60%;
```
ryanatkn, line 60: This causes some weirdness in some layouts, I think the better fix is adding `justify-content: space-between` to the `.workspace-header`.

src/lib/ui/WorkspaceHeader.svelte
```diff
@@ -42,6 +43,7 @@
 	}
 	.timestamp {
 		font-size: var(--font_size_md);
+		padding-left: 50%;
```
ryanatkn, line 46: This causes some weirdness in some layouts, I think the better fix is adding `justify-content: space-between` to the `.workspace-header`.

src/lib/ui/RoomItem.svelte
ryanatkn, line 60: Like with the `.workspace-header`, to fix this layout remove the padding and add `justify-content: space-between;` to `.signature`.

src/lib/ui/RoomItem.svelte
ryanatkn: Another option is we display `formatDistanceToNow` on the message, and put a `title` attribute it with the absolute formatting.

greatbacon: Cool! At least in the chat context I know a lot of apps will chunk by time & date. This is really mostly a quick fix to get the dates looking more human readable & space them out a little bit so they don't run into other data.

src/lib/ui/WorkspaceHeader.svelte
greatbacon: Thanks! I knew `padding` wasn't right, but my flex-box skills are weak. 😬

PR #232, opened 2022-01-28
refactor contextmenu declaration to use components instead of strings

ryanatkn:
This changes the contextmenu declarations to use Svelte components instead of strings. This means the data is now an array of tuples:

```svelte
use:contextmenu={[
  [AContextmenuComponent, propsA],
  [BContextmenuComponent, propsB]
]}
```

It feels slightly verbose -- we could consider going with a single array with alternating components/props, but it won't be extensible in the future if we want to add any other options as a third parameter. I can't think of a usecase for a third parameter though, so perhaps we should opt for this simpler version?

```svelte
use:contextmenu={[
  AContextmenuComponent, propsA,
  BContextmenuComponent, propsB
]}
```

The second option would be simpler and slightly better performance in components, at the cost of slightly more complex internal logic, and not being extensible in the future.

I'm keeping the contextmenu store and system-level components agnostic to everything in our app, so we'll be able to upstream it to Felt or extract as a standalone dependency. The main quirk here is that to get automatic link contextmenus, I needed to add the link component as a prop to `Contextmenu.svelte`.


greatbacon (approved):
Keeping things extensible seems fine to me.

PR #235, opened 2022-01-31
upgrade felt and use its upstreamed dialogs

ryanatkn:
Upgrades Felt to use the upstreamed `dialog.ts` and `Dialogs.svelte`: https://github.com/feltcoop/felt/pull/169

todo

- [x] declaratively extend the schema -- we use absolute references prefixed with `/schema/` now, which means we can import schemas from any project (this one uses `/schemas/DialogData.json` from Felt) and easily reference them local to the project, and still publish schemas relative to any root like `https://felt.social/` -- in the future we'll probably automatically publish schemas from the project root URL like this, giving us the best of both worlds of having them staticly/locally resolved at build time, and publishable to whatever URL, so schema dependencies like Felt are read from npm and imported like every other code dependency, not random web addresses, without requiring any path mapping
- [x] upgrade Felt and use the `Schema` suffixed version: https://github.com/feltcoop/felt/pull/170


src/lib/ui/ui.events.ts
```diff
@@ -47,14 +49,8 @@ export const events: EventInfo[] = [
 		type: 'ClientEvent',
 		name: 'OpenDialog',
 		params: {
+			...DialogState,
```
ryanatkn, line 52: Instead of doing this, I think the better way is to register the `DialogState` schema and use it like this:

```json
"allOf": [{ "$ref": "/schema/base" }],
```

src/lib/app/eventTypes.gen.ts
```diff
@@ -60,23 +86,9 @@ ${await eventInfos.reduce(
 	async (str, eventInfo) =>
 		(await str) +
 		`
-${await jsonSchemaToTypescript(eventInfo.params, toParamsName(eventInfo.name))}${
+${await jsonSchemaToTypescript(eventInfo.params, toParamsName(eventInfo.name), opts)}${
```
ryanatkn, line 89: This is the bug that may have caused our `EntityData` schema issue.

greatbacon (approved):

PR #240, opened 2022-02-02
refactor initial session and session updating to a single code path

ryanatkn:
This changes the `ui` initialization to start up with empty values, and then calls `ui.setSession` so there's a single code path.

I also renamed a few things for clarity.

- [x] fix the initialization problem on login (fixed by initializing objects directly from the session instead of using the derived stores; it's fine for now)
- [x] delete the logging

followup

- fix bug where the wrong community is initialized because it's not reading from the URL (not caused by this PR)

### DIFF
--- a/src/lib/ui/BoardItem.svelte
+++ b/src/lib/ui/BoardItem.svelte
@@ -10,12 +10,12 @@
 	import EntityContextmenu from '$lib/app/contextmenu/EntityContextmenu.svelte';
 
 	const {
-		ui: {contextmenu, findPersonaById},
+		ui: {contextmenu, personaById},
 	} = getApp();
 
 	export let entity: Readable<Entity>;
 
-	$: persona = findPersonaById($entity.actor_id); // TODO should this be `Actor` and `actor`?
+	$: persona = personaById.get($entity.actor_id)!; // TODO should this be `Actor` and `actor`?
 
 	// TODO refactor to some client view-model for the actor
 	$: hue = randomHue($persona.name);
--- a/src/lib/ui/CommunityNavButton.svelte
+++ b/src/lib/ui/CommunityNavButton.svelte
@@ -10,7 +10,7 @@
 
 	const {
 		dispatch,
-		ui: {contextmenu, spaceIdByCommunitySelection, findSpaceById, sessionPersonaIndices},
+		ui: {contextmenu, spaceIdSelectionByCommunityId, findSpaceById, sessionPersonaIndices},
 	} = getApp();
 
 	// TODO should this just use `ui` instead of taking all of these props?
@@ -20,9 +20,8 @@
 	export let community: Readable<Community>;
 	export let selected: boolean = false;
 
-	$: communitySelectionSpaceId = $spaceIdByCommunitySelection[$community.community_id];
-	$: communitySelectionSpace =
-		communitySelectionSpaceId === null ? null : findSpaceById(communitySelectionSpaceId);
+	$: spaceIdSelection = $spaceIdSelectionByCommunityId[$community.community_id];
+	$: selectedSpace = spaceIdSelection === null ? null : findSpaceById(spaceIdSelection);
 
 	$: isPersonaHomeCommunity = $community.name === $persona.name;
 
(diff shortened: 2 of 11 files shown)


src/lib/ui/BoardItem.svelte
```diff
@@ -10,12 +10,12 @@
 	import EntityContextmenu from '$lib/app/contextmenu/EntityContextmenu.svelte';
 
 	const {
-		ui: {contextmenu, findPersonaById},
+		ui: {contextmenu, personaById},
```
ryanatkn, line 13: Deleted two `find` helpers that we no longer need -- they predate the maps like `personaById`, and we should use the maps directly now.

src/lib/ui/CommunityNavButton.svelte
```diff
@@ -20,9 +20,8 @@
 	export let community: Readable<Community>;
 	export let selected: boolean = false;
 
-	$: communitySelectionSpaceId = $spaceIdByCommunitySelection[$community.community_id];
-	$: communitySelectionSpace =
-		communitySelectionSpaceId === null ? null : findSpaceById(communitySelectionSpaceId);
+	$: spaceIdSelection = $spaceIdSelectionByCommunityId[$community.community_id];
+	$: selectedSpace = spaceIdSelection === null ? null : $spaceById.get(spaceIdSelection)!;
```
ryanatkn, line 24: Still might want to prefix these with `community`, but I figured since we're local to the component it wasn't necessary, happy to change tho.

src/lib/ui/ui.ts
```diff
@@ -302,6 +282,58 @@ export const toUi = (
...
+					: Object.fromEntries(
+							$session.communities
+								.map(($community) => {
+									const $firstSpace = $spacesByCommunityId.get($community.community_id)?.[0];
+									return $firstSpace ? [$community.community_id, get($firstSpace).space_id] : null!;
+								})
+								.filter(Boolean),
+					  ),
+			);
+		},
```
ryanatkn, line 336: This diff is bad because I moved `setSession` up so the events are grouped together, and also moved up the data objects to the top that weren't already there, so things are co-located by type. Nothing substantial changed here except these two selection stores, which are causing problems on login.

src/lib/ui/ui.ts
```diff
@@ -261,25 +221,26 @@ export const toUi = (
...
 			spaces.update(($spaces) => $spaces.concat(spacesToAdd!.map((s) => writable(s))));
 		}
-		spaceIdByCommunitySelection.update(($spaceIdByCommunitySelection) => {
-			$spaceIdByCommunitySelection[community.community_id] = communitySpaces[0].space_id;
-			return $spaceIdByCommunitySelection;
-		});
+		spaceIdSelectionByCommunityId.update(($v) => ({
+			...$v,
+			[community.community_id]: communitySpaces[0].space_id,
+		}));
```
ryanatkn, line 237: This wasn't working because it wasn't immutable.

src/lib/ui/ui.ts
```diff
@@ -455,7 +434,6 @@ export const toUi = (
 		DeleteMembership: async ({params, invoke}) => {
 			const result = await invoke();
 			if (!result.ok) return result;
-			console.log('[ui.DeleteMembership]', params);
```
ryanatkn, line 458: I removed some logging where it was just logging params, because that's handled generically by the event dispatcher. In a followup PR I'll also make it log the result of `invoke` so we can delete those lines as well.

src/lib/ui/CommunityNavButton.svelte
greatbacon: I think the name is fine in this local context.

greatbacon (commented):
Alright, I looked through it. Lots of good cleanup! I'll hold off on approving though until you can get the selection store bug fixed.

ryanatkn:
Sweet, I fixed it (with a workaround -- we can't seem to use derived stores in `setSession`) but didn't update the paragraph, just the todo line item:

> (fixed by initializing objects directly from the session instead of using the derived stores; it's fine for now)

greatbacon (approved):

PR #239, opened 2022-02-01
Refactor space content to view

greatbacon:
This PR refactors the space `content` column into a column named `view` of type JSONB.

It now uses the previously created ViewData type around the system.

I know we had talked about moving the `media_type` column into `view`. I'm going to do that in a follow up PR though.

PR #241, opened 2022-02-02
Remove `media_type` from Spaces

greatbacon:
This is a follow up to #239 that drops the superfluous `media_type` column. If we need something like it in the future we will just add it to the `view` object.


ryanatkn (approved):
# :tada:

PR #242, opened 2022-02-02
Fixes a bug in the restart logic

greatbacon:
Quick fix in the infra tasks to give us remote server restart. Helps with deployments :D


ryanatkn (approved):
v nce

PR #244, opened 2022-02-04
add communityById map

ryanatkn:
- add `communityById` map and align it with `personaById`
- rename some objects local to `ui.ts` to prefix with the `$` convention

future:

- maybe have mutable stores wrapping maps instead of plain arrays for `communities` and `personas`? with `mutable`?


src/lib/ui/ui.ts
```diff
@@ -555,12 +566,3 @@ const toInitialPersonas = (session: ClientSession): Persona[] =>
...
 		  );
-
-// TODO delete this and lookup from `communityById` map instead
-export const getCommunity = (
-	communities: Readable<Community>[],
-	community_id: number,
-): Writable<Community> =>
-	// TODO typecast allows `Readable` input and `Writable` return value for usage in components,
-	// but this function will be deleted soon anyway (see the comment above)
-	communities.find((c) => get(c).community_id === community_id) as Writable<Community>;
```
ryanatkn, line 566: # :wave:

src/lib/ui/ui.ts
```diff
@@ -209,33 +208,40 @@ export const toUi = (
...
 		memberships.update(($memberships) =>
 			$memberships.concat(
-				writable({community_id: community.community_id, persona_id} as Membership),
+				writable({community_id: $community.community_id, persona_id} as Membership),
 			),
 		);
 
+		// TODO what's the right order of updating `communities` and `spaces`?
+		// We may get circular derived dependencies that put things in a bad state if either one is
+		// updated first, in which case we may need something like deferred store transaction updates.
```
ryanatkn, line 224: pretty sure this will be a source of bugs before long

src/lib/ui/ui.ts
```diff
@@ -284,12 +291,22 @@ export const toUi = (
...
 			personaById.clear();
-			get(personas).forEach((persona) => personaById.set(get(persona).persona_id, persona));
+			$personas.forEach((p, i) => personaById.set($$personas[i].persona_id, p));
+			personas.set($personas);
+
 			const $sessionPersonas = $session.guest ? [] : $session.personas;
 			sessionPersonas.set($sessionPersonas.map((p) => personaById.get(p.persona_id)!));
-			communities.set($session.guest ? [] : $session.communities.map((p) => writable(p)));
+
+			const $$communities = $session.guest ? [] : $session.communities;
```
ryanatkn, line 304: Slightly strange naming convention here and above with `$$personas`, but it allows us to avoid the hacky use of `get(store)` in loops. Perhaps `$communitiesArray` or `$communitiesItems` or `$communityItems`? Not sure I love any other the possibilities, though maybe the `Array` suffix isn't so bad? Or is `$$` fine?

greatbacon: I can't speak to performance or forking, but as far as naming goes using the `Array` suffix maxes sense to me.

greatbacon (approved):

PR #250, opened 2022-02-08
add gro task schemas

ryanatkn:
Adds schemas for each of our tasks that have args, so we'll get typesafety and documentation on the args, and soon, automatic defaults and arg validation. They're implemented as task directories which were [added here in Gro](https://github.com/feltcoop/gro/pull/311). Here's the [task directory docs](https://github.com/feltcoop/gro/blob/main/src/docs/task.md#task-directories) and the [task args docs](https://github.com/feltcoop/gro/blob/main/src/docs/task.md#task-args).

See the new behavior by running `gro` or `gro help` or `gro --help`, and also `gro check --help` to see how we extend the base help text with the `migrations` arg.

Gro currently only supports plain schemas with `properties` for its task args, not composed schemas using `allOf` and references. That'll require additional tooling, which shouldn't be too hard, but it's easy enough to extend schemas manually for now. (see how our check task extends the base Gro check schema)


src/check.task.ts
```diff
@@ -1,24 +1,31 @@
...
+						'Migration named index does not match its directory index.' +
+							` Expected index '${i}' but got '${migrationIndex}' for file ${migrationFile}`,
+					);
+				}
 			}
 		}
 
 		// Perform default checks.
-		await invokeTask('gro/check');
+		await invokeTask('gro/check', {_: [], ...restArgs});
```
ryanatkn, line 29: The `_: []` shouldn't be needed but Gro currently requires it. I'll see if it's easy to change.

greatbacon (approved):

PR #251, opened 2022-02-08
fix layout scrolling

ryanatkn:
Fixes the main layout scrollbars when the space is taller than the viewport (happens on the `Home` view on smaller screens) along with some other minor issues, like using `403` instead of `401`.

### DIFF
--- a/src/lib/ui/Workspace.svelte
+++ b/src/lib/ui/Workspace.svelte
@@ -39,12 +39,14 @@
 	<div class="column">
 		<!-- TODO pass stores here instead of dereferncing -->
 		<WorkspaceHeader space={selectedSpace} community={selectedCommunity} />
-		{#if selectedPersona && selectedCommunity && selectedSpace}
-			<SpaceView persona={selectedPersona} community={selectedCommunity} space={selectedSpace} />
-		{:else if selectedPersona && selectedCommunity}
-			<!-- TODO this should be the form, not dialog trigger -- fix after refactoring dialogs -->
-			<SpaceInput persona={selectedPersona} community={selectedCommunity} />
-		{/if}
+		<div class="content">
+			{#if selectedPersona && selectedCommunity && selectedSpace}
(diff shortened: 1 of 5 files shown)


src/lib/ui/view/Home.svelte
```diff
@@ -24,7 +25,7 @@
 	$: communityPersonas = $personasByCommunityId.get($community.community_id)!;
 </script>
 
-<div class="markup">
+<div class="home">
```
ryanatkn, line 28: This gets rid of the bullet points.

src/lib/ui/WorkspaceHeader.svelte
```diff
@@ -7,14 +7,18 @@
 	import {format} from 'date-fns';
 
 	const {
-		ui: {expandMainNav},
+		ui: {expandMainNav, expandMarquee},
```
ryanatkn, line 10: The code here copies the same pattern we use for the luggage button but for the marquee on the right side. Without it the marquee button overlaps the timestamp when it's closed.

src/lib/ui/Workspace.svelte
```diff
@@ -36,15 +36,17 @@
...
+	<div class="space column">
 		<!-- TODO pass stores here instead of dereferncing -->
 		<WorkspaceHeader space={selectedSpace} community={selectedCommunity} />
-		{#if selectedPersona && selectedCommunity && selectedSpace}
-			<SpaceView persona={selectedPersona} community={selectedCommunity} space={selectedSpace} />
-		{:else if selectedPersona && selectedCommunity}
-			<!-- TODO this should be the form, not dialog trigger -- fix after refactoring dialogs -->
-			<SpaceInput persona={selectedPersona} community={selectedCommunity} />
-		{/if}
+		<div class="content">
```
ryanatkn, line 42: This `.content` wrapper element now sets the bounds for the loaded space, controlling its scrolling.

greatbacon: Content makes sense to me.

PR #253, opened 2022-02-09
Add checkable `Tasks`

greatbacon:
This PR adds a new view of type `Task` as well as a backing `Task` entity type. 

I referred to them as Todos in the doc, not Tasks. I'm flexible on any names right now.

It also adds an relatively generic `UpdateEntityData` event to the server as well.

It still has a couple of outstanding things in it, but I've reached the point where all the issues are front end so I wanted to get it in front of you.
- [x] Fix type error in ui.ts
- [x] Double check EntityData typing
- [x] Fix checkbox not broadcasting
- [x] Add tests


src/lib/vocab/entity/entity.events.ts
```diff
@@ -31,6 +31,36 @@ export const CreateEntity: ServiceEventInfo = {
...
+		type: 'object',
+		properties: {
+			entity_id: {type: 'number'},
+			data: {type: 'object', tsType: 'EntityData'},
+		},
+		required: ['entity_id', 'data'],
+		additionalProperties: false,
+	},
+	response: {
+		$id: '/schemas/UpdateEntityResponse.json',
```
ryanatkn, line 49: `UpdateEntity` looks like a good event to add for now, but it probably won't be used long term because many actions will affect a subset of data (like just `content`), and/or multiple entities and ties. So we'll get a lot of mileage out of a generic endpoint that can return an arbitrary number of changed entities/ties and update them through a single endpoint.

Because more granular and complex actions will be handled through the activity endpoint (anything that affects the ties or multiple entities or a subset of properties), I think we should treat this endpoint as simple as possible - "here's some data, store it". So the `response` schema here could be `type: 'null`', the repo wouldn't have a `RETURNING *` clause, and the service would return `null`.

src/lib/vocab/entity/entity.events.ts
```diff
@@ -31,6 +31,36 @@ export const CreateEntity: ServiceEventInfo = {
...
+		type: 'object',
+		properties: {
+			entity: {$ref: '/schemas/Entity.json', tsType: 'Entity'},
+		},
+		required: ['entity'],
+		additionalProperties: false,
+	},
+	returns: 'Promise<UpdateEntityResponseResult>',
+	route: {
+		path: '/api/v1/spaces/:space_id/entities/:entity_id',
```
ryanatkn, line 59:
```suggestion
		path: '/api/v1/entities/:entity_id',
```

ryanatkn: The HTTP endpoint fails because of this, you can test by setting `websockets: false` on the `UpdateEntity` in `entity.events.ts`. Smoke tests could catch this sort of thing. (we'd need to generate the randomizers from schemas first)

```
validation failed: {
  entity_id: 64,
  data: { done: false, type: 'Task', content: '1213' },
  space_id: NaN
} [
  {
    instancePath: '',
    schemaPath: '#/additionalProperties',
    keyword: 'additionalProperties',
    params: { additionalProperty: 'space_id' },
    message: 'must NOT have additional properties'
  }
]
```

src/lib/vocab/random.ts
```diff
@@ -132,6 +142,18 @@ export const toRandomVocabContext = (db: Database): RandomVocabContext => {
...
+			if (!persona) persona = await random.persona(account);
+			if (!community) community = await random.community(persona, account);
+			if (!space) space = await random.space(persona, account, community);
+			const params = randomEntityParams(persona.persona_id, space.space_id);
+			const entity = unwrap(
+				await db.repos.entity.create(params.actor_id, params.space_id, params.data),
+			);
+			random.entities.push(entity);
+			return entity;
+		},
```
ryanatkn, line 156: # :clap: 

lol that's a mouthful

ryanatkn (approved):

src/lib/vocab/view/view.ts
```diff
@@ -17,4 +17,5 @@ export const availableViewTypes: string[] = [
 	'Voice',
 	'Iframe',
 	'EntityExplorer',
+	'Todo',
```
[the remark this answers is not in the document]
greatbacon: None of our other space views have a plural name, so I'm inclined to stick to convention?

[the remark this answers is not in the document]
ryanatkn: makes sense, also "Todo space" rolls off the tongue where "Todos space" does not

PR #259, opened 2022-02-15
add entity editor

ryanatkn:
Adds a contextmenu entry for the entity with an `Edit Entity` option that opens a dialog to a new `EntityEditor.svelte`. It has some stubbed out functionality with functional editing for the `content` property, and if `$devmode` is enabled there's a table with  stubbed stuff with editable raw JSON data.

The width of the dialog form is a little janky because it grows as content changes. Not sure of the best way to fix it that also works on mobile.


src/lib/app/contextmenu/EntityContextmenu.svelte
```diff
@@ -2,10 +2,26 @@
...
+
+	const {dispatch} = getApp();
 </script>
 
-<li role="none" class="markup">
-	<p>TODO use entity_id: {$entity.entity_id}</p>
-</li>
+<ContextmenuSubmenu>
+	<svelte:fragment slot="entry">
+		<span class="menu-item-entry"><UnicodeIcon icon="☉" /><span class="title">Entity</span></span>
```
ryanatkn, line 18: first thing that I saw from https://wikipedia.org/wiki/Miscellaneous_Symbols

greatbacon (approved):
Cool, awesome initial work! No worries on the dialog jank, we can always fix that in another pass.

PR #262, opened 2022-02-15
add contextmenu persona selection

ryanatkn:
Adds persona selection to the contextmenu. Semantic HTML could use improvement but the concept is implemented.

### DIFF
--- a/src/lib/app/contextmenu/AppContextmenu.svelte
+++ b/src/lib/app/contextmenu/AppContextmenu.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+	import {get} from 'svelte/store';
+
 	import {getApp} from '$lib/ui/app';
 	import ContextmenuEntry from '$lib/ui/contextmenu/ContextmenuEntry.svelte';
 	import ContextmenuSubmenu from '$lib/ui/contextmenu/ContextmenuSubmenu.svelte';
@@ -7,8 +9,12 @@
 	import About from '$lib/ui/About.svelte';
 	import {session} from '$app/stores';
 	import AccountForm from '$lib/ui/AccountForm.svelte';
+	import PersonaAvatar from '$lib/ui/PersonaAvatar.svelte';
 
-	const {dispatch} = getApp();
+	const {
+		dispatch,
+		ui: {sessionPersonas, personaSelection},
+	} = getApp();
 </script>
 
 <ContextmenuSubmenu>
@@ -18,6 +24,26 @@
 		>
 	</svelte:fragment>
 	<svelte:fragment slot="menu">
+		{#each $sessionPersonas as sessionPersona (sessionPersona)}
+			{#if $personaSelection === sessionPersona}
+				<li class="menu-item panel-inset" role="none">
+					<PersonaAvatar persona={sessionPersona} />
+				</li>
+			{:else}
+				<!-- TODO support store param? only? -->
+				<ContextmenuEntry
+					action={(e) => {
+						if (e) {
+							e.stopPropagation();
+							e.preventDefault();
+						}
+						dispatch('SelectPersona', {persona_id: get(sessionPersona).persona_id});
+					}}
+				>
+					<PersonaAvatar persona={sessionPersona} />
+				</ContextmenuEntry>
+			{/if}
+		{/each}
 		<ContextmenuEntry
 			action={() =>
 				dispatch('OpenDialog', {
@@ -35,8 +61,10 @@
 		>
 			<span class="title">About</span>
 		</ContextmenuEntry>
-		<li role="none">
-			<AccountForm guest={$session.guest} />
+		<li role="none" on:click|stopPropagation>
+			<div>
+				<AccountForm guest={$session.guest} />
+			</div>
 		</li>
 	</svelte:fragment>
 </ContextmenuSubmenu>
--- a/src/lib/app/contextmenu/AppContextmenu.svelte
+++ b/src/lib/app/contextmenu/AppContextmenu.svelte
@@ -55,10 +55,12 @@
 		>
 			<span class="title">About</span>
 		</ContextmenuEntry>
-		<li role="none" on:click|stopPropagation>
-			<div>
-				<AccountForm guest={$session.guest} />
-			</div>
-		</li>
+		{#if !$session.guest}
+			<li role="none">
+				<div>
+					<AccountForm guest={$session.guest} />
+				</div>
+			</li>
+		{/if}
 	</svelte:fragment>
 </ContextmenuSubmenu>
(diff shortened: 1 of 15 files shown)


src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -18,6 +24,26 @@
...
+					<PersonaAvatar persona={sessionPersona} />
+				</li>
+			{:else}
+				<!-- TODO support store param? only? -->
+				<ContextmenuEntry
+					action={(e) => {
+						if (e) {
+							e.stopPropagation();
+							e.preventDefault();
+						}
```
ryanatkn, line 39: Stopping the event like this means the contextmenu remains open, whereas normally it'll close if it detects a button or menuitem being pressed. I needed to add the functionality for `action` to take the event. The event is a new optional param, so action triggers don't actually need to provide an event.

src/lib/ui/PersonaAvatar.svelte
```diff
@@ -0,0 +1,11 @@
...
+	import {type Readable} from 'svelte/store';
+
+	import Avatar from '$lib/ui/Avatar.svelte';
+	import {toName, toIcon} from '$lib/vocab/entity/entityHelpers';
+	import {type Persona} from '$lib/vocab/persona/persona';
+
+	export let persona: Readable<Persona>;
+</script>
+
+<Avatar name={toName($persona)} icon={toIcon($persona)} />
```
ryanatkn, line 11: followup PR will use this avatar elsewhere

src/lib/ui/PersonaAvatar.svelte
```diff
@@ -0,0 +1,13 @@
...
+	import Avatar from '$lib/ui/Avatar.svelte';
+	import {toName, toIcon} from '$lib/vocab/entity/entityHelpers';
+	import {type Persona} from '$lib/vocab/persona/persona';
+
+	export let persona: Readable<Persona>;
+	export let showName = true;
+	export let showIcon = true;
+</script>
+
+<Avatar name={toName($persona)} icon={toIcon($persona)} {showName} {showIcon} />
```
ryanatkn, line 13: will use this elsewhere in a followup PR

src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -18,6 +24,26 @@
 		>
 	</svelte:fragment>
 	<svelte:fragment slot="menu">
+		{#each $sessionPersonas as sessionPersona (sessionPersona)}
+			{#if $personaSelection === sessionPersona}
+				<li class="menu-item panel-inset" role="none">
```
ryanatkn, line 29: these selection visuals are minimal -- many things could work in the design language here, and "selected" already means something on a menuitem, so perhaps the avatar itself is selected or something

src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -18,6 +24,23 @@
...
+					<PersonaAvatar persona={sessionPersona} />
+				</li>
+			{:else}
+				<!-- TODO support store param? only? -->
+				<ContextmenuEntry
+					action={(e) => {
+						e?.stopPropagation();
```
ryanatkn, line 36: Stopping the event like this means the contextmenu remains open, whereas normally it'll close if it detects a button or menuitem being pressed. I needed to add the functionality for `action` to take the event. The event is a new optional param, so action triggers don't actually need to provide an event.

src/lib/ui/PersonaInput.svelte
```diff
@@ -45,7 +45,7 @@
 </script>
 
 <div class="markup">
-	<h2>Create a persona</h2>
+	<h1>Create a Persona</h1>
```
ryanatkn, line 48: thinking about what it might take to linkify words like `Persona` here, civiclopedia-style

src/lib/ui/style.css
```diff
@@ -26,13 +26,27 @@ main {
 	width: 100%;
 }
 
+/* TODO upstream to Felt? what about other simple layout helpers? */
+.row {
+	width: 100%;
+	display: flex;
+	align-items: center;
+}
```
ryanatkn, line 34: This pattern keeps recurring so I'm opening the door to some simple layout primitives. We already have `.column` giving things a `max-width`, so `.row` seems like a good step forward.

src/lib/ui/style.css
```diff
@@ -26,13 +26,27 @@ main {
...
 /* TODO refactor or upstream -- `navbar_size` maybe should be `input_height` */
 .icon-button {
 	width: var(--navbar_size);
 	height: var(--navbar_size);
 }
 
 /* TODO refactor or upstream */
+[role='menuitem'] {
+	cursor: pointer;
+}
```
ryanatkn, line 49: I like this because it ties an a11y concern to the mouse users UX, so the mouse-using devs will notice

src/lib/ui/PersonaAvatar.svelte
```diff
@@ -0,0 +1,20 @@
...
+	export let showName = true;
+	export let showIcon = true;
+</script>
+
+<Avatar
+	name={toName($persona)}
+	icon={toIcon($persona)}
+	{showName}
+	{showIcon}
+	contextmenuAction={[[PersonaContextmenu, {persona}]]}
```
ryanatkn, line 19: this feels great right here

src/lib/app/contextmenu/AppContextmenu.svelte
greatbacon: Is there a reason the menu shouldn't close when changing personas? I can't really think of examples of this behavior in the wild.

ryanatkn: Hmm no lol there's no good reason. A bug currently causes it to sometimes close and sometimes stay open and I think I just assumed staying open is the better behavior, but on second thought I agree it shouldn't. Undoing this change adding `event` to the action params since we don't have a usecase. I'll see if the bug is easy to fix.

src/lib/ui/contextmenu/ContextmenuEntry.svelte
```diff
@@ -21,11 +21,15 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.
...
+<!-- TODO This calls `action` after a tick to let event handlers react
+	to the current DOM before the action's changes are applied,
+	but it won't work if we need to forward the event to handlers.
+-->
 <li
 	class="menu-item"
 	role="menuitem"
 	class:selected
-	on:click={() => action()}
+	on:click={() => setTimeout(() => action())}
```
ryanatkn, line 32: This fixes the contextmenu sometimes not closing. The issue was that the contextmenu rerendered immediately with the newly selected persona, causing the bubbling event to have a target that's no longer mounted to the DOM, so the query to know whether it should close the contextmenu failed. The problem as described in the comment is that it won't work for some behaviors, like if contextmenu elements need to intercept the event with their own logic, but we can worry about those usecases when they appear.

greatbacon (approved):

PR #264, opened 2022-02-15
Add ties

greatbacon:
This PR adds the DB table `Ties` (used for linking entities in a graph). It also adds supporting vocab scaffolding, and a service/event for creating new ties. The UX does not currently actually call this event anywhere though.

It also has some additional extension to the Todo view which allows for the creation of lists.

I think I've also got a working query for Ties, but I wanted to get this code reviewed first before going too deep on the query stuff.


src/lib/db/migrations/00014-create-ties.js
```diff
@@ -0,0 +1,12 @@
+/** @param {import('postgres').Sql<any>} sql */
+export const up = async (sql) => {
+	await sql`
+	CREATE TABLE IF NOT EXISTS ties (
+		source_id int,
+		destination_id int,
+		type text,
```
ryanatkn, line 7: Constraints for these? Thinking id foreign keys and index on `type`

ryanatkn: Also alternative naming ideas:

`tie.from_id` `tie.to_d`

`tie.a_id` `tie.b_d`

`tie.source_id` `tie.target_d`

src/lib/ui/ui.ts
```diff
@@ -517,6 +517,12 @@ export const toUi = (
 			}
 			return spaceEntities;
 		},
+		CreateTie: async ({invoke}) => {
+			const result = await invoke();
+			if (!result.ok) return result;
+			//TODO figure out front end state for Ties
+			return result;
```
ryanatkn, line 524: notes to discuss:

- `ties` array in a `mutable`
- maps by each of the tie's 3 datapoints
- arrays/maps by entity id containing all ties to/from the entity

src/lib/ui/TodoItems.svelte
```diff
@@ -1,15 +1,20 @@
...
 
 <!-- TODO possibly remove the `ul` wrapper and change the `li`s to `div`s -->
 <ul>
 	{#each $entities as entity (entity)}
-		<TodoItem {entity} />
+		{#if get(entity).data.type === 'Note'}
+			<TodoItem {entity} />
+		{:else if get(entity).data.type === 'Collection'}
+			<CollectionItem {entity} />
+		{/if}
```
ryanatkn, line 18: This `get` usage isn't reactive and should generally be avoided in components. One easy fix is to move this `#if` check to the `TodoItem.svelte`, so the `TodoItem.svelte` would compose a `CollectionItem.svelte`.

src/lib/vocab/tie/tie.events.ts
```diff
@@ -0,0 +1,33 @@
...
+		type: 'object',
+		properties: {
+			tie: {$ref: '/schemas/Tie.json', tsType: 'Tie'},
+		},
+		required: ['tie'],
+		additionalProperties: false,
+	},
+	returns: 'Promise<CreateCommunityResponseResult>',
+	route: {
+		path: '/api/v1/communities',
```
ryanatkn, line 28: ties

ryanatkn, line 28:
```suggestion
		path: '/api/v1/ties',
```

src/lib/ui/view/Todo.svelte
```diff
@@ -24,9 +26,17 @@
 		const content = text.trim(); // TODO parse to trim? regularize step?
 
 		if (!content) return;
+
+		let data: EntityData;
+		if (list) {
+			data = {type: 'Collection', content, name: content};
+		} else {
+			data = {type: 'Note', content, name: content, checked: false};
+		}
```
ryanatkn, line 35:
```suggestion
		const data: EntityData = list
			? {type: 'Collection', content, name: content}
			: {type: 'Note', content, name: content, checked: false};
```

ryanatkn: Was the `name: content` for the `Note` intention?

src/lib/db/migrations/00014-create-ties.js
greatbacon: Do you dislike `source`/`destination` or are you just floating options?

src/lib/ui/view/Todo.svelte
greatbacon: Yes although I'm not attached to it. I was imagining a world where you do like quick notes, but then you might want a view later where you click in and edit the body and it would somehow be a better UX because *waves hands*. But now that I type this out I think it would be something that could easily be done w/ front end magic.

src/lib/db/migrations/00014-create-ties.js
ryanatkn: Floating options, not sure what to think of source/destination. Seems worth chewing on, maybe let's see what we're thinking on Thursday?

ryanatkn:
Brainstorming, if we currently have ☉ for entities, what should the tie symbol be? Looking here: https://en.wikipedia.org/wiki/Miscellaneous_Symbols

some siblings to the entity symbol: ☊ | ☋ | ☌ | ☍

src/lib/db/migrations/00014-create-ties.js
ryanatkn: Trying to think ahead to repo methods, first thought:

`repos.ties.filterTiesFrom`

`repos.ties.filterTiesTo`

`repos.ties.filterByType`

`repos.ties.filter({from_id, to_id, type})`

vs

`repos.ties.filterBySource`

`repos.ties.filterByDestination`


vs

`repos.ties.filterBySubject`

`repos.ties.filterByObject`

greatbacon: Having done some follow on work re: querying, most stuff I've found in context of graph walking seems to use a `source`/`destination` semantic (well, `src`/`dest`). I'm leaning towards sticking with that (although we could shorten `destination` to `dest`) or `source`/`target`.

src/lib/ui/ui.ts
greatbacon: For the purposes of this PR, are you okay with punting this work to a future PR?

greatbacon:
> Brainstorming, if we currently have ☉ for entities, what should the tie symbol be? Looking here: https://en.wikipedia.org/wiki/Miscellaneous_Symbols
> 
> some siblings to the entity symbol: ☊ | ☋ | ☌ | ☍
> 
> what if those map to `source` | `dest` | `type` | `tie`
> 
> Emoji are an option, like we use for Spaces. Though maybe it's best to keep the symbols for entities/ties more abstract and not full-color emoji?

I'm not actually sure if we should be directly exposing ties as something to interact with for users. To me it almost feels like something that should be managed through lens of entities.

src/lib/db/migrations/00014-create-ties.js
ryanatkn: I'm good with `source`/`dest`

src/lib/db/migrations/00014-create-ties.js
```diff
@@ -0,0 +1,17 @@
...
+		destination_id int REFERENCES entities (entity_id) ON UPDATE CASCADE ON DELETE CASCADE,
+		type text,
+		created timestamp NOT NULL DEFAULT now(),
+		updated timestamp		
+	)
+	`;
+
+	await sql`
+		CREATE
+		INDEX ON ties (LOWER(type));
```
ryanatkn, line 15:
```suggestion
		INDEX ON ties (type);
```

src/lib/db/migrations/00014-create-ties.js
ryanatkn: I'm getting hung up on this trying to formulate sentences without concrete examples -- what should the collection type be for example? `'Item'`? `'Contains'`? Something else? Might help me to talk through the queries too.

ryanatkn:
Makes sense, yeah ties are normally implicit to the UI. We'll display them at some depth of popping the hood, like the entity explorer.

ryanatkn (approved):

PR #266, opened 2022-02-18
Add default sort by created for entities

greatbacon:
Quick PR that sorts entities by their `created` timestamp in ascending order (i.e 0th index is chronologically first).

### DIFF
--- a/src/lib/vocab/entity/entityRepo.test.ts
+++ b/src/lib/vocab/entity/entityRepo.test.ts
@@ -21,6 +21,8 @@ test__entityRepo('entites return sorted by created', async ({db}) => {
 	assert.equal(entity0.space_id, entity1.space_id);
 	assert.equal(entity1.space_id, entity2.space_id);
 
+	// Ensure db sort order is shuffled from the insertion order.
+	await db.repos.entity.updateEntityData(entity1.entity_id, entity1.data);
 	const result = await db.repos.entity.filterBySpace(entity0.space_id);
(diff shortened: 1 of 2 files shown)


ryanatkn (approved):

src/lib/vocab/entity/entityRepo.test.ts
```diff
@@ -0,0 +1,32 @@
...
+test__entityRepo('entites return sorted by created', async ({db}) => {
+	const random = toRandomVocabContext(db);
+	const space = await random.space();
+	const entity0 = await random.entity(undefined, undefined, undefined, space);
+	const entity1 = await random.entity(undefined, undefined, undefined, space);
+	const entity2 = await random.entity(undefined, undefined, undefined, space);
+
+	assert.equal(entity0.space_id, entity1.space_id);
+	assert.equal(entity1.space_id, entity2.space_id);
+
```
ryanatkn, line 23:
```suggestion

	// Ensure db sort order is shuffled from the insertion order.
	await db.repos.entity.updateEntityData(entity1.entity_id, entity1.data);
```

Looks like the test passes without the new `ORDER BY` change. This makes sure we're not relying on insertion order.

ryanatkn:
This PR looks gtg to preserve existing UI behavior. In a future PR we'll probably make the default DESC, because most views probably load the most recent stuff, but that'll require sorting on the clients. I think it'll be relevant when we're implementing paging/history.

PR #268, opened 2022-02-28
quieter logging

ryanatkn:
Converts everything to Felt's logger. Tests now only log `info` and above.

- [x] fix logger's reliance on `process` -- https://github.com/feltcoop/felt/pull/180

future:

- proper importing of `testHelpers` instead of hackily relying on tests importing it (need to use uvu's `-r` with esm)
- probably add an env var like `VITE_LOG_LEVEL`
- possibly parameterize loggers in places like on the db, with a log extending API


src/lib/util/testHelpers.ts
```diff
@@ -1,4 +1,7 @@
 import sourcemapSupport from 'source-map-support';
+import {configureLogLevel, LogLevel} from '@feltcoop/felt/util/log.js';
+
+configureLogLevel(LogLevel.Info);
```
ryanatkn, line 4: This sets the minimum log level for tests. Calls to `log.trace` will now be no-ops.

src/lib/ui/init.ts
```diff
@@ -0,0 +1,15 @@
...
+	Logger.prototype.error = console.error.bind(console);
+	Logger.error.prefixes.length = 0;
+	Logger.error.suffixes.length = 0;
+	Logger.warn.prefixes.length = 0;
+	Logger.warn.suffixes.length = 0;
+	Logger.info.prefixes.length = 0;
+	Logger.info.suffixes.length = 0;
+	Logger.trace.prefixes.length = 0;
+	Logger.trace.suffixes.length = 0;
+};
```
ryanatkn, line 15: Hacky function that makes the logger play nicely in the browser. I'll think about how to redesign it to avoid needing this hack.

greatbacon (approved):
Thanks for cleaning up the mean streets of logging! 🧹 💨

PR #270, opened 2022-03-04
implement space view as SVAST

ryanatkn:
Changes the `space.view` to be a SVAST from [MDsveX](https://github.com/pngwn/MDsveX). Views no longer have a type, and instead `space.view` is the parsed SVAST JSON -- functionally, this means spaces are no longer tied to a single component, and arbitrary composition with Svelte syntax is now possible. All of our code uses the `$lib/vocab/view/view` module instead of importing SVAST-related things directly.

A view `template` is the stringified version of a `view` as currently implemented, e.g. `'<Room />'` -- this is used only in the frontend for the user-friendly list of available spaces. We can expand the available templates without any backend changes.

Currently, only Svelte components are rendered from view data; everything else is ignored.

Several things will be cleaned up when we add `space.icon`. (am midway through that PR now)


src/lib/ui/SpaceIcon.svelte
```diff
@@ -20,8 +21,9 @@
 	const DEFAULT_ICON = '🖊';
 	const DEFAULT_LABEL = 'Space';
 
-	$: icon = spaceTypeIcons[$space.view.type] ?? DEFAULT_ICON;
-	$: label = icon === DEFAULT_ICON ? DEFAULT_LABEL : $space.view.type;
+	$: type = toViewType($space.view);
+	$: icon = type === undefined || !(type in spaceTypeIcons) ? DEFAULT_ICON : spaceTypeIcons[type];
+	$: label = type === undefined ? DEFAULT_LABEL : type;
 </script>
```
ryanatkn, line 27: All of this code will be deleted once we add `space.icon`.

src/lib/vocab/view/view.ts
```diff
@@ -1,21 +1,99 @@
...
+
+export interface ViewContext {
+	persona: Readable<Persona>;
+	community: Readable<Community>;
+	space: Readable<Space>;
+}
+
+const KEY = Symbol();
+export const getViewContext = (): Readable<ViewContext> => getContext(KEY);
+export const setViewContext = (ctx: Readable<ViewContext>): void => setContext(KEY, ctx);
```
ryanatkn, line 85: The `ViewContext` seems like a good idea right now because it lets `SvastView.svelte` be agnostic to all of our app-specific stuff, but I'm not sure it'll be the best choice when we implement composition with child views. Threading props may be desired in that case; we can see when we get there.

ryanatkn, line 85: The `ViewContext` seems like a good idea right now because it lets `SvastView.svelte` be agnostic to all of our app-specific stuff, but I'm not sure it'll be the best choice when we implement composition with child views. Threading props may be desired in that case; we can see when we get there.

src/lib/vocab/view/view.ts
```diff
@@ -1,21 +1,99 @@
...
+			const v = prop.value[0];
+			if (v?.type === 'text') {
+				(props || (props = {}))[prop.name] = v.value;
+			}
+		}
+	}
+	return props;
+};
+
+// TODO delete this after adding icon to view templates (see also the above TODO)
```
ryanatkn, line 51: This is deleted in #273

PR #271, opened 2022-03-09
Adds event logic for deleting a single tie

greatbacon:
Quick PR that adds an event for deleting a single tie.

No concrete use case for it today, but we're filling out the CRUD actions on our Vocab resources.


ryanatkn:
This deletes all types of ties between two entities, that's not the intention right? We could add the `type` param as well to properly scope it, but I'm wondering if we instead want a `tie_id`? Thinking ahead it may be something that we'll need for a lot of reasons, but I'm not sure.

greatbacon:
> This deletes all types of ties between two entities, that's not the intention right? We could add the `type` param as well to properly scope it, but I'm wondering if we instead want a `tie_id`? Thinking ahead it may be something that we'll need for a lot of reasons, but I'm not sure.

I was operating off the assumption that there wouldn't be multiple types of ties between two entities since no examples of where that might be the case immediately sprang to mind. I can add `type` to scope it though. We can also assert that ties are unique on those lines by making those 3 attributes function as a Primary Key

ryanatkn:
Adding `type` sounds good -- we already have the unique constraint and after a bit of reading, it seems it functions similarly to a composite primary key, and the main relevant difference is that composite primary keys use clustered indexes instead of unique's non-clustered. Read performance is significantly better if we use a clustered index, so it may be a good idea to change the existing unique constraint to a composite primary key.

Kind of a fun thought experiment given we have only 3 types so far, here's a possible usecase that would have both replies and items between the same 2 entities in a custom view:

- user makes a post asking for replies for possible options
- other users reply to the post
- user chooses the best replies as items to the original post, so the post can display a summary collection

This made me realize that I don't know how we'll support ordered collections with ties. Maybe in a metadata field or `Collection` entity?

src/lib/vocab/tie/tie.events.ts
```diff
@@ -58,4 +58,29 @@ export const ReadTies: ServiceEventInfo = {
...
+		required: ['source_id', 'dest_id'],
+		additionalProperties: false,
+	},
+	response: {
+		$id: '/schemas/DeleteTieResponse.json',
+		type: 'null',
+	},
+	returns: 'Promise<DeleteTieResponseResult>',
+	route: {
+		path: '/api/v1/spaces/:space_id/ties',
```
ryanatkn, line 81: I think `/api/v1/ties` should be the path

src/lib/vocab/tie/tieServices.ts
```diff
@@ -49,3 +51,18 @@ export const readTiesService: Service<ReadTiesParams, ReadTiesResponseResult> =
 		return {ok: false, status: 500, message: 'error searching for entities'};
 	},
 };
+
+//deletes a single tie
+export const deleteTieService: Service<DeleteTieParams, DeleteTieResponseResult> = {
```
ryanatkn, line 56: This needs to be added to the array in `$lib/server/services.ts`.

src/lib/server/random.ts
```diff
@@ -146,6 +146,16 @@ export const randomEventParams = async (
...
+		case 'DeleteTie': {
+			return {
+				source_id: (
+					randomItem(random.entities) || (await random.entity(persona, account, community, space))
+				).entity_id,
+				dest_id: (
+					randomItem(random.entities) || (await random.entity(persona, account, community, space))
+				).entity_id,
+			};
+		}
```
ryanatkn, line 158: This should cause tests to fail after adding `deleteTieService` to the services list as described in another comment because there's no tie to delete between the entities. If it doesn't cause tests to fail there's something else wrong. I think the fix is to add a `random.tie` function that gets called here.

greatbacon: With the addition to the services array the test is passing. The way things are currently written, the test is leveraging a tie created during the `CreateTie` service test.  I'll update it with a random tie function to keep things separated though.

ryanatkn (approved):
saweet

PR #281, opened 2022-03-11
fix schema generation hack

ryanatkn:
- [x] fix the schema generation hack to use `$ref` instead
- [x] upgrade gro to use the new schema resolver and check that it's working


src/lib/app/eventTypes.gen.ts
```diff
@@ -16,17 +17,7 @@ const opts: Partial<JsonSchemaToTypeScriptOptions> = {
...
-					const schema = schemas.find((s) => s.$id === file.url);
-					if (!schema)
-						throw Error(
-							`Unable to find schema: "${file.url}".` +
-								' Is it unregistered in $lib/app/schemas.ts, or a typo, or outdated?',
-						);
-					return JSON.stringify(schema);
-				},
-			},
+			vocab: toVocabSchemaResolver(schemas),
```
ryanatkn, line 20: This is now imported from Gro, and Gro also uses it to resolve `.schema.` file `$ref`s.

src/lib/app/schemas.ts
```diff
@@ -1,12 +1,41 @@
-import type {SchemaObject} from 'ajv';
```
ryanatkn, line 1: I moved the `entities.ts` here -- the name wasn't correct, it was more like vocab, but I see no reason to just put all schemas together.

src/lib/app/schemas.ts
```diff
@@ -1,12 +1,41 @@
...
+import {TieSchema} from '$lib/vocab/tie/tie.schema';
 
-export const schemas: SchemaObject[] = entities;
+// TODO The casts to `as VocabSchema` in this file
+// are needed because the `json-schema` types are very strict,
+// and `as const` fails due to `readonly` properties.
+// It could be fixed by declaring schemas with their type,
+// e.g. `export const AccountSchema: VocabSchema = {`
+// but this has the downside of losing the inferred default types,
+// which are handy when the schemas are used directly.
```
ryanatkn, line 23: There's some unfortunate incompatibility with `json-schema`'s schema type (which we extend `VocabSchema` from) and `ajv`'s `SchemaObject`, so we do some casting in this module (this comment explains more) and also you'll see the union type of the two in a few places. It's a pretty benign issue overall, and it's better than the alternatives I could think of.

greatbacon (approved):

PR #282, opened 2022-03-11
add hue customization to community input

ryanatkn:
Adds hue customization when creating communities, hidden by default behind a `details` toggle. It also now shows a preview of the community icon+name.

Upgrades to our latest eslint config since I hit a problem we fixed.

Also deletes `HueInput.svelte` and uses the Felt version.


src/lib/app/contextmenu/ActingPersonaContextmenu.svelte
```diff
@@ -24,6 +24,7 @@
 				dispatch('OpenDialog', {
 					Component: CommunityInput,
 					props: {persona, done: () => dispatch('CloseDialog')},
+					dialogProps: {layout: 'page'},
```
ryanatkn, line 27: Most dialogs that grow in size should use this setting to prevent weird UX, it makes it flow downwards from the top.

src/lib/ui/CommunityInput.svelte
```diff
@@ -56,17 +70,39 @@
...
+	{/if}
 </div>
+
+<style>
+	details {
+		font-size: var(--font_size_lg);
+	}
+	.customize {
+		margin-top: var(--spacing_lg);
+	}
```
ryanatkn, line 107: This layout tweak is a hack, the "right" answer seems really difficult.

ryanatkn: Got this fixed in Felt when we publish the next version: `summary` elements inside `.markup` have a `margin-bottom` like many other elements now.

PR #284, opened 2022-03-16
Hard delete entities

greatbacon:
This PR adds an event for "hard" deleting entities.

It clarifies the previous delete functionality that creates `Tombstone` entities as a "soft" delete.

It also updates some View templates to properly render `Tombstone`s

### DIFF
--- a/src/lib/server/services.ts
+++ b/src/lib/server/services.ts
@@ -16,7 +16,7 @@ import {
 	readEntitiesService,
 	createEntityService,
 	updateEntityService,
-	deleteEntityService,
+	softDeleteEntityService,
 } from '$lib/vocab/entity/entityServices';
 import {
 	readSpaceService,
@@ -39,7 +39,7 @@ export const services: Map<string, Service<any, any>> = new Map(
 		createSpaceService,
 		createEntityService,
 		updateEntityService,
-		deleteEntityService,
+		softDeleteEntityService,
 		readCommunityService,
 		readCommunitiesService,
 		updateCommunitySettingsService,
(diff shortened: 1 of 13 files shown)


src/lib/server/services.ts
```diff
@@ -39,7 +39,7 @@ export const services: Map<string, Service<any, any>> = new Map(
 		createSpaceService,
 		createEntityService,
 		updateEntityService,
-		deleteEntityService,
+		softDeleteEntityService,
```
ryanatkn, line 42:
```suggestion
		softDeleteEntityService,
		hardDeleteEntityService,
```

src/lib/server/services.ts
```diff
@@ -16,7 +16,7 @@ import {
 	readEntitiesService,
 	createEntityService,
 	updateEntityService,
-	deleteEntityService,
+	softDeleteEntityService,
```
ryanatkn, line 19:
```suggestion
	softDeleteEntityService,
	hardDeleteEntityService,
```

ryanatkn (approved):
Good stuff, only had a small change to register the hard delete version.

I was curious to check if the hard delete failed tests and it looks like it does. Not sure how to read the error message, may be related to cascade.

greatbacon:
> Good stuff, only had a small change to register the hard delete version.
> 
> I was curious to check if the hard delete passes tests after registering in the services, and it looks like it fails. Not sure how to read the error message, may be related to cascade. I'm confused that there's a tie being created for the entity that's being deleted.

Probably has something to do with how parameters are being generated. I'll take a look at it.

PR #285, opened 2022-03-17
upgrade deps including SvelteKit

ryanatkn:
Upgrades to the latest SvelteKit and adapter-node. Rewrites the cookie session handling to use `cookie` and `cookie-signature` instead of `cookie-session`, which was a heavier and now incompatible solution. The`cookie-session` API relies on `cookies` whose API requires both the request and response objects, but the SvelteKit hooks API requires the session to be loaded before the response object is available, making the API incompatible. The result is we use two lower-level cookie modules and add some glue code.

- [x] make the changes needed on all projects: https://github.com/feltcoop/felt-template/pull/3
- [x] fix login and session
- [x] fix `getSession` ambient types and return value
- [x] fix build for adapter-node

future:

- fix `import.meta` usage - see `server/env.ts`


src/hooks.ts
```diff
@@ -1,27 +1,36 @@
...
+
+export const getSession: GetSession = async (event) => {
+	const {account_id} = event.locals;
+	if (!account_id) return {guest: true};
+	const result = await db.repos.session.loadClientSession(account_id);
+	if (!result.ok) {
+		// TODO what's the best UX for this condition? just ask the user to try again?
+		// If needed, could set `event.locals` to have `handle` manage this condition:
+		// event.locals.failedToLoadSession = true;
+		log.error('failed to load session');
```
greatbacon, line 32: Should probably include the `account_id` the session tried to load for here.

src/hooks.ts
```diff
@@ -1,27 +1,36 @@
...
+	}
+	return response;
+};
+
+export const getSession: GetSession = async (event) => {
+	const {account_id} = event.locals;
+	if (!account_id) return {guest: true};
+	const result = await db.repos.session.loadClientSession(account_id);
+	if (!result.ok) {
+		// TODO what's the best UX for this condition? just ask the user to try again?
```
greatbacon, line 29: Should we ask the user to repeat the action or just clear the cookie at this point? I'm not sure what situations would result in a failure to load, but not fail again on refresh.

src/lib/server/ApiServer.ts
```diff
@@ -65,7 +61,7 @@ export class ApiServer {
 
 		// Set up the app and its middleware.
 		this.app
-			.use(bodyParser.json()) // TODO is deprecated, but doesn't let us `import {json}`
+			.use(bodyParser.json())
```
greatbacon, line 64: Is this no longer deprecated!? 😀

greatbacon (approved):
Thanks for doing all this work! 🎉 

Maybe before merging this, we should do a couple of test deployments just to make sure we don't totally mess up deploys.

src/hooks.ts
ryanatkn: The cookie should still be valid in this case, so clearing it won't do any good AFAICT. It should only happen if bad things go wrong, like Postgres failing to respond or corrupt data. I'm thinking a UX like showing a modal with an error message and a single button that's equivalent to refreshing the page is close to ideal and easy to implement -- I'll make a note for a followup PR.

src/lib/server/ApiServer.ts
ryanatkn: It's weird, `body-parser` previously had a deprecated message and I think it's because it's builtin to `express` now, but `body-parser` is used in many other libraries, so the upgraded version just removed the deprecation message it seems.

I lowkey desperately wanted a trusted (because body parsing can be a huge security vector) alternative to `body-parser` because it has 30 dependencies: <https://npm.anvaka.com/#/view/2d/body-parser> but I couldn't find a suitable alternative in a short period of time. 30 dependencies just to parse a JSON field. (we don't currently use any of its other encodings, but we may in the future) At least this is one of the most-depended-upon module chains in the entire ecosystem, mitigating the security concerns and making it more of an aesthetic preference.

PR #291, opened 2022-03-18
improve dispatch typesafety

ryanatkn:
Improves the typesafety of `dispatch` by using a proxy and changing the form to `dispatch.SomeEventName(params)` instead of the old `dispatch('SomeEventName', params)`. Due to TypeScript's behavior, the string parameter version didn't give us the desired behavior -- it didn't "trust" the string event name to then determine the params, so any type error messages were effectively useless.


src/lib/app/dispatch.ts
```diff
@@ -34,22 +34,24 @@ export interface ToDispatchClient {
...
+			);
+			const client = toClient(eventName);
+			return ui.dispatch({
+				eventName,
+				params,
+				dispatch,
+				invoke: client ? (p = params) => client.invoke(eventName, p) : null,
+			});
+		},
+	});
```
ryanatkn, line 54: This is the only change in the PR other than using the new call strategy and the codegen for the new type signatures. [Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) seem like a great solution here. We could also generate code that has a fixed set of functions, but that would bloat the JS payload for an irrelevant performance gain, and it's nicer to pass through any event as before -- the `ui` handler can choose what to do if an event is not registered.

src/lib/app/eventTypes.gen.ts
```diff
@@ -93,12 +93,9 @@ export interface Dispatch {
...
 			str +
-			`
-		(
-			eventName: '${eventInfo.name}',
-			params: ${toParamsName(eventInfo.name)},
-		): ${eventInfo.returns};
-`.trim(),
+			`${eventInfo.name}: (params: ${toParamsName(eventInfo.name)}) => ${
+				eventInfo.returns
+			};`.trim(),
```
ryanatkn, line 98: Minor adjustment to the codegen so:

```ts
(eventName: 'LoginAccount', params: LoginAccountParams): Promise<LoginAccountResponseResult>;
```

becomes:

```ts
LoginAccount: (params: LoginAccountParams) => Promise<LoginAccountResponseResult>;
```

greatbacon (approved):

PR #292, opened 2022-03-18
extract ui mutations

ryanatkn:
Extracts event handlers out of `app.ui` as "mutations", like `$lib/vocab/community/communityMutations.ts`

To continue using `Writable` stores in the mutations, but keeping the rest of the app exposed to only `Readable` stores, I had to add a slightly messy `WritableUi` type that's inferred from the `toUi` return value.

I tried converting `Ui` to a class, but it doesn't seem to improve things much because the constructor properties are needed for initializing properties, so they can't all be declared on the class body.


src/lib/app/mutationTypes.ts
```diff
@@ -0,0 +1,149 @@
+// generated by src/lib/app/mutationTypes.gen.ts
```
ryanatkn, line 1: Because it imports so many types, we may want to put this in `eventTypes.ts`. And maybe rename the module, maybe `appTypes.ts`?

src/lib/ui/ui.ts
```diff
@@ -72,15 +72,28 @@ export interface Ui extends Partial<UiHandlers> {
...
+
+// This ensures that the inferred `WritableUi` is assignable to `Ui`.
+// The latter type is used in components and it exposes its data as `Readable` stores,
+// while the format is used in mutations and exposes `Writable` stores.
+// TODO is there a better way to do this assertion without any runtime code?
+// @see https://github.com/feltcoop/felt-server/pull/292
+// and @see https://github.com/feltcoop/felt-server/pull/292/commits/f24a7377a7328df6071771facaacb6464e10a000
+// for runtime alternatives. (though they could be elided in production code probably)
+type Typecheck<T extends Ui> = T;
+export type Typechecked = Typecheck<WritableUi>;
```
ryanatkn, line 89: This is weird but at least it has no runtime representation. It ensures the return type of `toUi`, `WritableUi`, is assignable to `Ui`.

src/lib/ui/ui.ts
```diff
@@ -32,7 +31,9 @@ export const setUi = (store: Ui): Ui => {
 	return store;
 };
 
-export interface Ui extends Partial<UiHandlers> {
+export interface Ui {
+	session: Writable<ClientSession>;
```
ryanatkn, line 35: Change this.

greatbacon (approved):

PR #297, opened 2022-03-22
hoist dispatching mutations from ui

ryanatkn:
Simplifies the role of `ui.ts` to just store and expose data now that mutations are a separate concept. It now has no knowledge of mutations at all. This makes `dispatch` responsible for looking up and applying mutations rather than pushing that concern downstream to `ui.ts`, so now `dispatch` is more clearly the point of orchestration.

### DIFF
--- a/src/lib/app/dispatch.ts
+++ b/src/lib/app/dispatch.ts
@@ -59,7 +59,7 @@ export const toDispatch = (
 			};
 			const mutation = mutations[eventName];
 			if (!mutation) {
-				log.warn(`ignoring event with no mutation: ${eventName}`, ctx);
+				log.warn('ignoring event with no mutation', ctx);
 				return;
 			}
 			return mutation(ctx);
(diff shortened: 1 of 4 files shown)


src/lib/app/dispatch.ts
```diff
@@ -72,11 +87,17 @@ export const toDispatchBroadcastMessage =
...
 			dispatch,
 			invoke: () => Promise.resolve(message.result),
-		});
+		};
+		const mutation = mutations[eventName];
+		if (!mutation) {
+			log.warn(`ignoring broadcast event with no mutation: ${eventName}`, ctx);
+			return;
+		}
+		return mutation(ctx);
```
ryanatkn, line 102: This logic appears above with a slightly different error message, but it's so simple that I copypasted instead of extracting a helper. A benefit is that it keeps our call stacks shallow. (so far we've done pretty well there)

greatbacon (approved):

PR #299, opened 2022-03-24
sort spaces in space nav

ryanatkn:
- makes the space nav have a stable sort order -- the home space is first, followed by the spaces sorted by `name`
- removes the created timestamp from the workspace header -- the same value is visible on "Edit Community", and we talked about adding a contextmenu entry to view info like this for communities/spaces, "More Info" or something

future:

- let users specify the sort order for spaces

### DIFF
--- a/src/lib/ui/MainNav.svelte
+++ b/src/lib/ui/MainNav.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+	import {get} from 'svelte/store';
+
 	import CommunityNav from '$lib/ui/CommunityNav.svelte';
 	import SpaceNav from '$lib/ui/SpaceNav.svelte';
 	import Avatar from '$lib/ui/Avatar.svelte';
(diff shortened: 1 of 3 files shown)


src/lib/ui/MainNav.svelte
```diff
@@ -27,6 +29,13 @@
 	$: selectedCommunitySpaces =
 		selectedCommunity && $spacesByCommunityId.get($selectedCommunity!.community_id);
 
+	// TODO how to do this without using `get`?
```
greatbacon, line 32: Maybe do it as an insort when updating the store?

greatbacon (approved):

src/lib/ui/MainNav.svelte
ryanatkn: Cool yeah I made that change. It doesn't solve the `get` question but it seems more correct -- the arrays in `spacesByCommunityId` will always be sorted now. I thought about that initially but the performance implications aren't good, but we already had a todo to fix the performance of `spacesByCommunityId` and similar maps, so this seems better and we'll deal with performance later.

PR #300, opened 2022-03-28
Directory refactor - STEP 1

greatbacon:
Alright, so this is STEP 1 of what I think will be a 3 step process to migrate us from directly querying entities via space_id to a graph crawling process facilitated by a directory.

As a recap: currently entities are directly scoped to a single space and are directly queried by that space's id. As part of the entities/tie work, we decided that it would be more useful in the long run to have the system use a `directory` system. In this case, instead of each entity tracking a single space it belongs to, spaces would hold a reference to a Collection entity called `directory` which then serves as an entry point for navigating a graph of entities. This ultimately allows for entities to be cross referenced by different spaces & views across the community.

To get there I have outlined a 3 step process:

STEP 1: (i.e. this PR) set up the system that ensures all spaces have a `directory` and properly reference it
STEP 2: make all new and existing entities `tie` to the `directory` (not the long term plan, but preserves currently functionality)
STEP 3: move to the query by `directory_id` instead of by `space_id`


src/lib/vocab/community/CommunityRepo.ts
```diff
@@ -14,19 +13,15 @@ export class CommunityRepo extends PostgresRepo {
...
 			) RETURNING *
 		`;
 		log.trace('[db] created community', data[0]);
 		const community = data[0];
-		// TODO more robust error handling or condense into single query
-		const spacesResult = await this.db.repos.space.createDefaultSpaces(community); // TODO should this work happen elsewhere?
-		if (!spacesResult.ok) return spacesResult;
-		const spaces = spacesResult.value;
-		return {ok: true, value: {community, spaces}};
+		return {ok: true, value: community};
```
greatbacon, line 24: Part of the work of this PR moves the createDefault logic into it's own function (mostly to take advantage of the business logic around creating directories living at the service layer)

greatbacon:
Also, PR 300! 🎉

src/lib/db/README.md
```diff
@@ -54,6 +54,14 @@ Migration files are located in the [migrations](./migrations) directory.
 To create a new migration file use the `ley new` command (see Ley's docs for more details)
 Use `gro lib/db/migrate` to run migrations.
 
+## creating backups
+
+Using the following command one can quickly take backups of the felt table
```
ryanatkn, line 59:
```suggestion
Using the following command one can quickly take backups of the `felt` database.
```

src/lib/vocab/space/spaceServices.ts
```diff
@@ -21,6 +21,13 @@ import {
 	UpdateSpace,
 	DeleteSpace,
 } from '$lib/vocab/space/spaceEvents';
+import type {Community} from '../community/community';
+import type {Result} from '@feltcoop/felt';
+import type {Space} from './space';
+import type {ErrorResponse} from '$lib/util/error';
+import {toDefaultSpaces} from './defaultSpaces';
+import {db} from '$lib/db/db';
+import {type ISessionApi} from '$lib/server/SessionApi';
```
ryanatkn, line 30:
```suggestion
import type {ISessionApi} from '$lib/server/SessionApi';
```

The SvelteKit upgrade requires we use this syntax when all of the imports to a module are types. (TypeScript will report an error)

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -6,21 +6,31 @@ import {PostgresRepo} from '$lib/db/PostgresRepo';
 import type {Entity} from '$lib/vocab/entity/entity';
 import type {EntityData} from '$lib/vocab/entity/entityData';
 import type {ErrorResponse} from '$lib/util/error';
+import {type RowList} from 'postgres';
```
ryanatkn, line 9:
```suggestion
import type {RowList} from 'postgres';
```

See this comment for why: https://github.com/feltcoop/felt-server/pull/300/files#r840182574

src/lib/vocab/space/defaultSpaces.ts
```diff
@@ -35,23 +35,9 @@ export const toDefaultSpaces = ({community_id, name}: Community): CreateSpacePar
 	},
 	{
 		community_id,
-		name: 'voice',
```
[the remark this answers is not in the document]
ryanatkn: good stuff, I think we should remove the `Voice` stuff in `$lib/ui/view/components.ts` and `$lib/vocab/view/view.ts` too

src/lib/db/seed.ts
```diff
@@ -133,6 +140,37 @@ const entitiesContents: Record<string, string[]> = {
...
+	const list = await db.repos.entity.create(
+		persona_id,
+		{
+			type: 'Collection',
+			name: 'Grocery List',
+		},
+		space_id,
+	);
+	if (!list.ok) {
+		log.warn('Issue generating todo list');
```
ryanatkn, line 153: Instead of warning and returning here and below, it'll be more succinct to use `unwrap`, and I think we'd prefer to have an error thrown if seeding fails.

src/lib/db/seed.ts
```diff
@@ -106,15 +106,22 @@ const createDefaultEntities = async (db: Database, spaces: Space[], personas: Pe
...
 		if (!componentName || !(componentName in entitiesContents)) {
 			continue;
 		}
 		const entityContents = entitiesContents[componentName];
 		for (const entityContent of entityContents) {
-			await db.repos.entity.create(nextPersona().persona_id, space.space_id, {
-				type: 'Note',
-				content: entityContent,
-			});
+			await db.repos.entity.create(
```
ryanatkn, line 117: Realized this will be slightly more robust if we wrap all service/repo calls in `unwrap` in this module, even if we don't use the return value.

ryanatkn, line 117: Realized this will be slightly more robust if we wrap all service/repo calls in `unwrap` in this module, even if we don't use the return value.

src/lib/db/seed.ts
```diff
@@ -133,6 +140,37 @@ const entitiesContents: Record<string, string[]> = {
...
+				type: 'Note',
+				content: item,
+			},
+			space_id,
+		);
+		if (!result.ok) {
+			log.warn('Issue generating todo item');
+			return;
+		}
+		await db.repos.tie.create(list.value.entity_id, result.value.entity_id, 'HasItem');
```
ryanatkn, line 170:
```suggestion
		unwrap(await db.repos.tie.create(list.value.entity_id, result.value.entity_id, 'HasItem'));
```

reason given here: https://github.com/feltcoop/felt-server/pull/300/files#r840193927

src/lib/db/seed.ts
```diff
@@ -106,15 +106,22 @@ const createDefaultEntities = async (db: Database, spaces: Space[], personas: Pe
...
-				content: entityContent,
-			});
+			await db.repos.entity.create(
+				nextPersona().persona_id,
+				{
+					type: 'Note',
+					content: entityContent,
+				},
+				space.space_id,
+			);
```
ryanatkn, line 124:
```suggestion
			unwrap(
				await db.repos.entity.create(
					nextPersona().persona_id,
					{
						type: 'Note',
						content: entityContent,
					},
					space.space_id,
				),
			);
```

Realized this will be slightly more robust if we wrap all seed service/db calls in `unwrap`, even if they don't use the return value.

src/lib/db/seed.ts
ryanatkn, line 170: Had some overlapping suggestions here so it's coming as a big block.

- Instead of warning and returning early in 2 places, it's more succinct to use `unwrap`, and I think we'd prefer to have its behavior of throwing an error if any part of seeding fails.
- Use `unwrap` on `db.repos.tie.create` even though the return value isn't used so any errors will make seeding fail.

```suggestion
	const list = unwrap(
		await db.repos.entity.create(
			persona_id,
			{
				type: 'Collection',
				name: 'Grocery List',
			},
			space_id,
		),
	);
	const itemsContent = ['eggs', 'milk', 'bread'];
	for (const content of itemsContent) {
		const item = unwrap(
			await db.repos.entity.create(
				persona_id,
				{
					type: 'Note',
					content,
				},
				space_id,
			),
		);
		unwrap(await db.repos.tie.create(list.entity_id, item.entity_id, 'HasItem'));
```

src/lib/vocab/community/CommunityRepo.ts
ryanatkn: excellent, this was business logic that didn't belong in repos

src/lib/vocab/random.ts
```diff
@@ -21,6 +21,7 @@ import type {Tie} from '$lib/vocab/tie/tie';
 import {createAccountPersonaService} from '$lib/vocab/persona/personaServices';
 import {SessionApi} from '$lib/server/SessionApi';
 import {createCommunityService} from '$lib/vocab/community/communityServices';
+import {createSpaceService} from './space/spaceServices';
```
ryanatkn, line 24:
```suggestion
import {createSpaceService} from '$lib/vocab/space/spaceServices';
```

eventually vscode will get this right more often...

src/lib/vocab/space/spaceServices.ts
```diff
@@ -21,6 +21,13 @@ import {
 	UpdateSpace,
 	DeleteSpace,
 } from '$lib/vocab/space/spaceEvents';
+import type {Community} from '../community/community';
+import type {Result} from '@feltcoop/felt';
+import type {Space} from './space';
+import type {ErrorResponse} from '$lib/util/error';
+import {toDefaultSpaces} from './defaultSpaces';
```
ryanatkn, line 28:
```suggestion
import type {Community} from '$lib/vocab/community/community';
import type {Result} from '@feltcoop/felt';
import type {Space} from '$lib/vocab/space/space';
import type {ErrorResponse} from '$lib/util/error';
import {toDefaultSpaces} from '$lib/vocab/space/defaultSpaces';
```

vscode!!!! D:

ryanatkn (approved):

PR #313, opened 2022-04-03
move repo error messages to services

ryanatkn:
- moves error messages from repos to services
- Deletes error types from repos when there's only one type. Currently this is all cases, because we simplified repos to have no business logic -- I thought about this more, and these are currently useless and complicated the code (lots of net LOC deleted, many lines simplified), and we only added error types to about half of the repo methods. In the future, we may need to disambiguate multiple kinds of errors for a single repo call, and we can use this pattern for just those cases. If Postgres or the driver encounter an error, they'll throw, not return a `Result`, and our http/websocket middleware catch the error and return a 500.
- standardizes the exit-early-on-error pattern in the repos/services


src/lib/app/dispatch.ts
```diff
@@ -5,7 +5,7 @@ import type {WritableUi} from '$lib/ui/ui';
 import type {ApiClient} from '$lib/ui/ApiClient';
 import type {ApiResult} from '$lib/server/api';
 import type {Dispatch} from '$lib/app/eventTypes';
-import type {BroadcastMessage} from '$lib/server/websocketMiddleware';
+import type {BroadcastMessage} from '$lib/server/websocketServiceMiddleware';
```
ryanatkn, line 8: Renamed this module to match `httpServiceMiddleware` -- the original reasoning was that it's the only websocket middleware, and it'll likely do other things in the future, but for clarity I think it's better to name it the same as the http version, and when we add additional websocket middleware, we can make a new generic module with the name `websocketMiddleware`, and this one can stay the same.

src/lib/vocab/account/AccountRepo.ts
```diff
@@ -24,33 +23,26 @@ export class AccountRepo extends PostgresRepo {
...
-			return {ok: true, value: data[0]};
+		if (!data.length) {
+			return {ok: false, type: 'no_account_found'};
 		}
-		return {
-			ok: false,
-			type: 'no_account_found',
-			message: 'no account found',
-		};
+		return {ok: true, value: data[0]};
```
ryanatkn, line 35: Changed 7 of these to exit early on errors so they read consistently.

src/lib/vocab/community/CommunityRepo.ts
```diff
@@ -66,32 +61,23 @@ export class CommunityRepo extends PostgresRepo {
...
 
-	async deleteById(
-		community_id: number,
-	): Promise<Result<object, {type: 'deletion_error'} & ErrorResponse>> {
+	async deleteById(community_id: number): Promise<Result<object, {type: 'deletion_error'}>> {
 		log.trace('[deleteById]', community_id);
 		const data = await this.db.sql<any[]>`
 			DELETE FROM communities WHERE community_id=${community_id}
 		`;
-		if (data.count !== 1) {
```
ryanatkn, line 89: I changed the several `data.count !== 1` checks to `!data.count` because the condition where it deletes/updates more than one isn't exactly an error from the client's POV -- it should proceed as if the deletions happened, because they did, e.g. to update the local caches. If we need special handling for those cases we can address them with a separate error type returned from the repo.

greatbacon (approved):

PR #318, opened 2022-04-04
check for duplicate community names

ryanatkn:
Adds a check for duplicate community names and adds a test.

### DIFF
--- a/src/lib/vocab/community/communityServices.ts
+++ b/src/lib/vocab/community/communityServices.ts
@@ -68,6 +68,16 @@ export const createCommunityService: Service<CreateCommunityParams, CreateCommun
 		perform: async (serviceRequest) => {
 			const {repos, params, account_id} = serviceRequest;
 			log.trace('creating community account_id', account_id);
+
+			// Check for duplicate community names.
+			const findCommunityResult = await repos.community.findByName(params.name);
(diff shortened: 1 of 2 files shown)


src/lib/vocab/community/communityServices.ts
```diff
@@ -68,6 +68,16 @@ export const createCommunityService: Service<CreateCommunityParams, CreateCommun
 		perform: async (serviceRequest) => {
 			const {repos, params, account_id} = serviceRequest;
 			log.trace('creating community account_id', account_id);
+
+			// Check for duplicate community names.
+			const findCommunityResult = await repos.community.findByName(params.name);
+			if (!findCommunityResult.ok) {
+				return {ok: false, status: 500, message: 'failed to lookup existing community by name'};
```
ryanatkn, line 75: I'm wondering if we should use `unwrap` here instead of this check to pass through a generic 500 error to the client. The message doesn't seem very useful for end users, and `unwrap` will log an error for us with the stack trace. We currently use `unwrap` nowhere in actual code, only in tests.

greatbacon: That makes sense to me.

greatbacon (approved):

PR #319, opened 2022-04-04
use services in random

ryanatkn:
Changes the random vocab to use `createEntityService` and `createTieService` instead of the repos. Also extracts some logic from the service integration test to a new `entityServices.test.ts`, and simplifies the service integration tests to use `random` instead of duplicating the same logic -- random previously used repos, but now that it uses services there's no reason for the integration test to be more complicated.

### DIFF
--- a/src/lib/vocab/entity/entityServices.test.ts
+++ b/src/lib/vocab/entity/entityServices.test.ts
@@ -1,0 +1,34 @@
+import {suite} from 'uvu';
+import * as assert from 'uvu/assert';
+
+import {setupDb, teardownDb, type TestDbContext} from '$lib/util/testDbHelpers';
+import type {TestAppContext} from '$lib/util/testAppHelpers';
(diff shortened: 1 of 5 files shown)


src/lib/vocab/entity/entityServices.test.ts
```diff
@@ -0,0 +1,34 @@
...
+	assert.is(entity1.actor_id, persona.persona_id);
+	assert.is(entity2.actor_id, persona.persona_id);
+	assert.is(entity1.space_id, space.space_id);
+	assert.is(entity2.space_id, space.space_id);
+	assert.equal(entity1.data, entityData1);
+	assert.equal(entity2.data, entityData2);
+});
+
+test_entityServices.run();
+/* test_entityServices */
```
ryanatkn, line 34: this logic is all copypasted from `servicesIntegration.test.ts`

greatbacon (approved):

PR #322, opened 2022-04-05
Directory part 2

greatbacon:
This PR handles Part 2 of the directory architectural refactor.

It introduces a migration query that connects all orphan entities (i.e. all those without a tie where they are the dest, i.e. any top level or non connected entities) and ties them to their space's Directory (introduced in  #300 )

It also expands the CreateEntity service params to include a `source_id` attribute & option `type` (which defaults to HasItem) to make sure no Entity is created as an orphan from here out.

Lastly, it updates all uses of CreateEntity on the front end to now include the directory_id in the call (except in Todo, where we get to condense some existing Tie logic down to a single call! 🎉 )

### DIFF
--- a/src/lib/ui/view/Todo.svelte
+++ b/src/lib/ui/view/Todo.svelte
@@ -68,19 +68,12 @@
 		if (!content || !selectedList) return;
 
 		//TODO better error handling
-		//TODO create api call to do this in one step?
-		const entityResult = await dispatch.CreateEntity({
+		await dispatch.CreateEntity({
 			space_id: $space.space_id,
 			data: {type: 'Note', content, checked: false},
 			actor_id: $persona.persona_id,
+			source_id: selectedList.entity_id,
 		});
-		if (entityResult.ok) {
-			await dispatch.CreateTie({
-				source_id: selectedList.entity_id,
-				dest_id: entityResult.value.entity.entity_id,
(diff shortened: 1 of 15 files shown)


src/lib/ui/view/Todo.svelte
```diff
@@ -68,19 +68,12 @@
...
 			actor_id: $persona.persona_id,
+			source_id: selectedList.entity_id,
 		});
-		if (entityResult.ok) {
-			await dispatch.CreateTie({
-				source_id: selectedList.entity_id,
-				dest_id: entityResult.value.entity.entity_id,
-				type: 'HasItem',
-			});
-		}
```
ryanatkn, line 83: # :tada:

ryanatkn (approved):
Looks fantastic! I'm guessing the answer is yes, wanted to ask if the SQL migration works with production data to leave no orphaned entities.

greatbacon:
> Looks fantastic! I'm guessing the answer is yes, wanted to ask if the SQL migration works with production data to leave no orphaned entities.

It should, but I haven't formally tested it. I'll do a test run locally before merging.

greatbacon:
I tested things locally and it looks like it's creating ties properly (such that everything can be queried via the recursive statement). I'm gonna go ahead and merge and deploy this!

PR #323, opened 2022-04-06
fix tie repo test in directory-part-2 branch

ryanatkn:
Fixes the test assertions in #322 . The problem is `includes` compares by reference not value. This uses `.find` and the Node builtin `isDeepStrictEqual`. Also changed the second count assertion from 7 to 3 and used `unwrap` for concision.


greatbacon (approved):
Cool, thanks!

PR #324, opened 2022-04-11
fix db init error

ryanatkn:
Still investigating the root issue.

### DIFF
--- a/src/lib/db/README.md
+++ b/src/lib/db/README.md
@@ -11,18 +11,13 @@
 The Felt server has a dependency on [PostgreSQL](https://www.postgresql.org) 13.2+.
 You can find [setup instructions](https://www.postgresql.org/download/) on their website.
 
-For Ubuntu you can usually just get away with
-
-```bash
-sudo apt-get -y install postgresql
-```
-
 Felt's database driver is [`postgres`](https://github.com/porsager/postgres).
 See its [README](https://github.com/porsager/postgres#readme) to learn more.
 
 At the moment, the server [defaults to connecting](./postgres.ts)
 to the database with the following values,
-prioritizing environment variables if they're defined:
+prioritizing [environment variables](https://vitejs.dev/guide/env-and-mode.html#env-variables)
+if they're defined:
 
 ```
 database = PGDATABASE or 'felt'
--- a/src/lib/server/services.test.ts
+++ b/src/lib/server/services.test.ts
@@ -16,8 +16,8 @@ const test__services = suite<TestDbContext>('services');
 test__services.before(setupDb);
 test__services.after(teardownDb);
 
-test__services('perform services', async ({db, random}) => {
-	for (const service of services.values()) {
+for (const service of services.values()) {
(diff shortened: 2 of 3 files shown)


src/lib/db/README.md
```diff
@@ -11,18 +11,13 @@
 The Felt server has a dependency on [PostgreSQL](https://www.postgresql.org) 13.2+.
 You can find [setup instructions](https://www.postgresql.org/download/) on their website.
 
-For Ubuntu you can usually just get away with
-
-```bash
-sudo apt-get -y install postgresql
-```
-
```
ryanatkn, line 19: Removed this because the apt registry is out of date (worse now that Ubuntu moved to Snap), the official Postgres instructions include a custom PPA.

src/lib/server/services.test.ts
```diff
@@ -16,8 +16,8 @@ const test__services = suite<TestDbContext>('services');
 test__services.before(setupDb);
 test__services.after(teardownDb);
 
-test__services('perform services', async ({db, random}) => {
-	for (const service of services.values()) {
+for (const service of services.values()) {
+	test__services(`perform service ${service.event.name}`, async ({db, random}) => {
```
ryanatkn, line 20: Some quirk of uvu that I don't understand: this causes the tests to fail with an error rather than quietly pass with actual failures.

src/lib/app/events.test.ts
```diff
@@ -19,8 +19,8 @@ test__eventInfos.after(teardownDb);
 test__eventInfos.before(setupApp(noop as any)); // TODO either use `node-fetch` or mock
 test__eventInfos.after(teardownApp);
 
-test__eventInfos('dispatch random events in a client app', async ({app, random}) => {
-	for (const eventInfo of eventInfos.values()) {
+for (const eventInfo of eventInfos.values()) {
+	test__eventInfos(`dispatch event ${eventInfo.name} in a client app`, async ({app, random}) => {
```
ryanatkn, line 23: see https://github.com/feltcoop/felt-server/pull/324/files#r847587428

src/lib/server/servicesIntegration.test.ts
```diff
@@ -76,8 +76,7 @@ test_servicesIntegration('services integration test', async ({db, random}) => {
 	const {entities: filterFilesValue} = unwrap(
 		await readEntitiesService.perform({params: {space_id: space.space_id}, ...serviceRequest}),
 	);
-	assert.is(filterFilesValue.length, 2);
-	assert.equal(filterFilesValue, [entity1, entity2]);
+	assert.equal(filterFilesValue.slice(), [entity1, entity2]); // `slice` because `RowList` is not deep equal to arrays
```
ryanatkn, line 79: The `RowList` returned by the updated driver is no longer deeply equal to arrays. It's not because of enumerable properties, and uvu's reporting isn't displaying the problem, but I think it's inconsequential -- I think the best fix is to just call `.slice()` sometimes in tests. I don't think it's worth the performance hit to do this in every query that can return a `RowList`.

greatbacon (approved):

PR #325, opened 2022-04-12
streamline deploy

ryanatkn:
The first commit is everything from our pair programming session. It then restructures `infra/setup.task.ts` with nearly identical behavior:

- changes the `setup` script to be one large `ssh` Bash script with logged checkpoints; it adds the `-e` flag, so any nonzero exit condition will cause the script to bail at the point of failure
  - it's easy to print out the final script at once for e.g. debugging (see the `--dry` flag below)
  - it's simpler to treat it as an atomic unit because we don't care about recovering from an error (all setup task runs are inherently "clean")
  - we don't need to handle `spawn` results multiple times or make multiple SSH logins
- creates the nginx config in memory and embeds it in the script instead of a standalone `scp`
- adds Postgres install script (and a TODO for database initialization, see #327)
- adds `apt upgrade -y` which slows things down but ensures we're on the latest security patches
- adds `-y` to all of the apt lines for full automation
- writes a state file to the remote server home directory to avoid running setup twice on the same machine (we may want to add an override flag for e.g. debugging)
- add `--dry` flag which logs out the script instead of executing it (maybe allow outputting to a file or something? not sure the ideal interface with stdout and the task logging)


src/infra/setup.task.ts
```diff
@@ -1,58 +1,138 @@
...
+				logSequence('Installing Postgres...') +
+				`sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list;'
+				curl -L https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -;
+				sudo apt update;
+				sudo apt install -y postgresql;` +
+				//
+				//
+				// All done! Write a "state file" to avoid running the setup script twice.
+				logSequence(`Success! Writing setup state file to ${FELT_SETUP_STATE_FILE_PATH}`) +
+				`touch ${FELT_SETUP_STATE_FILE_PATH};`,
```
ryanatkn, line 118: Thinking about the script's atomicity, what do you think of writing this file at the very top of the script?

If desired for debugging reasons, we could add a `--debug` flag or something that would **not** exit early if it sees the existing state file. Or `--ignoreStageFile`

ryanatkn: I made this change here to write the state file immediately, preventing possible errors: https://github.com/feltcoop/felt-server/pull/325/commits/d82de2ed9b00b883231012c8855a5fea957538ed

It'd be easy to add an override flag.

src/infra/setup.task.ts
```diff
@@ -1,58 +1,138 @@
...
+			API_SERVER_HOST_PROD,
+			SVELTEKIT_SERVER_HOST,
+		);
+
+		// This file is used to detect if the setup script has already run.
+		const FELT_SETUP_STATE_FILE_PATH = '~/felt_state_setup';
+
 		const deployLogin = `${DEPLOY_USER}@${DEPLOY_IP}`;
+
+		const logSequence = toLogSequence();
```
ryanatkn, line 37: wondering if there's a better name for this thing

ryanatkn, line 37: wondering if there's a better name for this thing

greatbacon: Seems fine to me.

greatbacon (approved):

PR #327, opened 2022-04-12
automate db init

ryanatkn:
initializes the database on server setup

I wasn't sure exactly how to set things up with a custom db user with proper db permissions, so I left a TODO there and just create the db as `postgres`


src/infra/serverDeploy.task.ts
```diff
@@ -54,10 +54,9 @@ export const task: Task = {
 		await spawn('scp', [`src/infra/.env.default`, `${deployLogin}:${currentDeploy}/.env`]);
 		await spawn('scp', [ENV_PROD, `${deployLogin}:${currentDeploy}/${ENV_PROD}`]);
 		//TODO: re/start the server via pm2
-		// export NODE_ENV=production && pm2 start npm -- run start --prefix current_felt_server_deploy
+		// pm2 start npm -- run start --prefix ${currentDeploy}
```
ryanatkn, line 57: added `NODE_ENV=production` in the script to both `.profile` and `.bashrc` -- I think this is the best default, but there could be issues to sort out when running Gro scripts operationally

src/infra/setup.task.ts
```diff
@@ -80,27 +88,27 @@ export const task: Task<SetupTaskArgs> = {
...
+				ln -s ${REMOTE_NGINX_CONFIG_PATH} ${REMOTE_NGINX_SYMLINK_PATH};
+				systemctl start nginx;`,
 			//
 			//
-			//Make sure your DNS records are set up and configured first
-			//TODO stuff is still a little unstable arount this
+			// Install certbot for HTTPS:
 			logSequence('Enabling HTTPS with cerbot and nginx...') +
-				`ln -s ${REMOTE_NGINX_CONFIG_PATH} ${REMOTE_NGINX_SYMLINK_PATH};
+				`apt install -y certbot python3-certbot-nginx;
```
ryanatkn, line 111: I separated nginx and certbot here into two distinct sections so we can easily disable the latter for testing.

greatbacon (approved):

PR #329, opened 2022-04-12
Directory PHASE 3

greatbacon:
This PR tries to touch as little of the interface as possible, while replacing the query for Entities with the new directory crawl query.

Some future work off of this:
Move Entity stores away from indexing off of space_id
Modify event to expect both entities & ties to return

### DIFF
--- a/src/lib/vocab/entity/EntityRepo.ts
+++ b/src/lib/vocab/entity/EntityRepo.ts
@@ -46,6 +46,19 @@ export class EntityRepo extends PostgresRepo {
 		return {ok: true, value: entities};
 	}
 
+	// TODO maybe `EntityQuery`?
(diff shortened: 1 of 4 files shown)


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -46,6 +46,19 @@ export class EntityRepo extends PostgresRepo {
 		return {ok: true, value: entities};
 	}
 
+	// TODO maybe `EntityQuery`?
```
ryanatkn, line 49: need some more time to process this, more ideas: `readEntities` or `queryEntities` or `read`/`query`

ryanatkn (approved):
looks great, seeing the directory do the thing is neat

PR #330, opened 2022-04-13
fix date values in schemas

ryanatkn:
- "deserializes" JSON values returned by the server, so dates become `Date` on the client not strings
  - works generically by property name (because e.g. the `created` property is always the same type across all vocab terms)
  - mutates objects in place for efficiency (if we ever need the original JSON API response, we can just clone first, but I doubt this is ever needed outside of devtool/debug purposes)
  - include tests for both builtin and custom properties
- improves some types around the websocket API
- Install `ajv-keywords` and uninstall `ajv-formats` and use the former's `instanceof: 'Date'` instead of the latter's `format: 'date-time'` -- the latter is valid only with values of type `string | number` (it's for "formats of numbers and strings"), hence the warning messages we saw. Searching the web yields `date-time` as a suggestion for `Date` validation, but that appears to be invalid and/or deprecated and the internal logic to check the format did unnecessary `Date` coercion, so the code is faster now and the warning is gone. When we need to check object types in the future, we'll know `ajv-formats` is the wrong tool.


src/lib/vocab/account/account.schema.ts
```diff
@@ -5,8 +5,8 @@ export const AccountSchema = {
 		account_id: {type: 'number'},
 		name: {type: 'string'},
 		password: {type: 'string'},
-		created: {type: 'object', format: 'date-time', tsType: 'Date'},
-		updated: {type: ['object', 'null'], format: 'date-time', tsType: 'Date | null'},
+		created: {type: 'object', instanceof: 'Date', tsType: 'Date'},
+		updated: {anyOf: [{type: 'object', instanceof: 'Date'}, {type: 'null'}], tsType: 'Date | null'},
```
ryanatkn, line 9: the `instanceof` keyword does not work with the shorthand `type: ['object', 'null']`, hence the change to `anyOf`

src/lib/vocab/account/account.schema.ts
```diff
@@ -5,8 +5,8 @@ export const AccountSchema = {
 		account_id: {type: 'number'},
 		name: {type: 'string'},
 		password: {type: 'string'},
-		created: {type: 'object', format: 'date-time', tsType: 'Date'},
-		updated: {type: ['object', 'null'], format: 'date-time', tsType: 'Date | null'},
+		created: {type: 'object', instanceof: 'Date', tsType: 'Date'},
+		updated: {anyOf: [{type: 'object', instanceof: 'Date', tsType: 'Date'}, {type: 'null'}]},
```
ryanatkn, line 9: the `instanceof` keyword does not work with the shorthand `type: ['object', 'null']`, hence the change to `anyOf`

greatbacon (approved):

PR #331, opened 2022-04-14
replace plaintext content containing links and mentions with components

ryanatkn:
Adds initial hacky support for enhanced rich text, supporting links as a first pass. Hacky because the next version of MDsveX (Pfm) will be the base for this, and we'll write plugins instead of parsing and transforming the plaintext ourselves.

Adds the dependency [`estree-walker`](https://github.com/Rich-Harris/estree-walker) from a familiar author to replace SVAST text nodes with our custom transformed ones, which then get rendered by the existing `SvastView`.

This code can be tested by going to edit a space and adding a link. It's currently bugged though: it adds multiple nested `Link` components because of a quirk of how we're storing the parsed AST as `space.view` -- we should probably hold off on deploying the code until this followup PR to change `space.view` to the text format: https://github.com/feltcoop/felt-server/pull/337

future

- @ mentions
- convert `space.view` to store text not JSON (or if implemented as custom elements, can we just use HTML?)
- use this for regular entity content
- sanitize using an allowlist for elements/attributes
- community-scoped absolute URLs
- use Pfm to do this properly (as a plugin?)


src/lib/app/components.ts
```diff
@@ -1,9 +1,11 @@
...
 import {viewComponents} from '$lib/ui/view/components';
+import Link from '$lib/ui/Link.svelte';
 
 // The collection of components that can be dynamically mounted by the app.
 
 // TODO refactor this to load these components on demand,
 // instead of preloading the entire component library
 
-export const components: {[key: string]: typeof SvelteComponent} = viewComponents;
+export const components: {[key: string]: typeof SvelteComponent} = {...viewComponents, Link};
```
ryanatkn, line 11: When we add rich text behavior to entity content, we'll probably want to separate `spaceComponents` and `contentComponents` or something.

src/lib/ui/Link.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
+	export let href: string;
+</script>
+
+<a rel="noreferrer" {...$$restProps} {href}><slot /></a>
```
ryanatkn, line 5: This doesn't yet treat absolute root links like `/foo` as scoped to a community, it'll be easy to do that using the view context.

greatbacon (approved):

PR #334, opened 2022-04-18
add entityById map

ryanatkn:
- adds `entityById`, a step towards managing entity/tie data on the client in our central stores
- makes `eraseEntity` plural to match `deleteEntities`
- keeps `entityById` and `entitiesBySpace` in sync in the entity mutations
  - adds return value for `EraseEntities` service
- renders tombstone content with a unified component
- adds `"Delete Entity"` to the contextmenu
- handle erased entities in `EntityEditor`, just giving the user a delete button


src/lib/ui/ForumItem.svelte
```diff
@@ -28,11 +29,8 @@
 	]}
 >
 	<div class="markup formatted">
-		{#if $entity.data.type !== 'Tombstone'}
-			{$entity.data.content}
-		{:else}
-			<i>This message was deleted</i>
-		{/if}
+		{#if $entity.data.type === 'Tombstone'}<TombstoneContent {entity} />{:else}{$entity.data
+				.content}{/if}
```
ryanatkn, line 33: formatting is weird to avoid that sneaky whitespace

src/lib/ui/BoardItem.svelte
```diff
@@ -29,11 +30,8 @@
 >
 	<div class="markup formatted">
 		<p>
-			{#if $entity.data.type !== 'Tombstone'}
-				{$entity.data.content}
-			{:else}
-				<i>This message was deleted</i>
-			{/if}
+			{#if $entity.data.type === 'Tombstone'}<TombstoneContent {entity} />{:else}{$entity.data
+					.content}{/if}
```
ryanatkn, line 34: formatting is squished to avoid that sneaky whitespace which causes various effects on layout

src/lib/vocab/entity/entityMutations.ts
```diff
@@ -1,66 +1,52 @@
...
-	const existingEntities = entitiesBySpace.get(space_id);
-	// TODO probably check to make sure they don't already exist
-	const newEntities = result ? result.value.entities.map((f) => writable(f)) : [];
-	if (existingEntities) {
-		existingEntities.set(newEntities);
-	} else {
-		entitiesBySpace.set(space_id, writable(newEntities));
+	//TODO update ties once stores are in place: `result.value.ties`
+	for (const $entity of result.value.entities) {
+		updateEntity(ui, $entity);
```
ryanatkn, line 49: It may be a better pattern to change the helper to `updateEntities` so it can efficiently batch updating collections like the entities per space -- as it is, it's doing a ton of unnecessary work with the line `existingSpaceEntities.update(($entities) => $entities.concat(entity!));` getting called per entity.

We'll probably change the data structure of `existingSpaceEntities` to a `mutable`, maybe a `Set`, which would be a big improvement, but even then it'd still trigger N updates to all stores downstream.

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -70,21 +70,26 @@ export class EntityRepo extends PostgresRepo {
 		return {ok: true, value: result[0]};
 	}
 
+	// TODO need to either remove `AND data->>'type' != 'Tombstone'`
+	// from `eraseById` or add it to `updateEntityData`,
+	// because users can currently edit an erased's entity data but then not erase it again
```
ryanatkn, line 75: flagging this to talk about

greatbacon (approved):

PR #340, opened 2022-04-20
Add delete modal friction

greatbacon:
This PR adds some friction to the existing SpaceDelete modal, as well as adding one for CommunityDelete too.

Since Entities are a little more lightweight, I didn't add one there. I did rename the SoftDelete endpoint to match our Delete::Erase nomenclature though.

1 note: I'm not 100% sure on the input check logic in the modals. It doesn't seem to immediately parse the name, so I may have chosen the wrong event?

### DIFF
--- a/src/lib/ui/CommunityDelete.svelte
+++ b/src/lib/ui/CommunityDelete.svelte
@@ -7,6 +7,7 @@
 	import CommunityAvatar from '$lib/ui/CommunityAvatar.svelte';
 	import type {Community} from '$lib/vocab/community/community';
 	import type {Persona} from '$lib/vocab/persona/persona';
+	import PendingButton from '@feltcoop/felt/ui/PendingButton.svelte';
 
 	const {dispatch} = getApp();
 
@@ -15,10 +16,12 @@
 	export let done: (() => void) | undefined = undefined;
 
 	let errorMessage: string | undefined;
+	let pending = false;
 	let lockText = '';
 	$: locked = lockText.toLowerCase().trim() !== $community.name.toLowerCase();
 
 	const deleteCommunity = async () => {
+		pending = true;
 		errorMessage = '';
 		const result = await dispatch.DeleteCommunity({
 			community_id: $community.community_id,
@@ -28,6 +31,7 @@
 		} else {
 			errorMessage = result.message;
 		}
+		pending = false;
 	};
 
 	const onKeydown = async (e: KeyboardEvent) => {
@@ -59,8 +63,8 @@
 			placeholder=">enter name to unlock button"
 			bind:value={lockText}
 		/>
-		<button disabled={locked} type="button" on:click={deleteCommunity} on:keydown={onKeydown}>
+		<PendingButton {pending} disabled={locked || pending} on:click={deleteCommunity}>
 			Delete space
-		</button>
+		</PendingButton>>
 	</form>
 </div>
--- a/src/lib/ui/CommunityDelete.svelte
+++ b/src/lib/ui/CommunityDelete.svelte
@@ -33,13 +33,6 @@
 		}
 		pending = false;
 	};
-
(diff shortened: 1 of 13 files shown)


src/lib/ui/SpaceDelete.svelte
```diff
@@ -56,6 +62,16 @@
 		{#if errorMessage}
 			<Message status="error">{errorMessage}</Message>
 		{/if}
-		<button type="button" on:click={deleteSpace} on:keydown={onKeydown}> Delete space </button>
+		<input
+			type="text"
+			id="name"
+			name="name"
+			placeholder=">enter name to unlock button"
+			on:input={checkLocked}
```
ryanatkn, line 70: The problem seems to be the order that `bind:value` and `on:input` execute, the source of truth gets out of wack. Looks like a reactive statement works well, here's the changes to make both here and in `CommunityDelete.svelte`:

- delete `on:input={checkLocked}`
- delete `checkLocked`
- change `let locked = true;` to `$: locked = lockText.toLowerCase().trim() !== $space.name.toLowerCase();` (trim because copypasted whitespace gets me often on GitHub, lowercase just as a rare convenience)

src/lib/ui/SpaceDelete.svelte
```diff
@@ -56,6 +62,16 @@
...
+			bind:value={lockText}
+		/>
+		<button disabled={locked} type="button" on:click={deleteSpace} on:keydown={onKeydown}>
```
ryanatkn, line 73: If you want to do this now -- and no worries if not, there's other places that probably need it too -- could change this to use a `PendingButton` pretty easily, just need to track a variable before/after the async call. Looks like we have a few subtly different patterns scattered around for this, getting close to justifying extracting common components.

src/lib/ui/CommunityDelete.svelte
```diff
@@ -0,0 +1,71 @@
...
+		});
+		if (result.ok) {
+			done?.();
+		} else {
+			errorMessage = result.message;
+		}
+	};
+
+	const onKeydown = async (e: KeyboardEvent) => {
+		if (e.key === 'Enter') {
```
ryanatkn, line 34:
```suggestion
		if (!locked && e.key === 'Enter') {
```

same change is needed in `SpaceDelete.svelte`

ryanatkn (approved):
most excellent, found the input issue and some minor refinements

src/lib/ui/SpaceDelete.svelte
greatbacon: So I tested this out, and I'm not sure it 100% fits from a UX perspective. The pending `...` implies an action having happened & a result that will be returned, but that seems kind of jarring when you first load the modal since you just want the button greyed out. Maybe some way to distinguished between a `locked` state & a `pending` state?

src/lib/ui/CommunityDelete.svelte
greatbacon: great catch!

src/lib/ui/SpaceDelete.svelte
ryanatkn: I don't mean to change the `locked` behavior, I think that's great, I was thinking show the `...` animation only while the delete call to the server is in flight

src/lib/ui/CommunityDelete.svelte
```diff
@@ -33,13 +33,6 @@
 		}
 		pending = false;
 	};
-
-	const onKeydown = async (e: KeyboardEvent) => {
-		if (!locked && e.key === 'Enter') {
-			e.preventDefault();
-			await deleteCommunity();
-		}
-	};
```
ryanatkn: Could add `on:keydown` to the input to make this work -- the handler was previously on the button but that's unnecessary, browser already does that.

PR #342, opened 2022-04-26
refactor communityIdSelectionByPersonaId to a mutable map and keep it synced

ryanatkn:
Changes `communityIdSelectionByPersonaId` to use a `mutable` `Map` and properly keeps it in sync. Using a map makes the code a bit simpler and makes the key properly a `number` (the `persona_id`), where the previous strategy of using a plain object coerced them to strings, so iterating over the object's keys required casting.

also does some misc cleanup


src/lib/ui/ui.ts
```diff
@@ -290,32 +287,21 @@ export const toUi = (
...
-										(m) => m.persona_id === $persona.persona_id,
-									);
-									const $firstCommunity = $session.communities.find(
-										(c) => c.community_id === $firstMembership?.community_id,
-									)!;
-									return [$persona.persona_id, $firstCommunity.community_id];
-								})
-								.filter(Boolean),
+							// TODO first try to load this from localStorage
+							$sessionPersonas.map(($persona) => [$persona.persona_id, $persona.community_id]),
```
ryanatkn, line 291: this is a simplification made possible by the addition of `persona.community_id`

src/lib/ui/ui.ts
```diff
@@ -350,4 +336,5 @@ const toInitialPersonas = (session: ClientSession): Persona[] =>
 // and @see https://github.com/feltcoop/felt-server/pull/292/commits/f24a7377a7328df6071771facaacb6464e10a000
 // for runtime alternatives. (though they could be elided in production code probably)
 type Typecheck<T extends Ui> = T;
-export type Typechecked = Typecheck<WritableUi>;
+// @ts-expect-error because it's unused
+type Typechecked = Typecheck<WritableUi>;
```
ryanatkn, line 340: using the `// @ts-expect-error` feature to avoid the confusing export here -- I always forget this is available, I think it'll be nice in some key places

src/lib/ui/uiEvents.ts
```diff
@@ -82,7 +82,7 @@ export const SelectCommunity: ClientEventInfo = {
 		$id: '/schemas/SelectCommunityParams.json',
 		type: 'object',
 		properties: {
-			community_id: {type: ['number', 'null']},
+			community_id: {type: 'number'},
```
ryanatkn, line 85: Given that we now have a community loaded for every persona, it no longer makes sense to select no community.

src/lib/ui/ui.ts
```diff
@@ -60,8 +60,7 @@ export interface Ui {
 	personaSelection: Readable<Readable<Persona> | null>;
 	personaIndexSelection: Readable<number | null>;
 	communitiesBySessionPersona: Readable<Map<Readable<Persona>, Array<Readable<Community>>>>;
-	communityIdSelectionByPersonaId: Readable<{[key: number]: number}>;
-	communityIdSelection: Readable<number | null>;
```
ryanatkn, line 64: Deleted  `communityIdSelection` because it was an unnecessary middle layer between `communitySelection`.

src/lib/vocab/community/communityMutations.ts
```diff
@@ -49,17 +49,34 @@ export const UpdateCommunitySettings: Mutations['UpdateCommunitySettings'] = asy
...
+
+	communityById.delete(community_id);
+	communities.mutate(($communites) => $communites.splice($communites.indexOf(community), 1)); // TODO use fast volatile remove instead, or maybe a set?
+	communityIdSelectionByPersonaId.mutate(($c) => {
+		for (const [persona_id, communityIdSelection] of $c) {
+			if (communityIdSelection === community_id) {
+				$c.set(persona_id, get(personaById.get(persona_id)!).community_id);
+			}
+		}
+	});
```
ryanatkn, line 79: this cleans up stale community selections for personas that aren't active (the value for the active persona is updated through the `goto` above)

src/lib/ui/CommunityDelete.svelte
```diff
@@ -60,14 +59,13 @@
...
 		<input
 			type="text"
-			id="name"
 			name="name"
 			placeholder=">enter name to unlock button"
 			bind:value={lockText}
 			on:keydown={onKeydown}
 		/>
 		<PendingButton {pending} disabled={locked || pending} on:click={deleteCommunity}>
-			Delete space
```
greatbacon, line 70: 🙃

src/lib/ui/ui.ts
greatbacon: 🧹

greatbacon (approved):
🚢

PR #343, opened 2022-04-26
add fallbacks for missing community and space

ryanatkn:
Improves handling of the URL pointing to communities and spaces that are nonexistent or inaccessible. This can happen if you're linked to a space/community you don't have access to, like if you're on the wrong account, or if the object is deleted and you navigate to it through a link or the back button. Previously the code sometimes handled `null` cases and sometimes ignored them.

Once we add authorization we can show more helpful messages, like whether a community/space exists or if you just don't have access to it. This isn't leaking information because we disallow duplicate community/space names.

next PR:

- route to the community after creating it
- route to the space after creating it


src/lib/ui/MainNav.svelte
```diff
@@ -46,7 +46,7 @@
 		</div>
 		<div class="explorer">
 			<CommunityNav />
-			{#if selectedPersona && selectedCommunity && selectedCommunitySpaces && selectedSpace}
+			{#if selectedPersona && selectedCommunity && selectedCommunitySpaces}
```
ryanatkn, line 49: makes it so you'll see the space nav even if there's no space selected (happens if you're navigating to an inaccessible or nonexistent space URL)

src/lib/ui/Workspace.svelte
```diff
@@ -42,14 +44,24 @@
...
+							space={selectedSpace}
+						/>
+					{:else}
+						<SpaceInput persona={selectedPersona} community={selectedCommunity} />
+					{/if}
+				{:else}
+					<CommunityInput persona={selectedPersona} />
+				{/if}
+			{:else}
+				<PersonaInput />
```
ryanatkn, line 64: this shows the creation forms, in the future we'll detect the existence of spaces/communities and give better error messages -- this is halfway to a good UX

PR #347, opened 2022-04-26
parallelize entity creation in entityServices test

ryanatkn:
parallelizes entities creation in the entityServices test, and also cleans up a couple unrelated things

### DIFF
--- a/src/lib/vocab/entity/EntityRepo.ts
+++ b/src/lib/vocab/entity/EntityRepo.ts
@@ -6,6 +6,7 @@ import {PostgresRepo} from '$lib/db/PostgresRepo';
 import type {Entity} from '$lib/vocab/entity/entity';
 import type {EntityData} from '$lib/vocab/entity/entityData';
 import type {RowList} from 'postgres';
+import type {ErrorResponse} from '$lib/util/error';
 
 const log = new Logger(gray('[') + blue('EntityRepo') + gray(']'));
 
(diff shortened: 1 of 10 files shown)


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -40,20 +40,20 @@ export class EntityRepo extends PostgresRepo {
 		const entities = await this.db.sql<Entity[]>`
 			SELECT entity_id, data, actor_id, space_id, created, updated 
 			FROM entities WHERE space_id= ${space_id}
-			ORDER BY created ASC
+			ORDER BY entity_id ASC
```
ryanatkn, line 43: didn't want to touch this here to avoid needing to change the frontend for something we may remove soon, but just wanted to point out that it's `ASC`

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -40,20 +40,20 @@ export class EntityRepo extends PostgresRepo {
...
 		if (entityIdSet.length === 0) return {ok: true, value: []};
 		log.trace('[findBySet]', entityIdSet);
 		const entities = await this.db.sql<Entity[]>`
 			SELECT entity_id, data, actor_id, space_id, created, updated 
 			FROM entities WHERE entity_id IN ${this.db.sql(entityIdSet)}
-			ORDER BY created DESC
+			ORDER BY entity_id DESC
 		`;
 		log.trace('entity count:', entities.length);
 		return {ok: true, value: entities};
```
ryanatkn, line 59: Wondering if this should return `false` if the count of entities differs from what's expected.

greatbacon: Probably worth dropping an error in the log at least

greatbacon (approved):

src/lib/vocab/entity/EntityRepo.ts
ryanatkn: made it an error so we'll be sure to notice for now

PR #348, opened 2022-04-26
refactor space selection

ryanatkn:
Converts `spaceIdSelectionByCommunityId` to use a `mutable` store. Similar to with `communityIdSelectionByPersonaId`, this makes the code a bit cleaner and more efficient, and makes the `community_id` correctly a `number` instead of having it be coerced to a `string`.


src/lib/ui/CommunityNavButton.svelte
```diff
@@ -20,8 +20,8 @@
 	export let community: Readable<Community>;
 	export let selected = false;
 
-	$: spaceIdSelection = $spaceIdSelectionByCommunityId[$community.community_id];
-	$: selectedSpace = spaceIdSelection === null ? null : spaceById.get(spaceIdSelection)!;
+	$: spaceIdSelection = $spaceIdSelectionByCommunityId.value.get($community.community_id);
+	$: selectedSpace = spaceIdSelection ? spaceById.get(spaceIdSelection)! : null;
```
ryanatkn, line 24: I simplified some null checking here and ahead, using a truthy ternary instead of an explicit `===` -- it's both slightly more concise and defensive in a good way, so if by some chance `undefined` slips in, things won't crash.

greatbacon (approved):

PR #350, opened 2022-04-26
use services instead of repos in seed

ryanatkn:
Refactors the seed file to use services instead of repos. The only one left is `repos.account.create`, and also `servicesIntegration.test.ts` needs some new service endpoints to replace its repo usage.

### DIFF
--- a/src/lib/db/seed.ts
+++ b/src/lib/db/seed.ts
@@ -14,6 +14,8 @@ import {parseView, type ViewData} from '$lib/vocab/view/view';
 import {createAccountPersonaService} from '$lib/vocab/persona/personaServices';
 import {createCommunityService} from '$lib/vocab/community/communityServices';
 import {toServiceRequest} from '$lib/util/testHelpers';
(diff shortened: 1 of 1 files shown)


src/lib/db/seed.ts
```diff
@@ -133,30 +150,35 @@ const entitiesContents: Record<string, string[]> = {
...
+			params: {
+				actor_id: persona_id,
+				data: {type: 'Collection', name: 'Grocery List'},
+				space_id: space.space_id,
+				source_id: space.directory_id,
 			},
-			space_id,
-		),
+			...serviceRequest,
+		}),
```
ryanatkn, line 167: Looks like ties weren't set up here, might fix some issues.

greatbacon: Yes, thanks for doing this! I kept meaning to go back to fix this & then also kept getting distracted lol

greatbacon (approved):

PR #352, opened 2022-04-26
change ViewSpace to be serializable

ryanatkn:
To make our events recordable+scriptable, their params need to be serializable. This is already true for all service events because we send params as JSON to the server, but 2 client events use stores, so this changes the first of them to have serializable params instead of a store.

### DIFF
--- a/src/lib/app/events.test.ts
+++ b/src/lib/app/events.test.ts
@@ -43,6 +43,7 @@ for (const eventInfo of eventInfos.values()) {
 		}
 
 		// TODO this fails because the random space is not availabe in the client UI data
+		// maybe make the randomizer configurable for populating server and/or client data
 		if (eventInfo.name === 'ViewSpace') {
 			return;
 		}
--- a/src/lib/app/events.test.ts
+++ b/src/lib/app/events.test.ts
@@ -42,6 +42,11 @@ for (const eventInfo of eventInfos.values()) {
 			return;
 		}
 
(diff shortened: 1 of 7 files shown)


src/lib/util/randomEventParams.ts
```diff
@@ -1,5 +1,4 @@
 import {randomBool} from '@feltcoop/felt/util/random.js';
-import {writable} from 'svelte/store';
```
ryanatkn, line 2: This is a red flag going forward -- the `randomEventParams.ts` module should never have non-serializable objects like stores. We'll probably find an elegant way to check this automatically.

src/lib/app/events.test.ts
```diff
@@ -42,6 +42,11 @@ for (const eventInfo of eventInfos.values()) {
 			return;
 		}
 
+		// TODO this fails because the random space is not availabe in the client UI data
+		if (eventInfo.name === 'ViewSpace') {
+			return;
+		}
```
ryanatkn, line 48: I would be ok with blocking this PR until thinking this through -- the problem is the tests try to dispatch events with data that the client `ui` doesn't have. I think this requires a bigger rethinking of how to populate data during tests.

greatbacon: Yeah, maybe the answer is to make the randomizer configurable for populating server and/or client data?

greatbacon (approved):

PR #354, opened 2022-04-27
Add entities to home view

greatbacon:
This PR adds the logic for backing the `Home` rules & norms with entities.

It checks to see if the expected entities already exist, creates defaults if they don't, and utilizes the EntityEditor for updates.


src/lib/ui/view/Home.svelte
```diff
@@ -1,17 +1,78 @@
...
+		: null;
+	let entities: Entity[] | undefined;
+	let rules: Readable<Entity> | undefined;
+	let norms: Readable<Entity> | undefined;
+
+	$: void entitiesResult?.then((data) => {
+		if (data.ok) {
+			entities = data.value.entities;
+			rules = writable(entities.find((e) => e.data.name === 'rules'));
+			norms = writable(entities.find((e) => e.data.name === 'norms'));
```
ryanatkn, line 53: looks good for this PR, once this and #334 are in we can do another PR that makes these reactive (currently if you edit the content you have to refresh to see the changes)

src/lib/ui/view/Home.svelte
```diff
@@ -31,28 +92,34 @@
...
+					on:click={() =>
+						dispatch.OpenDialog({
+							Component: EntityEditor,
+							props: {entity: rules},
+							dialogProps: {layout: 'page'},
+						})}
+					>propose change ✍️
+				</button>
+			</div>
+			{@html $rules ? $rules.data.content : 'rules not found'}
```
ryanatkn, line 107: I'll do a followup where we parse and mount this text so it can contain links and mentions and other components

ryanatkn (approved):

PR #356, opened 2022-04-29
improve types for randomEventParams, events, and services

ryanatkn:
- add typechecking to `randomEventParams`  by generating a new types file `randomEventParamsTypes.ts` via `randomEventParamsTypes.gen.ts`
  - this is necessary because TypeScript's generics don't support lookup types, I verified by getting some help in the TypeScript Discord with [this example](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wObBgNQIYA2ArsAGLQDSwAnnALxwA8AKlbaDMMgCapwDWNCADM4AZQgBbbAAsAlsgwA+ABSDqALjisaASi0TpMeYoDaO6gF16SuAG8AUAEhUAdzkwEMuGr33nTgh4qMBwAOToUGFajk5xUNhEUMjhwnIAbsBhANxwAPR5cKBgwAgwChhwCTBJKTDUJXAwEHAARqFcwtAIwNxNLZEVAQC+AUEh4chEktH+cU7VtXAArLkFReCl5YpViclNDaHNbR3IXVA9fcdTku1QIw5wT89P3MDCeEQEMDGPL-9PdRrQrIYCZe4AgHGKAQVxwACiUBhUBUumyfwBoxeo1GDgUnCgHx64iksgq-hekS0g0U6JeNy0Nzu6NGQA) of what I was trying to accomplish, it appears simple but it's unsupported :(
  - though it's a new generated file, it adds no bloat to the runtime because it's just types, and it's derived from the same data as the normal events file, so there's really low added cost in perf+maintenance
  - this required changing it from a single function to an object keyed by event name
- make service type declarations more concise by generating and using `ServiceByName` 
- capitalize service identifiers so they match the event name suffixed with "Service"
- added event name type unions to `eventTypes.ts` for `ServiceEvent` and `ClientEvent` to make the new `randomEventParams` typecheck for generic consumers, and has the side benefit of more strongly typing the `ClientEvent` and `ServiceEvent` interfaces


src/lib/vocab/community/communityServices.ts
```diff
@@ -74,112 +61,108 @@ export const readCommunityService: Service<ReadCommunityParams, ReadCommunityRes
 //Creates a new community for an instance
 // TODO think about extracting this to a `.services.` file
 // that imports a generated type and declares only `perform`
-export const createCommunityService: Service<CreateCommunityParams, CreateCommunityResponseResult> =
-	{
```
ryanatkn, line 78: This becomes a 1-liner with the new concise declaration, so it affects the indentation of the whole function.

src/lib/ui/uiEvents.ts
```diff
@@ -6,7 +6,6 @@ export const Ping: ServiceEventInfo = {
 	params: {
 		$id: '/schemas/PingParams.json',
 		type: 'null',
-		tsType: 'void',
```
ryanatkn, line 9: Fixes a hack for `Ping` and `LogoutAccount`, the two service events that have no params. We have to pass `null` as the actual params, but we don't want to have to pass `dispatch.LogoutAccount(null)`. Previously we accomplished this with the `tsType: 'void'` hack, but this creates a mismatch between the two, which rears its head with the now-typechecked `randomEventParams` function. Instead, the `eventTypes.gen.ts` now checks for `type: 'nuill'` when generating the dispatch types, and just omits params if null.

src/lib/util/randomEventParamsTypes.ts
```diff
@@ -0,0 +1,163 @@
+// generated by src/lib/util/randomEventParamsTypes.gen.ts
```
ryanatkn, line 1: Generating this file because of the TypeScript limitation described in the PR OP, see [this playground link](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wObBgNQIYA2ArsAGLQDSwAnnALxwA8AKlbaDMMgCapwDWNCADM4AZQgBbbAAsAlsgwA+ABSDqALjisaASi0TpMeYoDaO6gF16SuAG8AUAEhUAdzkwEMuGr33nTgh4qMBwAOToUGFajk5xUNhEUMjhwnIAbsBhANxwAPR5cKBgwAgwChhwCTBJKTDUJXAwEHAARqFcwtAIwNxNLZEVAQC+AUEh4chEktH+cU7VtXAArLkFReCl5YpViclNDaHNbR3IXVA9fcdTku1QIw5wT89P3MDCeEQEMDGPL-9PdRrQrIYCZe4AgHGKAQVxwACiUBhUBUumyfwBoxeo1GDgUnCgHx64iksgq-hekS0g0U6JeNy0Nzu6NGQA)

greatbacon (approved):

PR #358, opened 2022-05-04
upgrade deps

ryanatkn:
Upgrades to the latest SvelteKit and Svelte, and deletes some global types that are no longer necessary.


greatbacon (approved):
🧹

PR #360, opened 2022-05-06
add entity explorer explorable tree

ryanatkn:
Still needs a lot of refactoring, but the basics seem to be in place -- todo views are a good example.

It's a good usecase for the querying and client caching system.

todo

- make it show pagination buttons


greatbacon (approved):
lookin' slick

PR #367, opened 2022-05-11
update sidebar and nav styles

ryanatkn:
Experimental change so we can see how it looks -- makes the top navbar the same height as the luggage width+height and community nav width. This makes the top/left UI flush instead of the current jagged layout. The top nav is a little thick, which may not be desired overall, but it's helpful to give the community icon some whitespace padding.

An alternative to this design is to visually detach the top navbar from the left community nav, so the difference in size is less pronounced.


greatbacon:
Yeah the top feels a little chunky, I'm not super hot on that. Although maybe just adjusting the width on the luggage might work? Or yeah, try the detached form of it.

ryanatkn:
I agree it's too chunky, next I'll try reducing the community nav width so they can both shrink.

ryanatkn:
Ok it's ready to review again. I tried the changes like we talked about, but the problem is the mobile view -- the luggage is this big honkin thing that looks out of place next to the top navbar.

The solution that's currently implemented is to shrink the height of the luggage to match the navbar. I think I may prefer it to the square button, but let's talk about it.

Discord avoids the issue by making the menu button move offscreen, and so a new, differently sized button appears under your pointer.

PR #371, opened 2022-05-12
async contextmenu

ryanatkn:
Adds asynchronous behavior to contextmenu entries. If `action` returns a promise, it'll show a pending animation and if it returns a `Result` with a `message`, it'll show the error in a dialog. (and a little warning emoji in the menu, with the item title set to the error message)

future:

- make each entry a store and don't mutate?


src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -15,23 +15,28 @@
...
 
 <ContextmenuSubmenu>
-	<svelte:fragment slot="entry">
-		<span class="menu-item-entry">
-			<UnicodeIcon icon="⚙️" /><span class="title">Account</span></span
-		>
+	<svelte:fragment slot="icon">
+		<UnicodeIcon icon="⚙️" />
 	</svelte:fragment>
+	AccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccountAccount
```
ryanatkn, line 21: TODO fix before merging

src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -42,20 +47,30 @@
 					props: {done: () => dispatch.CloseDialog()},
 				})}
 		>
-			<span class="title">Create Persona</span>
+			Create PersonaCreate PersonaCreate PersonaCreate PersonaCreate PersonaCreate PersonaCreate
+			PersonaCreate Persona
```
ryanatkn, line 51: TODO fix before merging

src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -42,20 +47,30 @@
...
+							// 	Component: About,
+							// }),
+							resolve({
+								ok: false,
+								message:
+									'yes it is error yes it is erroryes it is erroryes it is erroryes it is erroryes it is error',
+							}),
+						200,
+					),
+				)}
```
ryanatkn, line 68: TODO fix before merging

src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -42,20 +47,30 @@
...
 		>
-			<span class="title">About</span>
+			AboutAboutAboutAboutAboutAboutAboutAboutAboutAboutAbout
```
ryanatkn, line 70: TODO fix before merging

src/lib/server/httpServiceMiddleware.ts
```diff
@@ -59,6 +61,13 @@ export const toHttpServiceMiddleware =
 			const validationError = validateParams.errors![0];
 			return send(res, 400, {message: toValidationErrorMessage(validationError)});
 		}
+		if (service.event.name === 'LogoutAccount') {
+			await wait(1500);
+			return send(res, 500, {
+				message:
+					'errr expernced on the thingexpernced on the thingexpernced on the thingexpernced on the thing',
+			});
+		}
```
ryanatkn, line 70: TODO fix before merging

src/routes/__layout.svelte
```diff
@@ -54,7 +55,9 @@
 			() => dispatch.Ping(),
 		),
 	);
-	const ui = toUi(session, initialMobileValue, components);
+	const ui = toUi(session, initialMobileValue, components, (errorMessage) => {
+		dispatch.OpenDialog({Component: ErrorMessage, props: {text: errorMessage}});
+	});
```
ryanatkn, line 60: Not 100% sure about this API but it seems ok for now. If we like it, putting `onError` directly on the UI so it can be called from anywhere probably makes sense. An alternative design would make this a new dispatched event, not a special callback.

ryanatkn, line 60: Not 100% sure about this API but it seems ok for now. If we like it, putting `onError` directly on the UI so it can be called from anywhere probably makes sense. An alternative design would make this a new dispatched event, not a special callback.

greatbacon: looks ok to me for now!

greatbacon (approved):
🚀

PR #372, opened 2022-05-16
add gettable stores

ryanatkn:
Implements gettable stores. Uses the new [`@feltcoop/svelte-gettable-stores`](https://github.com/feltcoop/svelte-gettable-stores/) instead of `svelte/store`.

This is purely an optimization, but given our usage I believe it's a UX-relevant one with realworld amounts of data. Might want to do some benchmarking for curiosity. I think it's very hard to do non-contrived measurements to know the actual runtime effects on UX, however. We don't know what typical data load looks like, and I'm guessing a microbenchmark will make the gettable stores look *too* good. I'm happy with the gettable store implementation; I'll explain more in its library.


greatbacon (approved):
🥇

PR #373, opened 2022-05-16
Fix infra scripts

greatbacon:
This PR updates references to Ubuntu to the latest LTS (22)

It also includes a new doc for manually upgrading instances as well as tweaking some package.json to account for the NODE_ENV now properly being set to `production`.


src/infra/setup.task.ts
```diff
@@ -75,6 +75,9 @@ export const task: Task<SetupTaskArgs> = {
 				`apt install -y unzip;
 				curl -fsSL https://fnm.vercel.app/install | bash;
 				export PATH=/root/.fnm:$PATH;
+				echo "export PATH=$PATH
+				eval "\`fnm env\`";
```
greatbacon, line 79: So I'm having some issue with this line. I want to prepend the literal ```
eval "`fnm env`"
```
to this file (otherwise fnm doesn't initialize properly in the serverDeploy script & you can't use npm)

However, when calling this code `fnm env` is firing, just printing a bunch of path info instead. Any thoughts?

File: package.json [hunk 2]
```diff
@@ -22,49 +22,49 @@
...
-    "throttle-debounce": "^4.0.0",
     "tslib": "^2.3.1",
     "typescript": "^4.6.3"
   },
   "dependencies": {
     "@feltcoop/felt": "^0.28.0",
+    "@feltcoop/svelte-mutable-store": "^0.1.2",
     "@polka/send-type": "^0.5.2",
     "@ryanatkn/json-schema-to-typescript": "^11.1.2",
+    "@types/throttle-debounce": "^2.1.0",
```
ryanatkn, line 52: types we can keep as dev deps

src/infra/README.md
```diff
@@ -56,4 +56,18 @@ npm run deploy
...
+
+## manually upgrading cloud instances
+
+If upgrading your underlying cloud OS, we recommend spinning up a new server instance and restoring a DB backup to it. This helps keeps cruft from forming on your servers, helps you make sure you running on the latest hardware from your cloud provider, and tests your DB restoration process all in one.
+
+1. Spin up a new server with the new OS
+1. Point your DNS record to the new IP
+1. Run `infra/setup` on the new server
+1. Set up your DB password
+1. Take a dump of the DB from your old server instance
```
ryanatkn, line 70:
```suggestion
1. Use [`pg_dump`](https://www.postgresql.org/docs/current/backup-dump.html) to copy the DB from your old server instance
```

phrasing :no_mouth:

src/infra/setup.task.ts
```diff
@@ -90,7 +93,8 @@ export const task: Task<SetupTaskArgs> = {
 			logSequence('Installing pm2 and gro...') +
 				`export NODE_ENV=production;
 				echo "export NODE_ENV=production" >> ~/.profile;
-				echo "export NODE_ENV=production" >> ~/.bashrc;
+				echo "export NODE_ENV=production
+				$(cat ~/.bashrc)" > ~/.bashrc;
```
ryanatkn, line 97: what does this line do?

src/infra/setup.task.ts
ryanatkn: First guess is the `"` needs escaping on line 79. Not sure if you can use single quotes around the echo instead.

src/infra/setup.task.ts
greatbacon: This prepends the `NODE_ENV=production` line to the ~/.bashrc file, allowing `ssh` to source it to the terminal before bailing out of the rest of the .bashrc file.

ryanatkn: gotcha nice, can you put it on one line so it's clearer what's in the quotes? or if multiline is needed like above, indent it?

ryanatkn (approved):

PR #379, opened 2022-05-18
test

greatbacon:
test


greatbacon:
Closing; devs say they think they've fixed it; https://github.com/actions/virtual-environments/issues/5490#issuecomment-1129141874

Might be a while before the fix fully rolls out.

Will follow up after 48 hours :

PR #380, opened 2022-05-18
upgrade deps

ryanatkn:
Mainly wanted to upgrade `postgres` because it has a bunch of fixes.

Also upgrades SvelteKit. As far as I can tell, there are no runtime issues introduced, but before the page hydrates on load the layout gets strange and then pops into place. Seems fine for now to not worry about.


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -21,13 +21,13 @@ export class EntityRepo extends PostgresRepo {
 		if (space_id) {
 			entity = await this.db.sql<Entity[]>`
 			INSERT INTO entities (actor_id, space_id, data) VALUES (
-				${actor_id},${space_id},${this.db.sql.json(data)}
+				${actor_id},${space_id},${this.db.sql.json(data as any)}
```
ryanatkn, line 24: The casting here is unfortunate but it's a minor localized thing, shouldn't cause bugs.

greatbacon (approved):

PR #384, opened 2022-05-20
change random entity and tie to take a source_id instead of a space

ryanatkn:


src/lib/util/randomVocab.ts
```diff
@@ -211,7 +213,7 @@ export class RandomVocabContext {
 		persona?: Persona,
 		account?: Account,
 		community?: Community,
-		space?: Space,
+		parentSourceId?: number, // optional directory or other source id for the source and dest entities (not the tie)
```
ryanatkn, line 216: calling this `source_id` seemed confusing because it's not `tie.source_id`

greatbacon (approved):

PR #386, opened 2022-05-20
Refactors the `actor_id` attribute to `persona_id` on entities

greatbacon:
This PR came out of a conversation we had about the terminology of `actor` (which is used in the ActivityStreams spec) & persona (which is essentially our system implementation of actor). This brings the system into an internal alignment.


ryanatkn (approved):
# :broom:

PR #393, opened 2022-05-31
Add directories to session

greatbacon:
This PR finishes the back half of our "freshness" feature. It adds `directories` to the session object, uses those directories for calculating freshness, and sends update events on those entities on a view by view basis.

Plus now localstore dates or timestamps


src/lib/vocab/space/SpaceRepo.ts
```diff
@@ -32,6 +32,19 @@ export class SpaceRepo extends PostgresRepo {
...
+		const data = await this.db.sql<Space[]>`
+			SELECT s.space_id, s.name, s.url, icon, s.view, s.updated, s.created, s.community_id, s.directory_id
+			FROM spaces s JOIN (
+				SELECT DISTINCT m.community_id FROM personas p
+				JOIN memberships m ON p.persona_id=m.persona_id AND p.account_id=${account_id}
+			) apc
+			ON s.community_id=apc.community_id;
+		`;
+		return {ok: true, value: data};
+	}
```
ryanatkn, line 46: This looks is unused and unchanged from `filterByAccount`, I'm guessing the inline code in the SessionRepo belongs here?

src/lib/session/SessionRepo.ts
```diff
@@ -12,23 +14,33 @@ export class SessionRepo extends PostgresRepo {
 		const accountResult = await this.db.repos.account.findById(account_id);
 		if (!accountResult.ok) return accountResult;
 		const account = accountResult.value;
+
+		const data = await this.db.sql<Array<{space: Space; entity: Entity}>>`
+		SELECT json_build_object('space_id',s.space_id,'name',s.name,'url',s.url,'icon',s.icon,'view',s.view,'created',s.created,'updated',s.updated,'community_id',s.community_id,'directory_id',s.directory_id) space, 
+		json_build_object('entity_id',e.entity_id,'data',e.data,'persona_id',e.persona_id,'created',e.created,'updated',e.updated) entity
```
ryanatkn, line 20: neat, I was wondering how to return values from multiple tables

ryanatkn, line 20: neat, I was wondering how to return values from multiple tables

src/lib/ui/ui.ts
```diff
@@ -323,8 +326,9 @@ export const toUi = (
 						: $session.spaces.map(($space) => [
 								$space.directory_id,
 								writable(
-									(browser && localStorage.getItem(`${LAST_SEEN_KEY}${$space.directory_id}`)) ||
-										new Date().toString(),
+									(browser &&
+										Number(localStorage.getItem(`${LAST_SEEN_KEY}${$space.directory_id}`))) ||
+										new Date().getTime(),
```
ryanatkn, line 331:
```suggestion
										Date.now(),
```

ryanatkn, line 331:
```suggestion
										Date.now(),
```

src/lib/vocab/space/SpaceRepo.ts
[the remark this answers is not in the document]
greatbacon: I think I stubbed this out, and then when I started messing around with queries I realized I was primarily messing with the `SessionRepo`, which by our naming conventions is the primary place for SQL.

I think I'm leaning towards deleting this section & leaving the other inline query in place to build off of? But could also just move that query to the `SpaceRepo` here & name it `filterByAccountWithDirectories`. Thoughts?

src/lib/ui/view/Home.svelte
```diff
@@ -95,6 +95,7 @@
 		<div class="rules markup panel-inset">
 			<div class="header">
 				<h4>rules</h4>
+				<!--TODO how to trigger a directory freshen from result of this dialogue-->
```
ryanatkn, line 98: Good question, two problems problems to think about and address in future PRs:

- the `EntityEditor` is a generic component, and making it work differently by view is going to be complex -- in other cases, we're inlining `UpdateEntity` everywhere we dispatch certain events, but that doesn't work in generic code
- if the entity is in multiple directories, the others won't be marked as fresh (this makes sense because the logic is view-specific, but unfixed it'll be a bad UX)

src/lib/vocab/space/SpaceRepo.ts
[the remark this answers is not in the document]
ryanatkn: Gotcha, either sounds good to me, depends if we think we'll reuse the spaces+directory query for other purposes (seems likely? idk tho) or if we'll continue refactoring the session query to consolidate DB calls.

ryanatkn (approved):

src/lib/vocab/space/SpaceRepo.ts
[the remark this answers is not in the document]
greatbacon: Spaces + directories seems like a useful query, so I'll opt to move that down into the SpaceRepo!

src/lib/ui/ui.ts
```diff
@@ -288,6 +288,9 @@ export const toUi = (
 			$spaces.forEach((s, i) => spaceById.set($spaceArray[i].space_id, s));
 			spaces.swap($spaces);
 
+			const $directoriesArray = $session.guest ? [] : $session.directories;
+			$directoriesArray.forEach((d) => entityById.set(d.entity_id, writable(d)));
```
ryanatkn, line 292:
```suggestion
			$directoriesArray.forEach((d) => entityById.set(d.entity_id, writable(d)));
```

Should probably use `updateEntity` here from `entityMutationHelpers.ts`, is identical right now but may change.

PR #395, opened 2022-06-01
save selected spaces by community to local storage

ryanatkn:
This changes `spaceIdSelectionByCommunityId` to be loaded from localStorage if available, so when you refresh the page, each community nav button links to the last-selected space for that community on your device.

It also adds a `locallyStored` store helper that augments stores with storage. The API could use improvements but it's a decent start and I'm open to refining it before merging.

The `locallyStored` helper is used in the other places we used `localStorage`, and it also includes two helper functions work working with JSON in `localStorage`, given that it only stores strings. We'll probably use IndexedDB for this eventually so it may not be worth much effort right now. There's disabled logging on save/load to storage that can be used to get a feel for how much it's hitting it. We may want to batch some of these values into single JSON collections to reduce the read count, but that may be unnecessary for IndexedDB.

followup:

- namespace by `account_id`
- try to improve the type for usage
- think through handling the window `storage` event


src/lib/ui/ui.ts
```diff
@@ -338,9 +345,15 @@ export const toUi = (
 						? null
 						: $session.communities.map(($community) => [
 								$community.community_id,
-								$session.spaces.find(
-									(s) => s.community_id === $community.community_id && isHomeSpace(s),
-								)!.space_id,
+								spaceIdSelectionByCommunityId
```
ryanatkn, line 348: review this

src/lib/ui/locallyStored.ts
```diff
@@ -0,0 +1,75 @@
...
+	const stored: TStore & {getJson: () => TJson} = {
+		...store,
+		getJson: (): TJson =>
+			json === undefined
+				? (json = toJson(mutate || swap ? (store.get() as any).value : store.get()))
+				: json,
+	};
+	if (set) {
+		stored.set = function () {
+			set.apply(this, arguments as any); // eslint-disable-line prefer-rest-params
```
ryanatkn, line 52: probably return the result of this

src/lib/util/storage.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @returns
+ */
+export const loadFromStorage = <T extends Json>(
+	key: string,
+	validate?: (value: any) => asserts value is T,
+): T | undefined => {
+	if (!browser) return;
+	const stored = localStorage.getItem(key);
+	if (!stored) {
+		log.trace('loaded nothing', key);
```
ryanatkn, line 23: remove logging

greatbacon (approved):

src/lib/ui/ui.ts
ryanatkn: was able to simplify with `find` and optional chaining

PR #396, opened 2022-06-01
extract websocket component

ryanatkn:
Renames `SocketConnection` to `SocketConnectionControls` and adds `SocketConnection` to extract a bit of logic out of the main layout. It should change no behavior, the idea is to keep the layout tidy and make the websocket less tightly coupled to it.

### DIFF
--- a/src/lib/ui/SocketConnectionControls.svelte
+++ b/src/lib/ui/SocketConnectionControls.svelte
@@ -1,0 +1,16 @@
+<script lang="ts">
+	import type {SocketStore} from '$lib/ui/socket';
+
+	export let socket: SocketStore;
+
+	const connect = () => socket.connect($socket.url!); // TODO maybe make the socket state a union type for connected/disconnected states?
+	const disconnect = () => socket.disconnect();
+</script>
+
+<form>
+	<input value={$socket.url} on:input={(e) => socket.updateUrl(e.currentTarget.value)} />
+	<button type="button" on:click={$socket.ws ? disconnect : connect}>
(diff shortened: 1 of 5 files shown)


src/lib/ui/SocketConnectionControls.svelte
```diff
@@ -0,0 +1,16 @@
...
+	const disconnect = () => socket.disconnect();
+</script>
+
+<form>
+	<input value={$socket.url} on:input={(e) => socket.updateUrl(e.currentTarget.value)} />
+	<button type="button" on:click={$socket.ws ? disconnect : connect}>
+		{#if $socket.ws}disconnect{:else}connect{/if}
+	</button>
+</form>
+<div>status: <code>'{$socket.status}'</code></div>
```
ryanatkn, line 16: confused git but this file was only renamed

src/lib/ui/socket.ts
```diff
@@ -1,25 +1,15 @@
...
 const RECONNECT_DELAY = 1000; // matches the current Vite/SvelteKit retry rate, but we use a counter to back off
 const RECONNECT_DELAY_MAX = 60000;
 
-export const getSocket = (): SocketStore => getContext(KEY);
-
-export const setSocket = (store: SocketStore): SocketStore => {
-	setContext(KEY, store);
-	return store;
-};
-
```
ryanatkn, line 22: this context stuff was unused, I think we'll keep going with the "app" pattern if something wants access to the socket.

greatbacon (approved):

PR #397, opened 2022-06-01
extract `SetSession` event

ryanatkn:
Changes the `ui`'s only magic method `setSession` to be an event like all the rest. It simplifies the `ui` dependencies and internal magic (including removing the `destroy` method) by hoisting the behavior up to the layout and using automatic store subscriptions.

Saving the session/account change for a followup PR.


src/lib/ui/uiEvents.ts
```diff
@@ -47,7 +47,7 @@ export const OpenDialog: ClientEventInfo = {
 	name: 'OpenDialog',
 	params: {
 		$id: '/schemas/OpenDialogParams.json',
-		allOf: [{$ref: '/schemas/DialogData.json'}],
+		$ref: '/schemas/DialogData.json',
```
ryanatkn, line 50: This is just a simplification.

src/lib/app/dispatch.ts
```diff
@@ -45,7 +45,7 @@ export const toDispatch = (
 			log.trace(
 				'%c[dispatch.%c' + eventName + '%c]',
 				'color: gray',
-				'color: cyan',
+				'color: cornflowerblue',
```
ryanatkn, line 48: We changed the broadcast color to `darkCyan` but forgot this one.

src/lib/session/sessionEvents.ts
```diff
@@ -0,0 +1,21 @@
+import type {ClientEventInfo} from '$lib/vocab/event/event';
+
+export const SetSession: ClientEventInfo = {
```
ryanatkn, line 3: Seems we have an inconsistency here, where `LoginAccount` and `LogoutAccount` are in `sessionMutations.ts` along with `SetSession`, but there are 2 different modules, `accountEvents` and `sessionEvents`. I think they should be consolidated, but is it more session-y or account-y? I think I'm leaning towards session but IDK.

src/routes/__layout.svelte
```diff
@@ -88,7 +88,8 @@
 		Object.assign(window, app);
 		log.trace('app', app);
 	}
-	$: browser && log.trace('$session', $session);
+
+	$: dispatch.SetSession({session: $session});
```
ryanatkn, line 92: There's a slight change of behavior here because the session is initialized after component initialization instead of during. (the `$: ` means it'll run *after* initializing) It doesn't appear to cause bugs or issues, but it seems likely to be problematic.

We could set it synchronously with `dispatch.SetSession({session: $session});` but then we'd have to skip the first dispatch somehow. I've considered making a helper for that purpose in other contexts, but this would be simpler, so we'll see if it's stable.

greatbacon (approved):

PR #400, opened 2022-06-03
Add community freshness

greatbacon:
This PR builds on the freshness work at the space level & brings it to the community level.

### DIFF
--- a/src/lib/ui/CommunityNavButton.svelte
+++ b/src/lib/ui/CommunityNavButton.svelte
@@ -11,7 +11,14 @@
 
 	const {
 		dispatch,
-		ui: {contextmenu, spaceIdSelectionByCommunityId, spaceById, sessionPersonaIndices},
+		ui: {
+			contextmenu,
+			spaceIdSelectionByCommunityId,
+			spaceById,
+			sessionPersonaIndices,
+			spacesByCommunityId,
+			freshnessByDirectoryId,
+		},
 	} = getApp();
 
 	// TODO should this just use `ui` instead of taking all of these props?
@@ -27,6 +34,10 @@
 	$: isPersonaHomeCommunity = $community.name === $persona.name;
 
 	$: personaIndex = $sessionPersonaIndices.get(persona)!;
+
+	$: spaces = $spacesByCommunityId.get($community.community_id);
+	$: highlighted =
+		spaces && !spaces.every((s) => !$freshnessByDirectoryId.get(s.get().directory_id)!.get());
 </script>
 
 <!-- TODO can this be well abstracted via the Entity with a `link` prop? -->
@@ -36,6 +47,7 @@
 		persona: personaIndex + '',
 	})}
 	class:selected
+	class:highlighted
 	class:persona={isPersonaHomeCommunity}
 	style="--hue: {$community.settings.hue}"
 	use:contextmenu.action={[[CommunityContextmenu, {community, persona}]]}
@@ -64,5 +76,10 @@
 		width: calc(var(--icon_size_md) + var(--spacing_xs) * 2);
 		height: calc(var(--icon_size_md) + var(--spacing_xs) * 2);
 		--icon_size: var(--icon_size_sm);
+	}
+	/* TODO figure out why selected doesn't override highlighted */
+	.highlighted {
+		border-left: 2px;
+		background-color: grey;
 	}
 </style>
--- a/src/lib/ui/CommunityNavButton.svelte
+++ b/src/lib/ui/CommunityNavButton.svelte
@@ -72,9 +72,7 @@
 		display: flex;
 		justify-content: center;
 		align-items: center;
-		/* TODO do this layout without calculating */
-		width: calc(var(--icon_size_md) + var(--spacing_xs) * 2);
-		height: calc(var(--icon_size_md) + var(--spacing_xs) * 2);
+		height: var(--luggage_size);
 		--icon_size: var(--icon_size_sm);
 	}
 	/* TODO figure out why selected doesn't override highlighted */
--- a/src/lib/ui/CommunityNavItem.svelte
+++ b/src/lib/ui/CommunityNavItem.svelte
@@ -35,7 +35,7 @@
 	$: personaIndex = $sessionPersonaIndices.get(persona)!;
 
 	$: fresh = freshnessByCommunityId.get($community.community_id);
-	$: console.log(`% ${$community.community_id}`, fresh);
+	$: console.log(`% ${$community.community_id}`, $fresh);
 </script>
 
 <!-- TODO can this be well abstracted via the Entity with a `link` prop? -->
--- a/src/lib/ui/CommunityNavItem.svelte
+++ b/src/lib/ui/CommunityNavItem.svelte
@@ -42,7 +42,7 @@
 
 <!-- TODO can this be well abstracted via the Entity with a `link` prop? -->
 <a
-	class="community"
+	class="community selectable"
 	href={toSpaceUrl($community, selectedSpace && $selectedSpace, $page.url.searchParams, {
 		persona: personaIndex + '',
 	})}
@@ -63,10 +63,6 @@
 		/* TODO better way to have active state? this makes the community nav wider than the luggage button! */
 		padding: var(--spacing_xs);
 		text-decoration: none;
-	}
-	/* TODO jucier selected state, maybe scaling up 10ish percent */
-	.selected {
-		background-color: var(--bg);
 	}
 	.persona {
 		display: flex;
--- a/src/lib/ui/CommunityNavItem.svelte
+++ b/src/lib/ui/CommunityNavItem.svelte
@@ -8,6 +8,7 @@
 	import {getApp} from '$lib/ui/app';
 	import {toSpaceUrl} from '$lib/ui/url';
 	import CommunityContextmenu from '$lib/app/contextmenu/CommunityContextmenu.svelte';
+	import FreshnessIndicator from '$lib/ui/FreshnessIndicator.svelte';
 
 	const {
 		dispatch,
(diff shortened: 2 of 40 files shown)


src/lib/ui/CommunityNavButton.svelte
```diff
@@ -63,4 +75,9 @@
 		height: var(--luggage_size);
 		--icon_size: var(--icon_size_sm);
 	}
+	/* TODO figure out why selected doesn't override highlighted */
+	.highlighted {
```
greatbacon, line 79: This styling is super lightweight & is a good candidate for some pairing before merging.

src/lib/ui/ui.ts
```diff
@@ -59,7 +59,7 @@ export interface Ui {
 	//TODO maybe refactor to remove store around map? Like personaById
 	spacesByCommunityId: Readable<Map<number, Array<Readable<Space>>>>;
 	personasByCommunityId: Readable<Map<number, Array<Readable<Persona>>>>;
-	entityById: Map<number, Readable<Entity>>; // TODO mutable inner store
+	entityById: Mutable<Map<number, Readable<Entity>>>; // TODO mutable inner store
```
greatbacon, line 62: I ended up making this Mutable to get proper rebuild in when an Entity updates, but I'm not 100% if that's really the right path here.

src/lib/ui/CommunityNavButton.svelte
ryanatkn: Cool yeah, we may need to make a separate `StatusIndicator` component or something, it may not be possible to get the desired styling with a single class like I had hoped.

src/lib/ui/ui.ts
```diff
@@ -219,7 +220,25 @@ export const toUi = (
...
+			const map: Map<number, Writable<boolean>> = new Map();
+			for (const space of $spaces.value) {
+				const $space = space.get();
+				const $directory = $entityById.value.get($space.directory_id)!.get();
+				const $lastSeen = $lastSeenByDirectoryId.value.get($space.directory_id)!.get();
+				const systemTime = $directory.updated ?? $directory.created;
+
+				const clientTime = new Date($lastSeen);
+				const freshness = clientTime.getTime() < new Date(systemTime).getTime();
+				map.set($space.directory_id, writable(freshness));
```
ryanatkn, line 237: I was thinking about this in the space freshness PR too, I think a store per item may be too heavy handed, because when freshness is reset, `lastSeenByDirectoryId` changes too, which means it needs to recalculate everything, so I think the store is unnecessary.

src/lib/ui/ui.ts
```diff
@@ -219,7 +220,25 @@ export const toUi = (
...
+		([$spaces, $lastSeenByDirectoryId, $entityById]) => {
+			const map: Map<number, Writable<boolean>> = new Map();
+			for (const space of $spaces.value) {
+				const $space = space.get();
+				const $directory = $entityById.value.get($space.directory_id)!.get();
+				const $lastSeen = $lastSeenByDirectoryId.value.get($space.directory_id)!.get();
+				const systemTime = $directory.updated ?? $directory.created;
+
+				const clientTime = new Date($lastSeen);
+				const freshness = clientTime.getTime() < new Date(systemTime).getTime();
```
ryanatkn, line 236:
```suggestion
				const freshness = $lastSeen < systemTime.getTime();
```

src/lib/ui/ui.ts
```diff
@@ -219,7 +220,25 @@ export const toUi = (
 	);
 	const lastSeenByDirectoryId = mutable<Map<number, Writable<number> | null>>(new Map());
 	// TODO this does not have an outer `Writable` -- do we want that much reactivity?
-	const entityById: Map<number, Writable<Entity>> = new Map();
+	const entityById = mutable<Map<number, Writable<Entity>>>(new Map());
+
+	const freshnessByDirectoryId: Readable<Map<number, Writable<boolean>>> = derived(
+		[spaces, lastSeenByDirectoryId, entityById],
+		([$spaces, $lastSeenByDirectoryId, $entityById]) => {
```
ryanatkn, line 227: Did you try to see things worked reactively without changing `entityById` to a store? I would expect that to work, and if it doesn't there's another issue to think through. (like the entity mutations may need to change some state imperatively)

src/lib/ui/ui.ts
ryanatkn: Did you try to see if it works reactively without changing `entityById` to a store? I would expect that to work, and if it doesn't there's another issue to think through. (like maybe the entity mutations need to change some state imperatively) I'm pretty sure we should revert entityById back to a regular map, as implemented, everywhere it's used reactively, it'll recompute whenever any entity is added anywhere.

src/lib/ui/SpaceName.svelte
```diff
@@ -7,20 +7,16 @@
...
-		ui: {lastSeenByDirectoryId, entityById},
+		ui: {freshnessByDirectoryId},
 	} = getApp();
 
-	$: directory = entityById.get($space.directory_id)!;
-	$: lastSeen = $lastSeenByDirectoryId.value.get($space.directory_id)!;
-	$: systemTime = $directory.updated ?? $directory.created;
-	$: clientTime = new Date($lastSeen);
-	$: fresh = clientTime < systemTime;
+	$: fresh = $freshnessByDirectoryId.get($space.directory_id)!;
```
ryanatkn, line 13: this is great, moving the logic to the `freshnessByDirectoryId`

src/lib/ui/ui.ts
```diff
@@ -74,6 +74,7 @@ export interface Ui {
 	spaceIdSelectionByCommunityId: Mutable<Map<number, number | null>>;
 	spaceSelection: Readable<Readable<Space> | null>;
 	lastSeenByDirectoryId: Mutable<Map<number, Writable<number> | null>>;
+	freshnessByDirectoryId: Readable<Map<number, Writable<boolean>>>;
```
ryanatkn, line 77: What do you think of renaming this `freshByDirectoryId`, given `$: fresh = $freshnessByDirectoryId.get($space.directory_id)!;`

greatbacon: I think it reads nicer this way, but I'm not attached.

src/lib/ui/ui.ts
greatbacon: Maybe we can pair on this Monday?

src/lib/ui/ui.ts
greatbacon: I couldn't figure out the derived nature of `freshnessByDirectoryId` without it, but maybe that logic could live in a space by space component or something? Or some other option I'm not aware of.

ryanatkn: My thought was to remove it from the dependencies list of the `derived`, which means the derived calculation will no longer re-run when entities get added or removed, but that shouldn't be needed. The idea is that `entityById` is a plain non-reactive map for looking up any entity for which you have an id. In this case, freshness needs to change when `$spaces` and `$lastSeenByDirectoryId` have new values, and when their data is used to lookup entities, they entity store should already be available in `entityById`.

The reasoning behind the design was to keep adding/removing entities from the global stores lightweight and not observed -- any work like updating caches can be done imperatively in the mutations. Because entities will be changing often, this design should avoid a primary bottleneck.

src/lib/ui/CommunityNavItem.svelte
```diff
@@ -27,6 +34,11 @@
 	$: isPersonaHomeCommunity = $community.name === $persona.name;
 
 	$: personaIndex = $sessionPersonaIndices.get(persona)!;
+
+	$: spaces = $spacesByCommunityId.get($community.community_id) || [];
+
+	$: highlighted = !spaces.every((s) => !freshnessByDirectoryId.get(s.get().directory_id)!.get());
```
greatbacon, line 40: @ryanatkn So I got the space level state updates into a good place by removing the Mutable around the map. But now there's an issue with the actual meat of this PR. Which is this highlighted attribute isn't reactive when these underlying freshness stores update. Any thoughts?

ryanatkn: First thought is like we have `freshnessByDirectoryId`, add `freshnessByCommunityId` and then imperatively update it rather than trying to derived it. (I think technically it could be a `derived` taking as input the array of stores for each of a community's spaces)

src/lib/vocab/entity/entityMutations.ts
```diff
@@ -44,6 +44,7 @@ export const DeleteEntities: Mutations['DeleteEntities'] = async ({invoke, param
 	const {entity_ids} = params;
 	for (const entity_id of entity_ids) {
 		ui.entityById.delete(entity_id);
+		ui.freshnessByDirectoryId.delete(entity_id);
```
ryanatkn, line 47: we talked about moving this to `DeleteSpace` and adding the corresponding one for `CreateSpace`, which will need the directory entity added to the response payload

src/lib/ui/CommunityNavItem.svelte
```diff
@@ -36,6 +45,7 @@
 		persona: personaIndex + '',
 	})}
 	class:selected
+	class:highlighted={fresh}
```
ryanatkn, line 48: i think:

```suggestion
	class:highlighted={$fresh}
```

src/lib/ui/uiMutationHelper.ts
```diff
@@ -0,0 +1,27 @@
...
+		entity_id,
+		derived([directory, lastSeen], ([$directory, $lastSeen]) => {
+			return $lastSeen < ($directory.updated ?? $directory.created).getTime();
+		}),
+	);
+};
+
+export const upsertCommunityFreshnessById = (ui: WritableUi, community_id: number): void => {
+	const {spacesByCommunityId, freshnessByCommunityId, freshnessByDirectoryId} = ui;
+	const spaces = spacesByCommunityId.get().get(community_id) || [];
```
ryanatkn, line 22: wondering if we want to use `derived` from `spacesByCommunityId`

ryanatkn:
I think the conflicts in the other branch should supercede what we did here, but I may be mistaken

src/lib/ui/uiMutationHelper.ts
greatbacon: ooh, that would get us free work if spaces get added/removed from a community

ryanatkn (approved):

PR #402, opened 2022-06-09
remove border radius from luggage and marquee buttons

ryanatkn:
Makes the luggage and marquee buttons square. I tried making just the top corner have zero radius, but it looks a bit strange, and the community/space links all have square corners, so I think this is better.

I don't love how the outline style for the focused element gets cut off for various reasons, like going off the page (see the community links) or being overlapped by other elements (see the space links), but I'm not sure what to do about it.

### DIFF
--- a/src/lib/ui/CommunityNav.svelte
+++ b/src/lib/ui/CommunityNav.svelte
@@ -70,6 +70,7 @@
 		align-items: center;
 	}
 	.persona-group {
+		width: 100%;
 		margin-bottom: var(--navbar_size);
 	}
 	.persona-group:last-child {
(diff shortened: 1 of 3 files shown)


src/lib/ui/CommunityNav.svelte
```diff
@@ -70,6 +70,7 @@
 		align-items: center;
 	}
 	.persona-group {
+		width: 100%;
```
ryanatkn, line 73: This fixes the width of the community button group when a persona is joined to no communities other than its personal community.

greatbacon (approved):

PR #407, opened 2022-06-14
add `data.space_id` for directory entities

ryanatkn:
- [x] create directories with `data.space_id`
- [x] test
- [x] add migration

We should be able to detect directories with type safety via `'space_id' in $entity.data`, they continue to have `type: 'Collection'` but no longer have `name: 'directory'`.

### DIFF
--- a/src/lib/db/migrations/00025-directory-data-space-community-ids.js
+++ b/src/lib/db/migrations/00025-directory-data-space-community-ids.js
@@ -1,0 +1,6 @@
+/** @param {import('postgres').Sql<any>} sql */
+export const up = async (sql) => {
+	// TODO BLOCK for every `space.directory_id` set `entity.data.community_id` and `entity.data.space_id`
+	// await sql`
+	// `;
+};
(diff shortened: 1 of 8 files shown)


src/lib/db/migrations/00025-directory-data-space-community-ids.js
```diff
@@ -0,0 +1,26 @@
...
+			name: undefined, // remove `name` from directories
+		};
+		// eslint-disable-next-line no-await-in-loop
+		await sql`
+			UPDATE entities
+			SET data=${data}
+			WHERE entity_id=${space.directory_id}
+		`;
+	}
+};
```
ryanatkn, line 26: this seems inefficient, probably easy to do in a SQL one-liner? If so I'll try that first next time.

greatbacon: I think in a migration inefficient is fine

greatbacon (approved):

src/lib/vocab/entity/entityData.ts
```diff
@@ -11,6 +12,12 @@ export interface BaseEntityData {
 	checked?: boolean;
 }
 
+export interface DirectoryEntityData extends BaseEntityData {
+	type: 'Collection';
+	community_id: number;
```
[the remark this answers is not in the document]
ryanatkn: right on, got that implemented, so we can now narrow types to detect directories at runtime via `'space_id' in $entity.data`

PR #408, opened 2022-06-14
support key for opening contextmenu

ryanatkn:
kinda experimental, makes the spacebar and "Enter" open the contextmenu at the cursor -- not sure about those keys, or about the behavior in general, was just a random itch to scratch


greatbacon (approved):
I'm not sure when I'd use spacebar+enter to open the menu instead of just right clicking, but I don't think this hurts anything either lol

PR #409, opened 2022-06-22
Refactor update last seen

greatbacon:
This refactors the `UpdateLastSeen` event into `ClearFreshness` & extracts the logic into a mutation helper function, meaning we no longer have to pass "dispatch" around to access it from the helpers.


src/lib/ui/SpaceView.svelte
```diff
@@ -18,10 +18,7 @@
 	export let space: Readable<Space>;
 
 	$: directory_id = $space.directory_id;
-	$: dispatch.UpdateLastSeen({directory_id});
-	// const updateLastSeen = (directory_id: number) => {
-	// 	dispatch.UpdateLastSeen({directory_id});
-	// };
+	$: dispatch.ClearFreshness({directory_id});
```
ryanatkn, line 21: Not something to change in this PR, am wondering if we should call the new mutation helper from the `SelectSpace` mutation instead of doing that work here. Seems a bit better to have an event-oriented change in this case rather than a view-oriented one, but we may want `ClearFreshness` as a first-class UI event, and if this code were moved to `SelectSpace`, it would need an extra step to initialize it on page load. (because `SelectSpace` isn't called on startup)

ryanatkn (approved):

src/lib/ui/uiMutationHelper.ts
```diff
@@ -1,5 +1,8 @@
+import {browser} from '$app/env';
```
ryanatkn, line 1: might want to add an `s` at the end of this file, looks like the others have it

src/lib/ui/uiMutationHelper.ts
```diff
@@ -27,3 +31,34 @@ export const upsertCommunityFreshnessById = (ui: WritableUi, community_id: numbe
...
+	const timestamp = time ?? Date.now();
+
+	if (lastSeenByDirectoryId.has(directory_id)) {
+		lastSeenByDirectoryId.get(directory_id)!.set(timestamp);
+	} else {
+		lastSeenByDirectoryId.set(directory_id, writable(timestamp));
+	}
+
+	if (browser) {
+		localStorage.setItem(`${LAST_SEEN_KEY}${directory_id}`, `${timestamp}`);
```
ryanatkn, line 50: looks like it can be simplified a bit:

```suggestion
	time = Date.now(),
): void => {
	const {lastSeenByDirectoryId, entityById, spaceById} = ui;

	if (lastSeenByDirectoryId.has(directory_id)) {
		lastSeenByDirectoryId.get(directory_id)!.set(time);
	} else {
		lastSeenByDirectoryId.set(directory_id, writable(time));
	}

	if (browser) {
		localStorage.setItem(`${LAST_SEEN_KEY}${directory_id}`, `${time}`);
```

src/lib/ui/SpaceView.svelte
greatbacon: Yeah, good food for thought, I'll leave a TODO for future consideration

PR #410, opened 2022-06-23
First draft board

greatbacon:
This is a first pass at trying to implement a true `board` view type, i.e. a reddit-like with posts & nested comments.


src/lib/ui/BoardItem.svelte
```diff
@@ -67,6 +67,7 @@
 			await createEntity();
 		}
 	};
+	//TODO BLOCK add cursor: pointer style for clickable posts (and update Todo view too)
```
ryanatkn, line 70: thinking these should be links -- our space route uses a splat in `src/routes/[community]/[...space]` so its `index.svelte` could be changed to handle routes more complicated than a root space. Lots of unknowns here, this is just a first thought

ryanatkn (approved):

PR #411, opened 2022-06-23
Sketch signup space

greatbacon:
This is an initial sketch of a space used for signing up to join a community, based on design work by @trusttri

Currently I have done minimal styling, so I'd like to pair on doing a pass with that.

Edit: There's also some consideration around *how* this form should be accessed; standalone space w/ public publish, sub-component of home, unqiue per space like home, other???

There's also no connection to the backend. I think there are a couple of exciting things brought up from this design. 

1.) This is our 1st use case for a space wanting to reference an entity in another space, potentially even declaratively using Svast
2.) This is a potential beginning for building interesting entity ref chains (question <> answer <> discussion <> conclusion)

### DIFF
--- a/src/lib/ui/NewcomerSubmission.svelte
+++ b/src/lib/ui/NewcomerSubmission.svelte
@@ -32,16 +32,19 @@
 		if (!q1) {
 			errorMessage = 'please answer all questions';
 			q1El.focus();
+			status = 'initial';
 			return;
 		}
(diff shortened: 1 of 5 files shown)


greatbacon:
Rename to NewcomerSubmission & rollup into `Home` view w/ toggle button for now

ryanatkn (approved):

PR #412, opened 2022-06-23
shrink the member item persona icons

ryanatkn:
Persona icons in the community member list were a bit big, now they're smaller and spaced apart with padding.


greatbacon (approved):
🚢

PR #418, opened 2022-07-01
simplify ClearFreshness event

ryanatkn:
The `ClearFreshness` event previously used the `time` param but no longer does, so this removes it. It also ensures `lastSeen` is not overwritten with smaller values, which I believe handles some corner cases with network delays and clock differences between client and server. (though I think there's more to think through here...)


src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -23,7 +23,7 @@ export const updateEntity = (ui: WritableUi, $entity: Entity): Writable<Entity>
 	}
 
 	if (spaceSelection.get()?.get().directory_id === $entity.entity_id) {
-		updateLastSeen(ui, $entity.entity_id, $entity.updated!.getTime());
+		updateLastSeen(ui, $entity.entity_id, ($entity.updated || $entity.created).getTime());
```
ryanatkn, line 26: just being defensive

src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -23,13 +23,13 @@ export const updateEntity = (ui: WritableUi, $entity: Entity): Writable<Entity>
...
 		if (!freshnessByDirectoryId.get(entity_id)) {
-			setLastSeen(ui, entity_id, $entity.updated!.getTime());
+			setLastSeen(ui, entity_id, ($entity.updated || $entity.created).getTime());
 			setFreshnessByDirectoryId(ui, entity);
 		}
 		upsertFreshnessByCommunityId(ui, spaceById.get($entity.data.space_id)!.get().community_id);
 		// Is the directory's space selected? If so we don't want a notification.
 		if (entity_id === spaceSelection.get()?.get().directory_id) {
-			updateLastSeen(ui, entity_id, $entity.updated!.getTime());
+			updateLastSeen(ui, entity_id, ($entity.updated || $entity.created).getTime());
```
ryanatkn, line 32: Just being defensive here because `updateEntity` may be called with an entity that hasn't been updated. Made me realize `upsertEntity` is probably clearer like you did with `upsertFreshnessByCommunityId`.

src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -23,13 +23,13 @@ export const updateEntity = (ui: WritableUi, $entity: Entity): Writable<Entity>
 	// Handle directories.
 	if ('space_id' in $entity.data) {
 		if (!freshnessByDirectoryId.get(entity_id)) {
-			setLastSeen(ui, entity_id, $entity.updated!.getTime());
+			setLastSeen(ui, entity_id, ($entity.updated || $entity.created).getTime());
```
ryanatkn, line 26: Just being defensive here because `updateEntity` may be called with an entity that hasn't been updated. Made me realize `upsertEntity` is probably clearer like you did with `upsertFreshnessByCommunityId`.

src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -23,13 +23,13 @@ export const updateEntity = (ui: WritableUi, $entity: Entity): Writable<Entity>
...
 			setFreshnessByDirectoryId(ui, entity);
 		}
 		upsertFreshnessByCommunityId(ui, spaceById.get($entity.data.space_id)!.get().community_id);
 		// Is the directory's space selected? If so we don't want a notification.
 		if (entity_id === spaceSelection.get()?.get().directory_id) {
-			updateLastSeen(ui, entity_id, $entity.updated!.getTime());
+			updateLastSeen(ui, entity_id);
```
ryanatkn, line 32: I think this is correct?

greatbacon (approved):

PR #421, opened 2022-07-08
First pass on query & orphan cleanup

greatbacon:
This PR adds a query to find any orphaned entities & then makes a call to delete them when deleting other entities.

Additional work:
- [x] figure out the async in loop issue
- [x] add tests
- [x] check in on directories when space gets deleted
- [x] add all deleted entities to result set for client side event culling
- [x] add check to deleteEntitiesById to compare # input with # deleted, throw error if mismatch

### DIFF
--- a/src/lib/app/eventTypes.ts
+++ b/src/lib/app/eventTypes.ts
@@ -353,7 +353,9 @@ export type EraseEntitiesResponseResult = ApiResult<EraseEntitiesResponse>;
 export interface DeleteEntitiesParams {
 	entity_ids: number[];
 }
-export type DeleteEntitiesResponse = null;
+export interface DeleteEntitiesResponse {
+	removed_entity_ids: number[];
+}
 export type DeleteEntitiesResponseResult = ApiResult<DeleteEntitiesResponse>;
 
 export interface CreateTieParams {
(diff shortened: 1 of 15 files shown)


greatbacon:
All additional tasks are finished, so this is ready for review

greatbacon:
@ryanatkn did you have any more thoughts on this PR? I think I got everything we discussed earlier in the week.

ryanatkn (approved):

src/lib/vocab/entity/entityServices.test.ts
```diff
@@ -131,5 +134,40 @@ test_entityServices('assert default as max pageSize', async ({random}) => {
...
+				entity_ids: [list.entity_id],
+			},
+			...serviceRequest,
+		}),
+	);
+	//confirm all four entity ids come back
+	assert.ok(result.removed_entity_ids.includes(list.entity_id));
+	assert.ok(result.removed_entity_ids.includes(todo1.entity_id));
+	assert.ok(result.removed_entity_ids.includes(todo2.entity_id));
+	assert.ok(result.removed_entity_ids.includes(todo3.entity_id));
```
ryanatkn, line 169: For completion might be good to check the length of the array.

ryanatkn:
Looks great!

src/lib/app/eventTypes.ts
```diff
@@ -353,7 +355,9 @@ export type EraseEntitiesResponseResult = ApiResult<EraseEntitiesResponse>;
 export interface DeleteEntitiesParams {
 	entity_ids: number[];
```
ryanatkn, line 356: Talked about changing this to `entityIds` -- snake_case only if there's a 1:1 mapping to a column in the database

PR #431, opened 2022-07-21
Use localStore for managing luggage & marquee state

greatbacon:
This PR adds localStores around the luggage & marquee state stores in UI.

As an aside, I'm not sure if this is just a local dev thing, but there's some noticeable pop-in on the elements before collapse when saved as collapsed. Any thoughts?

### DIFF
--- a/src/lib/ui/ui.ts
+++ b/src/lib/ui/ui.ts
@@ -19,7 +19,7 @@ import type {Membership} from '$lib/vocab/membership/membership';
 import {createContextmenuStore, type ContextmenuStore} from '$lib/ui/contextmenu/contextmenu';
 import {initBrowser} from '$lib/ui/init';
 import {isHomeSpace} from '$lib/vocab/space/spaceHelpers';
-import {locallyStoredMap} from '$lib/ui/locallyStored';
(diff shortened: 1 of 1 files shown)


ryanatkn:
Oof yeah, it's because the SSR'd HTML has no access to the setting, so it's just using the initial value, and then when the client-side JS hydrates, it reads from storage, sets the value, and then updates the UI accordingly.

I don't see a good fix. I'm good to merge as-is if you are. Some options for later:

- do nothing and accept the kinda janky UX (sigh)
- move the setting to the account, so the SSR'd version is rendered correctly. We'll definitely do this for notifications. But it seems like the better UX is to have the setting be per-device, not per-account, and storing per-device info on the server seems difficult, complicated, and has implications for privacy.
- try to paper over the pop-in effect on the client, maybe hiding stuff with CSS until hydrated -- this too seems bad because we'd lose the benefits of SSR'd HTML

ryanatkn (approved):

greatbacon:
Alright, I'm okay with jank. If we get tired off it or get user complaints, we can figure something else out.

PR #433, opened 2022-07-21
extract Repos class

ryanatkn:
First step towards db transactions, this extracts a `Repos` class (would `DatabaseRepos` be a better name?) that can support both transactions and non-transactional usage. Individual repos no longer has the `db` instance reference, instead they have `sql` and the `repos` references. This properly scopes each repo to just accessing data, not the entire `db` instance, which allows us to have the same interface for both transactions and non-transactional `Repos` instances. (transactions won't have the whole `db`, just `sql` and `repos`)

To implement transactions in a followup PR, we'll lazily create a new `Repos` instance for requests that use the database with the transactional version of a `sql` instance. This means the underlying code in each repo should be able to stay the same.


greatbacon (approved):
This lgtm, I think we've established that `Repo` implies DB work, but if we find it's getting confusing down the road we can clarify it as `DatabaseRepo`

PR #435, opened 2022-07-22
implement transactions

ryanatkn:
- unfortunately I think we're forced to indent every service that uses a transaction, because we have to nest the code in the transaction promise callback -- all of that code is exactly the same as before minus the `transact` wrapper, just indented, except for the one spot I call out in the PR comments
- composition should just work, but there's currently no way to start a second transaction or set a savepoint - I left TODOs for that
- concurrent reads in `DeleteMembership` are failing -- I failed to reproduce in the driver so I can't report a bug. See the PR comment.


src/lib/vocab/membership/membershipServices.ts
```diff
@@ -3,87 +3,92 @@ import {Logger} from '@feltcoop/felt/util/log.js';
...
+				persona_id,
+				community_id,
+			);
+			// TODO why can't this be parallelized? bug in our code? or the driver? failed to reproduce in the driver.
+			// const [personaResult, communityResult] = await Promise.all([
+			// 	repos.persona.findById(persona_id),
+			// 	repos.community.findById(community_id),
+			// ]);
+			const personaResult = await repos.persona.findById(persona_id);
+			const communityResult = await repos.community.findById(community_id);
```
ryanatkn, line 63: This is the concurrency problem mentioned in the body of the PR

greatbacon (approved):

PR #436, opened 2022-07-26
setup db in deploy script

ryanatkn:
work in progress -- automates db creation during the initial deploy script, one of the last steps that's still manual


greatbacon:
So I think the issue was that we were trying to create the `postgres` user which already existed, so I just changed things to update the pw on it instead.  Did a test on a fresh droplet & everything seemed to work! 🎉 

Next on the automation list: automating migrations & the initial pm2 launch

src/infra/setup.task.ts
```diff
@@ -23,11 +23,12 @@ export const task: Task<SetupTaskArgs> = {
 		const EMAIL_ADDRESS = fromEnv('EMAIL_ADDRESS');
 		// TODO this is hacky because of `import.meta` env handling
 		const {API_SERVER_HOST_PROD, SVELTEKIT_SERVER_HOST} = await import('../lib/config.js');
+		const NODE_VERSION = 18.06;
```
ryanatkn, line 26:
```suggestion
		const NODE_VERSION = '18';
```

The version we're on is `"18.6.0"`, don't think the number `18.06` would work as expected. I also think it's better to just pin the major version, so security updates for new instances are automatic.

ryanatkn:
hurray!! Good find. I have one small change, and also need your approval for the PR since GitHub won't let me approve my own.

greatbacon (approved):
lol approving my own work

PR #437, opened 2022-07-26
add admin community

ryanatkn:
Creates the default `admin` community, whose members get superuser privileges over the instance.

questions

todos

- [x] write a docs page
- [x] when the first persona is created, create the `admin` community with them as a user 
- [x] disallow creating personas with name "admin"
- [x] disallow orphaning admin community (last persona cannot be removed)
- [x] disallow deleting admin community
- [x] add a migration to add the admin community and initial member (with `community_id=1`)

benefits:

- lets instance owners manage admins with the same community tools as everything else
- admin persona is a builtin concept for e.g. broadcasting messages instance-wide

followup:

- disallow joining the admin community unless the persona is an admin (implement `repos.personas.isAdmin(persona_id)`)
- disallow deleting admin personas -- need to remove them as admin first (need to implement `DeletePersona` first)
- restrict `ReadCommunities` to admins (and think about a proper API)
- test the constraints (but how with the admin community? need the particular account/persona -- maybe time to cleanup the db as we test?)
- when adding `DeletePersona`, change `entity.persona_id` to a foreign key
- if admin, show edit and field in Edit Entity in Devmode


src/lib/db/seed.ts
```diff
@@ -59,7 +59,7 @@ export const seed = async (db: Database): Promise<void> => {
 				}),
 			);
 
-			log.trace('created persona', persona);
+			log.trace('created persona!!!!!!!!!!!!!!!!!!!!!!', persona);
```
greatbacon, line 62: did you want to keep these `!`s?

src/lib/ui/Marquee.svelte
```diff
@@ -35,6 +35,7 @@
 	{#if $devmode}
 		<section>
 			<menu>
+				<li><a href="/admin">/admin</a></li>
```
greatbacon, line 38: Should we add a membership for the admin community here?

src/lib/db/migrations/00027-admin-community.js
```diff
@@ -0,0 +1,50 @@
...
+		return;
+	}
+
+	// At this point, we know we need to migrate existing data to the new assumptions.
+	// The "admin" community and its persona are expected to have id 1.
+
+	// Move the existing community -- all references to `communities.community_id` cascade.
+	await sql`
+		UPDATE communities
+		SET community_id = (1 + (SELECT count(*) from communities))
```
greatbacon, line 15:
```suggestion
		SET community_id = (1 + (SELECT max(community_id) from communities))
```
Ran into a bug where I had deleted communities so there were already communities at the count+1

src/lib/db/migrations/00027-admin-community.js
```diff
@@ -0,0 +1,50 @@
...
+		WHERE community_id = 1;
+	`;
+	// Move the existing persona -- all references to `personas.persona_id` cascade.
+	const [{persona_id}] = await sql`
+		UPDATE personas
+		SET persona_id = (1 + (SELECT count(*) from personas))
```
greatbacon, line 21:
```suggestion
		SET persona_id = (1 + (SELECT max(persona_id) from personas))
```
same issue here

src/lib/db/migrations/00027-admin-community.js
ryanatkn: nice catch

src/lib/db/seed.ts
ryanatkn: lol oops

src/lib/db/migrations/00027-admin-community.js
```diff
@@ -0,0 +1,50 @@
...
+	await sql`
+		INSERT INTO personas (persona_id, type, name, community_id) VALUES (
+			1, 'community', 'admin', 1
+		);
+	`;
+	// Create the membership for the persona that formerly had `persona_id` 1.
+	await sql`
+		INSERT INTO memberships (persona_id, community_id) VALUES (
+			${persona_id}, 1
+		);
```
greatbacon, line 48: I think we also need a default space for the admin community here.

src/lib/db/migrations/00027-admin-community.js
```diff
@@ -0,0 +1,62 @@
...
+	// Defer the `membership.persona_id` foreign key constraint to the end of the transaction:
+	await sql`
+		ALTER TABLE memberships
+		DROP CONSTRAINT memberships_persona_id_fkey,
+		ADD CONSTRAINT memberships_persona_id_fkey
+			FOREIGN KEY (persona_id)
+			REFERENCES personas (persona_id)
+			ON UPDATE CASCADE ON DELETE CASCADE
+			INITIALLY DEFERRED;
+	`;
```
ryanatkn, line 12: Seed was failing when this PR was combined with the transaction PR. The added bit is `INITIALLY DEFERRED` -- it allows the constraint to be checked only at the end of the transaction. More here: https://www.postgresql.org/docs/current/sql-createtable.html

ryanatkn: Hmm adding this makes the membership silently fail -- I think if seed ran in a single transaction, it would have caught this error.

I simply removed the constraint and the membership succeeds. Going to poke around a bit...

src/lib/db/migrations/00027-admin-community.js
ryanatkn: we discussed, it's easy enough to create one inside the app

src/lib/db/migrations/00027-admin-community.js
ryanatkn: It's now seeding with a single transaction, so if anything goes wrong nothing gets seeded, doesn't solve our problem though.

ryanatkn: Figured out the problem! `serviceRequest.repos` isn't using the transacting `repos`

# 🤦

src/lib/db/migrations/00027-admin-community.js
```diff
@@ -0,0 +1,50 @@
...
+			1, 'community', 'admin', 1
+		);
+	`;
+	// Create the membership for the persona that formerly had `persona_id` 1.
+	await sql`
+		INSERT INTO memberships (persona_id, community_id) VALUES (
+			${persona_id}, 1
+		);
+	`;
+};
```
ryanatkn, line 50: Creating the initial space is something we can do in the UI I believe, so I'm not adding the migration. (being lazy, because we'd also have to create the directory entity)

src/lib/ui/SpaceInput.svelte
```diff
@@ -16,7 +16,10 @@
 	import {toSpaceUrl} from '$lib/ui/url';
 
 	// TODO does this belong in `view`?
-	const creatableViewTemplates = viewTemplates.filter((v) => v.creatable !== false);
+	const admin = false; // TODO implement
```
ryanatkn, line 19: mark this for a followup

src/lib/db/migrations/00027-admin-community.js
greatbacon: Ah, yup, that'll do it. Great job figuring that out though Ryan!

greatbacon (approved):
Yay, you found the bug! Seems to me like between this bug & some of the work I'm going to be doing related to governance, we may have a rolling refactor on the horizon for splitting core biz logic into `Helpers`.

Otherwise I tested this locally and it lgtm!

PR #441, opened 2022-08-02
improve local storage types

ryanatkn:
hmmm seems to work? I don't quite trust it (bc I don't fully understand), and it can probably be improved, but ... it typechecks and maintains types for existing usage I think. I think the `locallyStoredMap` signature remains unwieldy for custom json types but we don't have any of those yet.


src/lib/ui/ui.ts
```diff
@@ -197,11 +197,10 @@ export const toUi = (
 				: null,
 	);
 	// TODO consider making this the space store so we don't have to chase id references
-	const spaceIdSelectionByCommunityId = locallyStoredMap<
-		Mutable<Map<number, number | null>>,
-		Map<number, number | null>,
-		Array<[number, number | null]>
-	>(mutable(new Map()), 'spaceIdSelectionByCommunityId');
+	const spaceIdSelectionByCommunityId = locallyStoredMap(
+		mutable(new Map<number, number | null>()),
```
ryanatkn, line 201: Notice we do need to annotate the inner map, but that makes sense, it'd have no way to understand it otherwise. The cool part is it infers the value of the json from the map type.

greatbacon (approved):

PR #446, opened 2022-08-11
Governance Proof Of Concept

greatbacon:
This is going to be a long lived PR most likely, so I'm going to be using this space to kind of keep track of work & open questions as I work through them. There might be a lot of offline convos, so apologies to any future archeologists looking at this.

My goal is to try & work through 4 stages of complexity to get to a "final" proof of concept state, working out APIs, DB tables & models, & any additional technologies in the stack as I go.

~~1. Hardcode simple eval logic into the service~~
~~2. Add a policy table & some sort of rules engine to replace the hardcoded check~~
3. Add roles & use those in a policy
4. maybe begin expanding policies to other services

Current phase: 3
Open Questions:

- [] Add Roles stack from table to service
- [] pair on Policy UX


ryanatkn:
> I'm initially setting policy evaluation as a service, but it's not quite a full blown service, is this okay?

I'm thinking that policy modeling should be implemented as helpers, basically generic code that's easy to test and call from any service. Related helpers/classes/types/etc might belong in `$lib/vocab/policy` starting with `$lib/vocab/policy/policy.ts`? I'm thinking `Policy` is a new vocab term.

ryanatkn:
Instead of `evaluation` what else could we name it? The JS dev in me got startled when I saw "eval" lol

a verb, `the server ___ policies`

- checks
- applies
- resolves
- verifies
- confirms
- passes

greatbacon:
From convo regarding personas: If schema has `authorized:true` on it, make sure it also has an `actorId` attribute, and pre-evaluate the account has access to the actor persona.

src/lib/server/authorize.ts
```diff
@@ -2,19 +2,39 @@ import {OK, type Result} from '@feltcoop/felt';
...
-	// TODO add authorization logic based on the account's roles and the requested service
+
+	//TODO hack to flag only a subset of events for policy checking
+	if (!ACTIVE_POLICY_SERVICES.includes(service.event.name)) return OK;
+
+	//TODO should params be schema validated prior to authorization? Or as part of auth?
+	const actor = params.actor;
+	if (!actor) {
+		return {ok: false, message: 'authorized events need an actor id'}; // TODO centralize error message strings
+	}
```
ryanatkn, line 32: Validating first makes sense, fixed in #458

PR #447, opened 2022-08-17
add client admin detection

ryanatkn:
The goal of this PR is to let the client efficiently ask "is this persona an admin?" There are several ways to do this and I left some TODOs for optimization -- basically, the upstream data structures make this less efficient than I'd like, but the code added for this PR seems fine.


src/lib/ui/ui.ts
```diff
@@ -68,6 +69,7 @@ export interface Ui {
 	lastSeenByDirectoryId: Map<number, Writable<number> | null>;
 	freshnessByDirectoryId: Map<number, Readable<boolean>>;
 	freshnessByCommunityId: Map<number, Writable<boolean>>;
+	adminPersonas: Readable<Set<Readable<Persona>>>;
```
ryanatkn, line 72: An alternative would be to store the `persona_id`, not the store reference. This question keeps arising and we answer it differently in different places.

src/lib/ui/ui.ts
```diff
@@ -221,6 +223,13 @@ export const toUi = (
 	const freshnessByDirectoryId: Map<number, Readable<boolean>> = new Map();
 	const freshnessByCommunityId: Map<number, Writable<boolean>> = new Map();
 
+	// TODO optimization: ideally this would recalculate only when the admin community's personas change, not when any membership changes
+	// TODO consider making the value of `personasByCommunityId` a set instead of array, then this could be simplified
+	const adminPersonas = derived(
+		[personasByCommunityId],
+		([$personasByCommunityId]) => new Set($personasByCommunityId.get(ADMIN_COMMUNITY_ID)),
+	);
```
ryanatkn, line 231: Wasn't sure where to place `adminPersonas` in the 3 lists -- below int he return value, here in the body of the `toUi` function, and above in the interface declaration. The current organization isn't consistent and could use some better descriptive buckets.

greatbacon (approved):

PR #454, opened 2022-08-18
fix redirect that should be a link

ryanatkn:
- [x] ensure `SelectPersona`, `SelectCommunity`, and `SelectSpace` are used only when syncing the url
- [ ] should each of `SelectPersona`, `SelectCommunity`, and `SelectSpace` be changed to no longer be events?


src/lib/ui/CommunityNavItem.svelte
```diff
@@ -48,7 +48,6 @@
 	class:persona={isPersonaHomeCommunity}
 	style="--hue: {$community.settings.hue}"
 	use:contextmenu.action={[[CommunityContextmenu, {community, persona}]]}
-	on:click={() => dispatch.SelectPersona({persona_id: $persona.persona_id})}
```
ryanatkn, line 51: this was redundant because it's a link

src/lib/app/contextmenu/AppContextmenu.svelte
```diff
@@ -20,25 +21,8 @@
...
-				<li class="menu-item panel-inset" role="none">
-					<div class="content">
-						<div class="icon"><PersonaAvatar persona={sessionPersona} showName={false} /></div>
-						<div class="title"><PersonaAvatar persona={sessionPersona} showIcon={false} /></div>
-					</div>
-				</li>
-			{:else}
-				<!-- TODO support store param? only? -->
-				<ContextmenuEntry
-					action={() => dispatch.SelectPersona({persona_id: sessionPersona.get().persona_id})}
```
ryanatkn, line 34: This was a bug because it didn't update the URL. The extracted component now does it correctly.

greatbacon (approved):

PR #456, opened 2022-08-22
add ephemera

ryanatkn:
- add the `Ephemera` event, with `broadcast: true` (also includes `params.actor` because I know we'll need it and I don't see a downside of including it now)
- add the `EphemeraService`, which echos back the params
- add the Ephemera mutation which updates `ui.ephemera`, a writable store that holds only the most recent value (store subscribers are synchronously updated, so this should work)
- log the value of `$ephemera` in the iframe view to mark the next steps to finish the integration in another PR -- you can test the whole chain by running this in your console in a space (id "3" in this example) with an `Iframe` view: `dispatch.Ephemera({space_id: 3, actor: 1, data: {type: 'a'}})`

future:

- forward ephemera to/from iframes in the `Iframe` view: #457
- use ephemera for cursor positions, typing indicators, and other presence effects
- broadcasting is currently coupled to the response value to the requesting user, which is slightly wasteful in this case because the user that creates the ephemera doesn't need the value echoed back to them, but doesn't seem worth sweating any time soon


src/lib/ui/uiEvents.ts
```diff
@@ -18,6 +18,41 @@ export const Ping: ServiceEventInfo = {
...
+		$id: '/schemas/EphemeraResponse.json',
+		type: 'object',
+		properties: {
+			actor: {type: 'number'},
+			space_id: {type: 'number'},
+			key: {type: 'string'},
+			value: {type: 'object'},
+		},
+		required: ['actor', 'space_id', 'key', 'value'],
+		additionalProperties: false,
```
ryanatkn, line 47: This is copypasta between the params and response, might be worth pulling out a separate object/schema like `EphemeralData` (or just `Ephemera`?), or even just declaring it as a variable here in this module and using it twice.

src/lib/server/uiServices.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const PingService: ServiceByName['Ping'] = {
+	event: Ping,
+	perform: async () => ({ok: true, status: 200, value: null}),
+};
+
+export const EphemeraService: ServiceByName['Ephemera'] = {
+	event: Ephemera,
+	perform: async ({params}) => ({ok: true, status: 200, value: params}),
+};
```
ryanatkn, line 12: this module was previously named `pingService.ts` with a TODO to move it, `uiServices` seems serviceable

src/lib/app/eventTypes.ts
```diff
@@ -396,6 +400,22 @@ export type PingParams = null;
 export type PingResponse = null;
 export type PingResponseResult = ApiResult<PingResponse>;
 
+export interface EphemeraParams {
+	actor: number;
+	space_id: number;
+	data: {
```
ryanatkn, line 406: Wasn't sure if this should be called `data`, `value`, or something else?

src/lib/app/eventTypes.ts
```diff
@@ -396,6 +400,22 @@ export type PingParams = null;
...
+		[k: string]: unknown;
+	};
```
ryanatkn, line 408: Wondering if we should enforce a `type` string property on `data`, or just let implementations do whatever. I made it an object because a plain string/number/boolean wouldn't have any context so it seems nonsensical.

greatbacon: I think this is fine as is, we can wait for more use cases to bring things to light.

greatbacon (approved):

PR #457, opened 2022-08-22
enable ephemera in iframes

ryanatkn:
extends #456

- [x] proof of concept -- see also https://github.com/feltcoop/dealt/pull/45
- [x] finish and refine the API and upstream to Felt: https://github.com/feltcoop/felt/pull/235
- [x] rename `guest` to `tenant` to avoid collisions with website guests with `session.guest`
- [x] finalize the data sent to the iframe (nothing needed! we do send the `space_id` as the `key`, and we can expand it with user permission with things like `persona.hue`, `persona.name`, etc)
- [x] add `Ephemera.data.type` to the schema because it's expected by the Host API and it's a standard way to differentiate data that we can leverage in various ways

future:

- think about querying persona names and additional context (security/privacy being the issue)


src/lib/ui/uiEvents.ts
```diff
@@ -28,7 +28,7 @@ export const Ephemera: ServiceEventInfo = {
 		properties: {
 			actor: {type: 'number'},
 			space_id: {type: 'number'},
-			data: {type: 'object'},
+			data: {type: 'object', properties: {type: {type: 'string'}}, required: ['type']},
```
ryanatkn, line 31: We need `type` to set up the initial connection -- the host listens for `{type: 'felt.connect'}` and responds with `{type: 'felt.connected', key}`.

Data needs *some* way to be differentiated -- duck typing seems wrong for ephemera, because the possibilities are endless -- so `type` seems solid. (could also be another name like `key` or `kind`) Having a standard type also allows us to do logging and various forms of processing generically in felt-server that would otherwise be impossible.

src/lib/ui/view/Iframe.svelte
```diff
@@ -21,16 +34,23 @@
...
-		title={$space.name}
-		{src}
-		on:load={() => (loaded = true)}
-	/>
+	<iframe frameborder="0" title={$space.name} {src} on:load={() => (loaded = true)} />
 	{#if !loaded}
 		<PendingAnimationOverlay />
 	{/if}
+	<FeltWindowHost
+		key={$space.space_id}
```
ryanatkn, line 42: This is the one bit of info we currently leak to iframes, but it seems completely benign

ryanatkn: Never mind, instead of leaking this info we'll make it optional, see https://github.com/feltcoop/felt/pull/236

PR #460, opened 2022-08-24
fix CreateMembership to query the new community data

ryanatkn:
Fixes the `CreateMembership` mutation, adding `broadcast: true` to the event and then ensuring that the client queries the needed data. Only the newly added member needs this data, so it's not added to the `CreateMembership` response.

As I pulled on this thread, I kept discovering more bugs, all to get the memberships reactively updating. One oversight that's fixed is returning directories along with spaces. There are some subtle changes to some events for that.

followup:

- fix the `DeleteMembership` mutation
- proactively send the new community info to the user when they're added (because broadcasting won't work after we scope for security)


src/lib/vocab/membership/membershipMutations.ts
```diff
@@ -21,7 +39,7 @@ export const DeleteMembership: Mutations['DeleteMembership'] = async ({
 }) => {
 	const result = await invoke();
 	if (!result.ok) return result;
-	// TODO also update `communities.personas`
+	// TODO also update `communities` to remove the community unless another persona has a membership
```
ryanatkn, line 42: This mutation likewise needs some work, will do in a followup.

src/lib/vocab/community/communityEvents.ts
```diff
@@ -50,8 +50,11 @@ export const ReadCommunity: ServiceEventInfo = {
 		type: 'object',
 		properties: {
 			community: {$ref: '/schemas/Community.json', tsType: 'Community'},
+			spaces: {type: 'array', items: {$ref: '/schemas/Space.json', tsType: 'Space'}},
+			memberships: {type: 'array', items: {$ref: '/schemas/Membership.json', tsType: 'Membership'}},
+			communityPersona: {$ref: '/schemas/Persona.json', tsType: 'Persona'},
```
ryanatkn, line 55: I think it makes sense to expand `ReadCommunity` to include all of the related data like this. It's the same as the return value from `CreateCommunity`.

src/lib/vocab/membership/membershipMutations.ts
```diff
@@ -1,15 +1,33 @@
...
-}) => {
+export const CreateMembership: Mutations['CreateMembership'] = async ({invoke, dispatch, ui}) => {
+	const {memberships, communityById} = ui;
 	const result = await invoke();
 	if (!result.ok) return result;
 	const {membership: $membership} = result.value;
-	// TODO also update `communities.personas`
+	const {community_id} = $membership;
+
+	// If there's no community locally, we were just added to it, so query its data in full.
```
ryanatkn, line 14: After going this direction, I'm realizing that changing `CreateMembership` to broadcast won't work for the added persona after we scope broadcasting by community. Instead I think the server will have to proactively send new data like this over websockets?

ryanatkn:
Found the bug here, forgot a `!` lol -- https://github.com/feltcoop/felt-server/pull/460/commits/92a2adef766c1d329c97a529f4e2b5eff5037a83

greatbacon (approved):
Thanks for going on this bug hunt!

PR #464, opened 2022-08-30
fix delete membership mutation

ryanatkn:
The delete membership mutation had broken logic, and it also needs to delete the community and related data, if no other personas are joined. Given the overlap with `DeleteCommunity`, I extracted the `deleteCommunity` mutation helper which is then called in both places. There are some TODOs for cleaning up orphaned objects in the helper, but I don't think they'll cause problems right now.


src/lib/vocab/community/communityMutationHelpers.ts
```diff
@@ -44,3 +46,43 @@ export const upsertCommunity = (
...
+	});
+
+	memberships.swap(memberships.get().value.filter(($m) => $m.get().community_id !== community_id));
+
+	// TODO delete all spaces
+
+	// TODO delete all entities
+
+	// TODO delete the community persona and all personas who were members
+	// but are no longer members of any other communities
```
greatbacon, line 87: Do we need it to do the second step here? Since personas always have a home community which cannot be deleted.

greatbacon (approved):

src/lib/vocab/community/communityMutationHelpers.ts
ryanatkn: Good catch, I updated the TODO, instead of `and all personas who were members` I meant `and all non-session personas who were members`

PR #465, opened 2022-08-30
return directories with spaces

ryanatkn:
Fixes the remaining events that return spaces but no directories. ~~Got a question on one of them.~~


greatbacon:
LGTM, is there still a pending question somewhere though?

ryanatkn:
Oh nope, the question was whether `UpdateSpace` needed to return directories as well, and I think the answer is no. The directory doesn't change in that case, and if you're receiving that event you should have the space loaded in the UI already.

greatbacon (approved):

PR #466, opened 2022-09-11
fix membership cleanup on delete community

ryanatkn:
When deleting a community, memberships and personas should now be correctly cleaned up on the client.

Adds two new mutation helpers, `deleteMemberships` which calls `deletePersonas`


src/lib/vocab/membership/membershipMutationHelpers.ts
```diff
@@ -0,0 +1,31 @@
...
+			// Check if we need to delete the persona along with the membership.
+			const {persona_id} = membership.get();
+			if (!$memberships.some((m) => m.get().persona_id === persona_id && m !== membership)) {
+				personasToRemove.add(personaById.get(persona_id)!);
+			}
+			return false;
+		}),
+	);
+
+	deletePersonas(ui, personasToRemove);
```
ryanatkn, line 30: This should be all personas that no longer have memberships, including the community personas and non-session personas that aren't a member of any other loaded community.

src/lib/ui/ui.ts
```diff
@@ -57,25 +57,15 @@ export interface Ui {
 	personaById: Map<number, Readable<Persona>>;
 	communityById: Map<number, Readable<Community>>;
 	spaceById: Map<number, Readable<Space>>;
+	entityById: Map<number, Readable<Entity>>; // TODO mutable inner store
+	// derived state
```
ryanatkn, line 61: This file has no substantial changes, just rearranging things to be organized correctly.

src/lib/vocab/community/communityMutationHelpers.ts
```diff
@@ -62,7 +63,7 @@ export const deleteCommunity = (ui: WritableUi, community_id: number): void => {
 
 	if (communitySelection.get() === community) {
 		const persona = personaSelection.get()!;
-		void goto('/' + persona.get().name + location.search, {replaceState: true});
+		await goto('/' + persona.get().name + location.search, {replaceState: true});
```
ryanatkn, line 66: This needs to be awaited or else the view code throws an error, because we're synchronously deleting all of the objects from the ui state.

greatbacon (approved):

PR #468, opened 2022-09-11
cleanup entities on delete community and batch entity/tie mutations

ryanatkn:
Improves the entity/tie mutation helpers to be batched, consolidating their helpers, and changes `DeleteSpace`/`DeleteEntities` to no longer return the deleted entity ids, because it doesn't scale and the client has all of the information it needs to clean things up via following ties.

It doesn't do the work to cleanup fully after deleting entities yet, because the PR is big enough as is, but I marked those todos for a followup.

Also renames to stash/evict from upsert/delete.

followup:

- evict orphaned entities
- address some todos for efficiency


src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -1,4 +1,5 @@
-import {writable, mutable, type Writable} from '@feltcoop/svelte-gettable-stores';
```
ryanatkn, line 1: The changes in this file look like a lot, but it's mostly moving existing code inside loops with batched semantics. There's a new pattern using a `mutated` local to track the mutable stores that'll be changed at the end of the mutation helper.

greatbacon (approved):

PR #469, opened 2022-09-12
add TextInput component

ryanatkn:
Adds `ui/TextInput.svelte` and uses it in the view components. It includes the persona icon to the left of the input.


src/lib/ui/TextInput.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+	import {createEventDispatcher} from 'svelte';
+	import type {Readable} from '@feltcoop/svelte-gettable-stores';
+
+	import PersonaAvatar from '$lib/ui/PersonaAvatar.svelte';
+	import type {Persona} from '$lib/vocab/persona/persona';
+
+	const dispatcher = createEventDispatcher<{submit: string}>();
```
ryanatkn, line 8: Normally I'd call this `dispatch` but that conflicts with our app's version. There's also `dispatchEvent` that's a DOM global, so `dispatcher` was the next best thing I could think of.

src/lib/ui/EntityInput.svelte
```diff
@@ -69,8 +76,16 @@
...
+<div class="entity-input markup">
 	<h1>New {entityName}</h1>
+	<section class="row">
+		<span class="spaced">in</span>
+		<CommunityAvatar {community} />
+	</section>
+	<section class="row">
+		<span class="spaced">as</span>
+		<PersonaAvatar {persona} />
+	</section>
```
ryanatkn, line 88: Adds contextual info so you can see where you're making the entity and as whom. In a followup PR I'm going to extract this into a reusable pattern because it keeps appearing.

greatbacon (approved):

PR #475, opened 2022-09-16
authorize actor

ryanatkn:
Adds generic authorization logic to the http/websocket middlewares. If `params.actor` exists, it now ensures the authenticated `account_id` matches `persona.account_id`, and returns a helpful error message if something goes wrong.


src/lib/server/authorize.ts
```diff
@@ -2,19 +2,35 @@ import {OK, type Result} from '@feltcoop/felt';
...
+	// Params validation ensures that omitting `actor` is caught ahead of this function call.
+	if (params.actor !== undefined) {
+		const personaResult = await repos.persona.findById(params.actor);
+		if (!personaResult.ok) {
+			return {ok: false, status: 400, message: 'actor cannot be found'};
+		}
+		if (personaResult.value.account_id !== account_id) {
+			return {ok: false, status: 401, message: 'actor is not authorized for this account'};
+		}
+		// TODO BLOCK return the persona object and forward with the request?
```
ryanatkn, line 32: Thinking it'd be wise to return the `actor` `Persona` object here and forward it with the request, just like we have the `account_id` always attached.

src/lib/server/authorize.ts
```diff
@@ -2,19 +2,41 @@ import {OK, type Result} from '@feltcoop/felt';
...
+		if (personaResult.value.account_id !== account_id) {
+			return {ok: false, status: 403, message: 'actor is not authorized for this account'};
+		}
+		// TODO BLOCK return the persona object and forward with the request?
+		// We could also use the same pattern to fix the problem where
+		// `account_id` is available for some but not all service requests.
+		// Maybe three types of service requests:
+		// `AuthorizedServiceRequest`	- yes `account_id`, yes `persona`/`params.actor`
+		// `UnauthorizedServiceRequest` - yes `account_id`, no `persona`/`params.actor`
+		// `UnathenticatedServiceRequest` - no `account_id`, no `persona`/`params.actor`
```
ryanatkn, line 38: Thinking it'd be wise to return the `actor` `Persona` object here and forward it with the request, just like we have the `account_id` always attached when available. We already had a problem where `account_id` isn't present for the _unauthenticated_ event requests, and now we'll have a similar problem with `actor` for _unauthorized_ event requests.

(there may be better words than unauthenticated/unauthorized here, the point is they don't need authentication or authorization)

greatbacon (approved):

PR #478, opened 2022-09-19
Add role backend services down to DB

greatbacon:
This PR adds the initial set of logic for creating, reading, updating, & deleting Roles from the Service layer down to the DB (technically it also adds the relation table for personas too but that work is for a later PR)


src/lib/db/migrations/00029-create-roles.js
```diff
@@ -0,0 +1,19 @@
+/** @param {import('postgres').Sql<any>} sql */
+export const up = async (sql) => {
+	await sql`
+	CREATE TABLE IF NOT EXISTS roles (
+		role_id serial primary key,
+		community_id int REFERENCES communities (community_id) ON UPDATE CASCADE ON DELETE CASCADE,
+		title text,
```
ryanatkn, line 7: Should be unique right? Probably via an index, not a unique constraint? That way checking collisions on create can be fast with a good error message.

Thinking the field should be lowercased, or possibly using the `LOWER()` trick we do with persona/account names.

Also what do you think of renaming this from `title` to `name`? We haven't used `title` anywhere yet, and we've been using `name` as the unique string identifier for objects. I think `title` makes more sense when we need a displayed value that differs from the name.

src/lib/vocab/role/roleEvents.ts
```diff
@@ -0,0 +1,112 @@
+import type {ServiceEventInfo} from '$lib/vocab/event/event';
+
+export const CreateCommunityRole: ServiceEventInfo = {
+	type: 'ServiceEvent',
+	name: 'CreateCommunityRole',
```
ryanatkn, line 5: Looks like the `Community` prefix is unnecessary and used only in some places, should be a simple global replace to fix unless there's a reason for it?

src/lib/vocab/role/roleEvents.ts
```diff
@@ -0,0 +1,112 @@
...
+};
+
+export const ReadCommunityRoles: ServiceEventInfo = {
+	type: 'ServiceEvent',
+	name: 'ReadCommunityRoles',
+	params: {
+		$id: '/schemas/ReadCommunityRolesParams.json',
+		type: 'object',
+		properties: {
+			community_id: {type: 'number'},
```
ryanatkn, line 39: The delete schema has `actor` but the other events are missing it.

src/lib/vocab/role/roleHelpers.ts
```diff
@@ -0,0 +1,62 @@
+import {blue, gray} from 'kleur/colors';
+import {Logger} from '@feltcoop/felt/util/log.js';
+import type {ErrorResponse} from '$lib/util/error';
+import type {Result} from '@feltcoop/felt';
+import type {Repos} from '$lib/db/Repos';
+import type {Role} from '$lib/vocab/role/role';
+
+const log = new Logger(gray('[') + blue('roleHelpers') + gray(']'));
+
+export const createCommunityRole = async (
```
ryanatkn, line 10: Do we expect these helpers to be called from multiple places where it'd be inconvenient to call them as services? If not it'd be simpler to inline this logic in the services and include the status codes.

ryanatkn (changes requested):

src/lib/vocab/role/RoleRepo.ts
```diff
@@ -0,0 +1,49 @@
...
+		if (!result.count) return NOT_OK;
+		return {ok: true, value: result[0]};
+	}
+
+	async deleteByIds(roleIds: number[]): Promise<Result<object>> {
+		log.trace('[deleteByIds]', roleIds);
+		const result = await this.sql<any[]>`
+			DELETE FROM roles WHERE role_id IN ${this.sql(roleIds)}
+		`;
+		if (!result.count) return NOT_OK;
```
ryanatkn, line 45:
```suggestion
```

src/lib/vocab/role/roleHelpers.ts
greatbacon: I started doing this as a future proof mechanism, as calling services from other services feels clunky & I've been enjoying having the "core business" logic separate from the "service response" logic.

ryanatkn: Agree on the clunkiness, it's nice to just pass `repos` and inherit the transaction without the `serviceRequest` construction, which is verbose and more error prone.

How about we add the `.server` infix, giving us one of:

- `$lib/vocab/role/roleHelpers.server.ts`
- `$lib/vocab/role/roleServiceHelpers.server.ts`

greatbacon: Fair enough, I'll roll up the helper functions to the services then.

ryanatkn (approved):
# 💯 

got one benign fix on a response schema

src/lib/vocab/role/roleEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+		additionalProperties: false,
+	},
+	response: {
+		$id: '/schemas/UpdateRoleResponse.json',
+		type: 'object',
+		properties: {
+			actor: {type: 'number'},
+			role: {$ref: '/schemas/Role.json', tsType: 'Role'},
+		},
+		required: ['role'],
```
ryanatkn, line 83:
```suggestion
		required: ['actor', 'role'],
```

src/lib/vocab/role/roleEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+			name: {type: 'string'},
+		},
+		required: ['actor', 'role_id', 'name'],
+		additionalProperties: false,
+	},
+	response: {
+		$id: '/schemas/UpdateRoleResponse.json',
+		type: 'object',
+		properties: {
+			actor: {type: 'number'},
```
ryanatkn, line 80:
```suggestion
```

PR #480, opened 2022-09-20
add `authorize` service property

ryanatkn:
Adds the `authorize` property to service events. Like `authenticate`, it's omitted by default which means `true`. If `authenticate` is false, `authorize` must be false, and if `authorize` is true/omitted, `authenticate` must be true/omitted. Includes tests to ensure we don't mess up these constraints. We could probably use types instead by splitting `ServiceEventInfo` apart into 3 versions for authorized, non-authorized, and non-authenticated, but I'm not sure it's worth the complexity.

followup:

- generate types for authorized, non-authorized, and non-authenticated service requests using `authorize` and `authenticate` schema values in #482


src/lib/server/services.test.ts
```diff
@@ -9,6 +9,66 @@ import {services} from '$lib/server/services';
...
+			);
+			assert.ok(
+				event.authenticate === undefined || event.authenticate,
+				'authorized events must have an authenticate value of true or undefined',
+			);
+		}
+	});
+}
+
+test__serviceDefinitions(`check for duplicate HTTP route paths`, async () => {
```
ryanatkn, line 57: This is the same block as below, just relocated so we test for static problems before calling the services below.

src/lib/server/services.test.ts
```diff
@@ -18,51 +78,40 @@ test__services.after(teardownDb);
 const session = new SessionApiMock(); // reuse the session so it tests login sequentially
 
 for (const service of services.values()) {
-	test__services(`perform service ${service.event.name}`, async ({db, random}) => {
+	const {event} = service;
```
ryanatkn, line 81: Just destructured `event` here, shortening the code below, no other changes in this test.

src/lib/server/services.test.ts
```diff
@@ -18,51 +78,40 @@ test__services.after(teardownDb);
...
-
-	for (const service of services.values()) {
-		const key = service.event.route.method + ':' + service.event.route.path;
-		if (paths.has(key)) {
-			throw Error(`Duplicate service event route ${key}`);
-		}
-		paths.add(key);
-	}
-});
-
```
ryanatkn, line 66: This test was relocated above, makes sense to test for static problems before running the live service tests.

greatbacon (approved):

PR #481, opened 2022-09-20
Missing `persona` property on TodoItem svelte:self component

greatbacon:
Looks like this got missed in the actor refactor. Understandable since it throws no compile or type errors 🙃 

Not sure what we could change to catch this bug in the future though 😞


ryanatkn:
Good catch, yeah hmm, looks like it's just an unimplemented feature of the Svelte typechecker: https://github.com/sveltejs/language-tools/issues/ 1229

ryanatkn (approved):

PR #482, opened 2022-09-21
add service request actor

ryanatkn:
Adds `actor` to the service request alongside `account_id`, `params`, `db`, etc. The services are now typesafe with `account_id` whereas previously they weren't.

To achieve typesafety, it splits the `Service` into 3 types:

- `NonAuthenticated` - no `account_id`, no `actor` - `Login`
- `NonAuthorized` - yes `account_id`, no `actor` - `Logout`, `CreateAccountPersona`, `Ping`
- `Authorized` - yes `account_id`, yes `actor` - the default for all other events

The types of `Service` and `ServiceRequest` now have one of the 3 prefixes depending on the event data.

Because authorized service requests now have an `actor` property, it required a lot of refactoring, particularly in the tests.

I'm open to renaming things, maybe:

- `NonAuthenticated` -> `Public`
- `NonAuthorized` -> `Authenticated`
- `Authorized` -> same

I originally called them `Unauthenticated` and `Unauthorized` but that seems misleading.

Followup:

- use `serviceRequest.actor` with permissions checks


src/lib/vocab/community/communityServices.ts
```diff
@@ -291,7 +291,7 @@ export const initAdminCommunity = async (
 
 	// Create the community's default spaces.
 	const createSpacesResult = await createDefaultAdminSpaces(
-		serviceRequest,
+		{...serviceRequest, actor: persona}, // TODO note this uses the community persona, because no user personas exist yet, but this may cause issues
```
ryanatkn, line 294: This is the biggest questionmark that I know of in this PR. I'll look again to see if things can be restructured.

src/lib/server/service.ts
```diff
@@ -60,7 +112,12 @@ export const toServiceRequest = <TParams = any, TResult extends Result = any>(
 							}
 						}),
 		params,
-		account_id, // TODO how to handle this type for services that don't require an account_id?
 		session,
 	};
-};
+	// TODO this is a bit hacky -- it may be preferred to have 3 different versions of `toServiceRequest` instead
```
ryanatkn, line 117: Calling out this todo because it's one of the rougher parts of the PR, combined with the gnarly type signature above.

An alternative might be to create 3 different functions instead of `toServiceRequest`: `toNonAuthenticatedServiceRequest`, `toNonAuthorizedServiceRequest`, and `toAuthorizedServiceRequest` -- but that may cause more problems than it solves.

src/lib/app/eventTypes.gen.ts
```diff
@@ -82,7 +83,12 @@ ${eventInfo.name}: ${toResponseName(eventInfo.name)};
 export interface ServiceByName {
 	${Array.from(services.values()).reduce((str, service) => {
 		const {name} = service.event;
-		return str + `${name}: Service<${toParamsName(name)}, ${toResponseResultName(name)}>;`.trim();
+		return (
+			str +
+			`${name}: ${toServiceAuthPrefix(service)}Service<${toParamsName(
+				name,
+			)}, ${toResponseResultName(name)}>;`.trim()
+		);
```
ryanatkn, line 91: This code adds one of the new prefixes to the generated service types based on `event.authenticate` and `event.authorize`, see the helper below.

src/lib/server/service.ts
greatbacon: I think some messiness up front is okay. I think refactors are easier when you come back to code a little cold anyway, since it makes it easier to spot hard to understand/unclear pieces.

greatbacon (approved):

PR #483, opened 2022-09-21
log timestamps in production

ryanatkn:
Adds timestamp logging to production.


src/lib/server/server.ts
```diff
@@ -5,19 +5,22 @@ import {createServer as create_http_server} from 'http';
...
 import {services} from '$lib/server/services';
 import {db} from '$lib/db/db';
 import {API_SERVER_PORT, VITE_DEPLOY_SERVER_HOST} from '$lib/config';
 
-const log = new Logger('[server]');
-
+// Global logging setup
 if (process.env.NODE_ENV === 'production') {
 	configureLogLevel(LogLevel.Info);
+	(Logger.prefixes as unknown[]).unshift(() => format(new Date(), 'M/d H:mm:ss.SSS'));
```
ryanatkn, line 19: The typecast `as unknown[]` will be unnecessary after upgrading Felt: https://github.com/feltcoop/felt/pull/246

ryanatkn, line 19: The format `'M/d H:mm:ss.SSS'` outputs `9/21 15:02:42.946` in local time.

Any thoughts? Not sure if the timezone/GMT is worth the noise.

src/lib/server/server.ts
```diff
@@ -5,19 +5,24 @@ import {createServer as create_http_server} from 'http';
...
 }
 
+const log = new Logger('[server]');
+
+log.info('startup time:', format(new Date(), 'P H:mm:ss.SSS O'));
```
ryanatkn, line 24: I wanted to keep the timestamp concise but useful, so this logs the full time including the year and timezone, and doubles as the moment we can see when the server (re)starts: `[server] startup time: 09/25/2022 15:02:12.727 GMT-5`

greatbacon (approved):
Yay timestamps!

PR #484, opened 2022-09-22
optimize tie collections

ryanatkn:
Optimizes tie data structures for efficiency and ergonomics using sets and a global `tieById` map using the `tie_id` added in #485

Some view code had to change in a minor way because we no longer have the sorted arrays in the lowest-level data layer. Instead, sorting/filtering becomes a concern of the views and queries. Queries will contain filtered+sorted subsets of the core collections, which are now backed by efficient, easy-to-use sets (see also #476). This paves the way for our query system that's currently in the concept stage.

followup:

- change `DeleteTie` to use `tie_id`
- look for other places to use `tie_id` instead of more complex logic


src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -42,46 +42,48 @@ export const stashEntities = (ui: WritableUi, $entities: Entity[]): void => {
...
 		const $sourceTies = sourceTies.get().value;
-		// TODO optimize lookup, maybe with a map by cached tie key
-		if (
-			!$sourceTies.find(
-				($t) => $t.dest_id === dest_id && $t.source_id === source_id && $t.type === type,
-			)
-		) {
-			$sourceTies.push($tie);
+		if (!$sourceTies.has($tie)) {
+			$sourceTies.add($tie);
```
ryanatkn, line 74: so much cleaner/faster

greatbacon (approved):

greatbacon (approved):

PR #485, opened 2022-09-22
add `tie.tie_id`

ryanatkn:
Adds `tie.tie_id`, saving client-side implementation for a followup. The client needs some unique identifier to efficiently handle ties, and we also want to mirror the same data on the server for simplicity. (otherwise you'd have to have a `ClientTie` version of `Tie`) Rather than doing a key computed by either the server repos or client deserialization, we decided to add a `tie_id` property to all ties.

We talked about using [`bigserial`](https://www.postgresql.org/docs/current/datatype-numeric.html) instead of `serial`, but unfortunately [`ajv-keywords`](https://github.com/ajv-validator/ajv-keywords) doesn't support it yet: https://github.com/ajv-validator/ajv-keywords/issues/ 251 - it should be fairly easy to convert to `bigserial` when it's supported, because SvelteKit supports `BigInt` in data by default. The `serial` type supports 2  billion items so it's ok to punt.

followup:

- use this in #484


src/lib/db/migrations/00029-add-tie-id.js
```diff
@@ -0,0 +1,19 @@
...
+	await sql`
+		ALTER TABLE ties
+		ADD UNIQUE (source_id, dest_id, type);
+	`;
+	// TODO we'd prefer to use `bigserial` here instead of `serial`,
+	// but ajv-keywords doesn't currently support it:
+	// https://github.com/ajv-validator/ajv-keywords/issues/251
+	await sql`
+		ALTER TABLE ties
+		ADD COLUMN tie_id serial PRIMARY KEY;
```
greatbacon, line 17: I presume this auto-backfills the `tie_id` column for our existing entities?  Also I scooped your migration number in another PR so you'll need to increment to 00030 😸

greatbacon (approved):

src/lib/db/migrations/00029-add-tie-id.js
ryanatkn: Yep it does, tested to make sure with `gro lib/db/testMigration`, and it seems to preserve the original insertion order linearly. Also good thinking, it would have caused CI to fail only after merging.

PR #488, opened 2022-09-23
add entity data indices

ryanatkn:
Adds database indices for `entity.data.type` and `entity.data.space_id`.

References:

- https://stackoverflow.com/questions/36075918/postgresql-index-on-json
- https://stackoverflow.com/questions/26499266/whats-the-proper-index-for-querying-structures-in-arrays-in-postgres-jsonb/27708358#27708358


src/lib/db/migrations/00031-index-entity-data.js
```diff
@@ -0,0 +1,14 @@
...
+	// Add indices for the `data` property of entities.
+	await sql`
+		CREATE INDEX entities_data_type_idx
+		ON entities ((data->>'type'));
+	`;
+
+	await sql`
+		CREATE INDEX entities_data_space_id_idx
+		ON entities (((data->>'space_id')::int))
+		WHERE (data->>'space_id') IS NOT NULL;
```
greatbacon, line 12:
```suggestion
		CREATE INDEX entities_data_type_idx ON entities USING gin (data);
```
Okay, I think I've got a working solution based on this page: https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/

It's a more generic index purpose built for JSONB data. We'll need to update our queries to take advantage of it though. Like instead of
```
SELECT * FROM entities WHERE data->>'type' = 'Collection';
SELECT * FROM entities WHERE (data->>'space_id')::int = 1;
```

We would instead use path operators
```
SELECT * FROM entities WHERE data @> '{"type":"Collection"}';
SELECT * FROM entities WHERE data @> '{"space_id":1}';
```

Could also still build B-Tree indexes out too if needed/prefered.

greatbacon: If you want to test it out locally, I found that the size of the DB on fresh create isn't big enough, but if you run `gro test` a couple of times it gets big enough to prefer the index.  You'll know it's running because it says `Bitmap Heap scan` plus the index name lol

ryanatkn: Awesome, this looks good to me, gives us a lot of flexibility going forward. The downside may be performance in some cases but we can always address that when/if it arises.

greatbacon (approved):

PR #489, opened 2022-09-23
use same port for both dev and prod

ryanatkn:
This simplifies port usage to be the same for both dev and prod. Since Vite/SvelteKit moved to port 5173 by default, we can now point the API server to 3000 in both environments and eliminate a lot of complexity.

Looks like our production system won't need any changes.

For development, we'll now browse to `localhost:5173` like all of our other SvelteKit projects. Vite also now auto-increments the port if it's taken.


greatbacon (approved):
🧹

PR #491, opened 2022-09-27
Add account settings

greatbacon:
This PR adds a functional first pass for managing account settings.

The editor dialogue utilizes the same raw JSON editor that the EntityEditor uses. We'll want a more usable design for UX purposes, for for now it gets the job done, lol


src/lib/vocab/account/accountServices.ts
```diff
@@ -0,0 +1,20 @@
+import {blue, gray} from 'kleur/colors';
+import {Logger} from '@feltcoop/felt/util/log.js';
+
+import type {ServiceByName} from '$lib/app/eventTypes';
+import {UpdateAccountSettings} from '$lib/vocab/account/accountEvents';
+
+const log = new Logger(gray('[') + blue('communityServices') + gray(']'));
```
ryanatkn, line 7:
```suggestion
const log = new Logger(gray('[') + blue('accountServices') + gray(']'));
```

src/lib/vocab/account/accountEvents.ts
```diff
@@ -0,0 +1,26 @@
+import type {ServiceEventInfo} from '$lib/vocab/event/event';
+
+export const UpdateAccountSettings: ServiceEventInfo = {
+	type: 'ServiceEvent',
+	name: 'UpdateAccountSettings',
+	broadcast: true,
```
ryanatkn, line 6:
```suggestion
	broadcast: false,
```

I think we'll want to broadcast this to all connected clients for this account, and that may reuse `broadcast: true` with proper scoping (scopes I can think of: account, community, space), but for now it'll be less of a security problem to just make it not broadcast.

src/lib/vocab/account/account.schema.ts
```diff
@@ -18,9 +21,24 @@ export const AccountModelSchema = {
...
+	additionalProperties: false,
+};
+
+export const AccountSettingsSchema = {
+	$id: '/schemas/AccountSettings.json',
+	type: 'object',
+	properties: {
+		darkmode: {type: 'boolean'},
+	},
+	required: ['darkmode'],
```
ryanatkn, line 38:
```suggestion
	// required: [], // currently no settings are required, so we don't need migrations to add defaults for every property
```

I think this is better for now, maybe even as a permanent design choice. By making all settings optional, we won't have to do migrations for settings additions. Otherwise we'd have to change the migration above from `NOT NULL DEFAULT '{}';` to `NOT NULL DEFAULT '{"darkmode": false}';` and keep it in sync with `toDefaultAccountSettings `.

src/lib/vocab/account/account.schema.ts
```diff
@@ -18,9 +21,24 @@ export const AccountModelSchema = {
...
-	required: ['account_id', 'name', 'created', 'updated'],
+	required: ['account_id', 'name', 'settings', 'created', 'updated'],
+	additionalProperties: false,
+};
+
+export const AccountSettingsSchema = {
+	$id: '/schemas/AccountSettings.json',
+	type: 'object',
+	properties: {
+		darkmode: {type: 'boolean'},
```
ryanatkn, line 36:
```suggestion
		theme: {anyOf: [{type: 'string', enum: ['light', 'dark']}, {type: 'object'}]},
```

I think `theme` is better for flexibility, this way we can add any app-supported strings in the future beyond `'light` | 'dark'`, and also have a custom object, something like `{extends: 'light', vars: {color: '#123'}}`, the details of which we can figure out later. Implementing that wouldn't take much effort at all. Renaming this now means we won't need a migration.

greatbacon: I think for now, leaving darkmode as a place holder should be okay (especially since we're not asserting any required fields on the schema). That way we can leave the type def for the setting up to the actual PR to implement it.

I could also just blow away darkmode as a property on the schema.

ryanatkn: Sounds good, I forgot since we can edit the JSON directly we can fix any issues ourselves.

ryanatkn (approved):

PR #492, opened 2022-09-28
remove color logging from production

ryanatkn:
- remove color logging from production
- log dispatch colors on the client (doesn't work on the server, just creates noise)
- properly configure client logging on the server during SSR -- the root layout now calls `configureLogLevel` similarly to `server.ts`


greatbacon (approved):
Love a good logging PR 🤓

PR #493, opened 2022-09-29
Create/Read Role ux

greatbacon:
This PR sets the stage for building out the Role UX.  It's adds `Roles` to the `ClientSession` as well as a `Manage Roles` option to the CommunityContextMenu.  You can now create roles in a community, and the information will broadcast across the system, resulting in reactive behavior in the menu.

Updating & Deleting functionality to follow in another PR.


src/lib/ui/ui.ts
```diff
@@ -52,16 +53,19 @@ export interface Ui {
 	sessionPersonaIds: Readable<Set<number>>;
 	sessionPersonaIndices: Readable<Map<Readable<Persona>, number>>;
 	communities: Mutable<Array<Readable<Community>>>;
+	roles: Mutable<Array<Readable<Role>>>;
 	spaces: Mutable<Array<Readable<Space>>>;
 	memberships: Mutable<Array<Readable<Membership>>>;
 	personaById: Map<number, Readable<Persona>>;
 	communityById: Map<number, Readable<Community>>;
+	rolesById: Map<number, Readable<Role>>;
```
ryanatkn, line 61: should be `roleById` instead

src/lib/vocab/role/roleMutationHelpers.ts
```diff
@@ -0,0 +1,22 @@
...
+	const {rolesById, roles} = ui;
+	const {role_id} = $role;
+	let role = rolesById.get(role_id);
+	if (role) {
+		role.set($role);
+	} else {
+		role = writable($role);
+		rolesById.set(role_id, role);
+		roles.get().value.push(role);
+		roles.mutate();
```
ryanatkn, line 20:
```suggestion
		roles.mutate(($v) => $v.push(role));
```

Been using the `_.mutate();` pattern only for batching or conditional mutation.

ryanatkn (approved):
nice nice nice, getting so close

src/lib/ui/ui.ts
greatbacon: 🙃

PR #498, opened 2022-10-03
Add update & delete functionality for Roles

greatbacon:
This PR adds update & delete functionality for Roles.

The update is based on a selection & can currently only be triggered by enter key
Delete uses a button and a new confirmation dialog to trigger.

Currently there is a type issue around the delete open dialog that I am unsure of how to deal with. I guess it might be fixed by pulling things out into their own component like my todo says but I got lazy lol


src/lib/vocab/role/roleMutations.ts
```diff
@@ -10,18 +10,22 @@ export const CreateRole: Mutations['CreateRole'] = async ({invoke, ui}) => {
...
+		name: params.name,
+	}));
 	return result;
 };
 
-export const DeleteRoles: Mutations['DeleteRoles'] = async ({invoke}) => {
+export const DeleteRoles: Mutations['DeleteRoles'] = async ({invoke, params, ui}) => {
 	const result = await invoke();
 	if (!result.ok) return result;
 	//TODO update stores here
```
ryanatkn, line 27:
```suggestion
```

ryanatkn:
In `CommunityContextmenu.svelte`, adding this line with `dialogProps` makes the dialog flow from the top, so it doesn't recenter on the page as you do things:

```
					Component: ManageRolesForm,
					dialogProps: {layout: 'page'},
					props: {persona, community},
```

ryanatkn (approved):
excellent!

PR #499, opened 2022-10-06
Default roles

greatbacon:
This PR adds a DB migration & systems for creating a default role per community from a systems perspective. 

It currently does not add any UX for managing or changing default roles atm


src/lib/vocab/role/RoleRepo.ts
```diff
@@ -59,4 +60,15 @@ export class RoleRepo extends PostgresRepo {
 		log.trace('[filterByAccount]', result.length);
 		return {ok: true, value: result};
 	}
+
+	async containsDefaultRole(roleIds: number[]): Promise<Result<{value: boolean}>> {
```
ryanatkn, line 64: This seems like business logic that belongs in the `DeleteRoles` service. Doesn't it seem simpler to load the community, run the business logic, return the appropriate code+message? That way we won't need a special-cased repo method, we won't need to index settings, and the constraint lives in one place in the service instead of being duplicated on the client.

src/lib/vocab/role/roleServices.ts
```diff
@@ -56,10 +59,37 @@ export const DeleteRolesService: ServiceByName['DeleteRoles'] = {
...
 				return {ok: false, status: 500, message: 'failed to delete roles'};
 			}
 			return {ok: true, status: 200, value: null};
 		}),
 };
+
+//TODO we might also want to confirm we're not deleting the last role in a community
+//but that would involve getting the DB into a state where there is no defaultRole or
+//it is not actually pointing at a role inside the community
+export const validateRolesForDeletion = async (
```
ryanatkn, line 78: Adding this helper seems to complicate things -- if it's inline in the service, you can return the appropriate code+message instead of a generic failure, without adding the surface area of the `delete` value and the verbose function signatures. Any code that needs to delete roles can call the service function directly. Related is the repo method -- it seems simpler to just load the community record in the service and run business logic like we do elsewhere.

src/lib/vocab/persona/personaEvents.ts
```diff
@@ -19,14 +19,15 @@ export const CreateAccountPersona: ServiceEventInfo = {
 		properties: {
 			persona: {$ref: '/schemas/AccountPersona.json', tsType: 'AccountPersona'},
 			community: {$ref: '/schemas/Community.json', tsType: 'Community'},
+			role: {$ref: '/schemas/Role.json', tsType: 'Role'},
```
ryanatkn, line 22: It may be convenient later to make this `roles`, similar to `spaces` and `directories`, so we can return any number of roles in the future, like when creating a community from a template that includes both mods and members.

src/lib/vocab/community/communityServices.ts
```diff
@@ -286,5 +313,39 @@ export const initAdminCommunity = async (
...
 	if (!createMembershipResult.ok) return createMembershipResult;
 
-	return {ok: true, value: {community, persona}};
+	return {ok: true, value: {community, persona, role}};
+};
+
+export const initDefaultRoleForCommunity = async (
+	repos: Repos,
+	community: Community,
+): Promise<Result<{value?: {community: Community; defaultRole: Role}}>> => {
```
ryanatkn, line 322: Looks like `value` shouldn't be optional, and then you can remove the `!` in the two `.value!.` assertions above.

src/lib/vocab/role/RoleRepo.ts
greatbacon: Yeah, fair enough. I think I was trying to solve a different problem and had some blinders on. I think there are some complications around the "just load the community record" though, with a function that allows for bulk deletions like this.

I'll scrap this function and condense things back into the service, I think I'll reduce the surface of the function to only handle deleting one role at a time though (as that's how the UX presents it anyway)

src/lib/ui/ManageRolesItem.svelte
```diff
@@ -5,17 +5,21 @@
 	import type {Role} from '$lib/vocab/role/role';
 
 	export let role: Readable<Role>;
+	export let defaultRoleId: number | undefined;
```
ryanatkn, line 8:
```suggestion
	export let defaultRoleId: number;
```

src/lib/vocab/community/community.schema.ts
```diff
@@ -21,6 +21,7 @@ export const CommunitySettingsSchema = {
 	type: 'object',
 	properties: {
 		hue: {type: 'number'},
+		defaultRoleId: {type: 'number'},
 	},
 	required: ['hue'],
```
ryanatkn, line 26:
```suggestion
	required: ['hue', 'defaultRoleId'],
```

ryanatkn (approved):
awesome! big one. I think `defaultRoleId` is required now, had a couple suggested merges if so.

PR #500, opened 2022-10-12
Fix authorization in membership invitation flow

greatbacon:
Fixed a bug in the invite flow where we were using the person_id of the member to be invited as the actor as well, resulting in an authorization block when inviting personas not on your account.


ryanatkn (approved):
whoops, good to see the authorization check works tho!

PR #503, opened 2022-10-13
batch `stashRoles` and implement `ReadRoles` mutation

ryanatkn:
Batches `stashRoles` and uses it to implement the `ReadRoles` mutation.


src/lib/vocab/role/roleMutations.ts
```diff
@@ -1,19 +1,19 @@
 import type {Mutations} from '$lib/app/eventTypes';
-import {evictRoles, stashRole} from '$lib/vocab/role/roleMutationHelpers';
+import {evictRoles, stashRoles} from '$lib/vocab/role/roleMutationHelpers';
 
 export const CreateRole: Mutations['CreateRole'] = async ({invoke, ui}) => {
 	const result = await invoke();
 	if (!result.ok) return result;
 	const {role: $role} = result.value;
-	stashRole(ui, $role);
+	stashRoles(ui, [$role]);
```
ryanatkn, line 8: Once our patterns shake out a bit more, I think we'll be able to have `stashRole` that's composed by `stashRoles`, avoiding the need for this wrapper array, but for now this is a bit simpler. The followup PR in #504 makes a lot of progress on making them seamlessly composable, so we can look at things after that to see if we want to extract singular helpers from the batched ones.

greatbacon (approved):

PR #504, opened 2022-10-13
add `Mutated` to batch multiple mutations

ryanatkn:
Adds a new helper class to solidify the `mutated` pattern that we've been using in mutation helpers, and uses it to batch `SetSession` mutations all in one go. It also adds a new pattern to stash mutation helpers, the optional `replace` boolean, to add the functionality used in `SetSession` of replacing whatever's stashed, so `SetSession` no longer duplicates much of the mutation helpers.

`Mutated` lets us batch changes across multiple mutation helpers, improving the composition of the helpers. (e.g. we can have `stashRole` that gets called by `stashRoles` with seamless batching, currently this isn't possible) This is an optimization and avoids some issues where store listeners would get called with in-between states. It also gives us granular control over the mutation batch, like sorting if we ever need it, and centralizes stashing in one place instead of two with `SetSession`, which will be important as we refine the reactive data layer.

The idea is that mutations can instantiate the class with some key as a default arg, forward the instance to any other mutations they call (either created right then or received from the caller), and then call `end` with the original key. This means only the outermost mutation helper's `mutated` instance will apply the mutations, so we get everything in a single batch. 

Importantly, the state of the mutable stores is updated as the mutations run, but the store change function is only called at the end, meaning any derived stores will not be fresh inside a mutation. This is the same as before, but now it also applies across composed mutation helpers.

followup:

- thread a `mutated` instance in mutations that call multiple helpers (feels like this PR is already big enough)
- avoid repeated `.get().value` in helpers using a local var (unless decomposed into singular helpers?)


src/lib/ui/ui.ts
```diff
@@ -49,7 +49,7 @@ export interface Ui {
 	account: Readable<AccountModel | null>;
 	personas: Mutable<Array<Readable<Persona>>>;
 	session: Readable<ClientSession>;
-	sessionPersonas: Readable<Array<Readable<Persona>>>;
+	sessionPersonas: Mutable<Array<Readable<Persona>>>;
```
ryanatkn, line 52: Made this data type change so it can be batched with everything else. Mutable seems to be working really well for us, just adds a little churn when `.value` is needed.

src/lib/session/sessionMutations.ts
```diff
@@ -43,43 +39,26 @@ export const SetSession: Mutations['SetSession'] = async ({params, ui}) => {
...
-	const $spaces = guest ? [] : $session.spaces.map((s) => writable(s));
-	spaceById.clear();
-	$spaces.forEach((s) => spaceById.set(s.get().space_id, s));
-	spaces.swap($spaces);
+	const mutated = new Mutated('SetSession');
+	stashPersonas(ui, guest ? [] : toInitialPersonas($session), mutated, true);
+	stashCommunities(ui, guest ? [] : $session.communities, mutated, true);
+	stashRoles(ui, guest ? [] : $session.roles, mutated, true);
+	stashSpaces(ui, guest ? [] : $session.spaces, undefined, mutated, true);
+	mutated.end('SetSession');
```
ryanatkn, line 50: This will be a lot easier to maintain: now that the `stash` pattern includes the `replace` boolean, we no longer have all of this logic duplicated.

src/lib/vocab/community/communityMutationHelpers.ts
```diff
@@ -12,20 +12,39 @@ import type {Membership} from '$lib/vocab//membership/membership';
...
+	replace = false,
+): void => {
+	const {communityById, communities} = ui;
+	if (replace) {
+		communityById.clear();
+		communities.get().value.length = 0;
+		mutated.add(communities);
+	}
+	for (const $community of $communities) {
+		stashCommunity(ui, $community, EMPTY_ARRAY, EMPTY_ARRAY, EMPTY_ARRAY, mutated);
```
ryanatkn, line 31: These `EMPTY_ARRAY`s are a bit weird -- now that we can compose helpers with batching, it may make sense to separate `stashCommunity` from `stashCommunityWithData` or something, or decompose the existing `stashCommunity` and call `stashSpaces`/`stashEntities`/`stashMemberships`/etc

greatbacon (approved):

PR #507, opened 2022-10-13
improve confirm dialog

ryanatkn:
Improves the confirm dialog a bit -- it's now simpler to use from the caller's perspective and automatically handles promises, results, and error messages.


src/lib/ui/ConfirmDialog.svelte
```diff
@@ -1,16 +1,51 @@
...
+
+	import {getApp} from '$lib/ui/app';
+
+	const {dispatch} = getApp();
+
 	export let done: (() => void) | undefined = undefined;
-	export let action: (() => void) | undefined = undefined;
-	export let description: string;
+	export let action: (() => any) | undefined = undefined;
+	export let promptText = 'Are you sure?';
```
ryanatkn, line 11: These props to customize the question/buttons works okay, but it doesn't allow arbitrary markup. Currently none of our dialogs can be instantiated with custom markup (via Svelte slots), because we call `dispatch` with some plain data, and you can't pass Svelte markup that way, because it needs to be compiled. I'll be thinking if we want to do away with the `dispatch.OpenDialog` event entirely, and change the API to be declarative, so you'd conditionally render `<Dialog ...` in any component rather than calling `dispatch.OpenDialog`. It means dialogs can't be driven programmatically in the same way, which is why I originally wanted to make them event-driven -- will see how our use cases align with a redesign. It won't take much work to change patterns.

greatbacon: I don't know if we need to go down this rabbit hole at the moment? I find the current OpenDialog system to be very functional and works hand in glove with our event system.  I've had no real issue with it.

I don't really think we want to be passing arbitrary Markdown to dialogs anyway. Like at that point we can just build another dialog component if we're going to be having radically different content between things.

I would strongly suggest we not go down this path and instead keep these foundational architecture pieces simple & sturdy in place, especially since we're trying to get a stable alpha out the door before the end of the year.

ryanatkn: That sounds good. My main problem with it right now is type safety. The other aspects won't be relevant probably til after beta.

greatbacon (approved):

PR #508, opened 2022-10-13
unwrap 500 errors in services

ryanatkn:
As discussed here: https://github.com/feltcoop/felt-server/pull/318#discussion_r841353074

Changes service code to unwrap most repo method results, removing ~300 LOC and making services significantly easier to read/write. The criteria for using `unwrap`:

- calls to `Result`-returning repo methods and helpers that compose repo methods
- where `!result.ok` is checked
- that return `status: 500` (this isn't a hard requirement for the future, we could change `unwrap` to handle this, but for now this is all I've changed -- this PR in felt paves the way for further improvements: https://github.com/feltcoop/felt/pull/252 )

Here's tradeoffs I can think of:

- more concise service code (when error messages are useless to the client, we no longer need to manually handle or write messages for them -- in most cases these are unexpected failures, and expected failures can continue to be handled manually for UX reasons)
- we're generically logging the stack trace of those errors, so we can still diagnose where things go wrong as well as before
- avoids leaking error messages from the repos to the client, and isolates the message concern more explicitly in the services, not repos, similar to status codes (will also be useful for i19n) (repos can still return named errors that are manually handled by services, so we don't lose that ability to have customized errors for UX reasons)

Inconsistencies:

- the `findById` repo methods that return `NOT_OK` and have services assume a 404. I think the concern probably belongs in the service, so the type should be changed to `T | undefined` so the service can infer `undefined` means `status: 404`, but I didn't make that change.
- `repos.entity.filterByIds` is the only repo method that returns a custom message, I left it in place -- this is related to the `findById` calls possibly returning `undefined`
- `repos.persona.filterByIds` doesn't check for the same count like the entity version


greatbacon (approved):
Sweet! 🧹 

Only thought I have is maybe leave some comment breadcrumbs around the inconsistencies you noted in the PR body?

ryanatkn:
Sounds good, added those todos for a followup PR.

PR #510, opened 2022-10-15
propagate `ResultError`s and use `unwrap` for service composition

ryanatkn:
Uses the `ResultError` added in https://github.com/feltcoop/felt/pull/252 to improve error handling in our websocket/http middlewares. The idea is that `unwrap` now forwards the result through a custom error class, so we can use `unwrap` in more places without losing any information.


src/lib/server/service.ts
```diff
@@ -93,16 +93,15 @@ export function toServiceRequest<TParams = any, TResult extends Result = any>(
 	let result: TResult; // cache to pass through if the inner transaction promise rejects
 	const req: NonAuthenticatedServiceRequest = {
 		repos: db.repos,
-		// TODO support creating new transactions outside of this singleton, which is needed for service composition
```
ryanatkn, line 96: This todo is resolved by simply calling `db.sql.begin` wherever we'd want a separate transaction. The rest of the changes in this module are no-op cleanups.

PR #514, opened 2022-10-16
add query concept that includes the async status

ryanatkn:
This is a first step towards a more usable entity/tie query system. It renames the existing `entitiesBySourceId`, which wrapped a reactive collection of entities, to `queryByKey`, which wraps the entities along with an `AsyncStatus` (initial/pending/success/failure). This lets us disambiguate an empty query from an in-progress one.

Followups:

- handle failures and display the error message, maybe adding `query.error = {message, code}` (see the comment below)
- figure out pagination, sorting, and filtering
- query by more than a source id (currently the "key" is always just a `source_id`)
- include ties?


src/lib/ui/view/Board.svelte
```diff
@@ -58,6 +59,9 @@
 						})}>Submit a new post</button
 				>
 			{/if}
+			<!-- TODO handle query failures -->
+			<!-- {:else if $queryStatus === 'failure'}
+				<Message status="error">{$queryError.message}</Message> -->
```
ryanatkn, line 64: This is what the error condition could look like.

greatbacon: In that case would that technically be a different query, since the key would be different? Like `source_id=1&page=1` vs `source_id1&page=2` as the keys?

greatbacon (approved):

PR #520, opened 2022-10-18
Replace DeleteMembership use in UX with LeaveCommunity

greatbacon:
This PR is mostly a semantic change. It takes the old DeleteMembership logic & ports it to a new LeaveCommunity event. The old DeleteMembership event will be refactored to rely on `membership_id` at a later date.


src/lib/vocab/community/communityMutations.ts
```diff
@@ -84,3 +84,26 @@ export const DeleteCommunity: Mutations['DeleteCommunity'] = async ({params, inv
...
+export const LeaveCommunity: Mutations['LeaveCommunity'] = async ({params, invoke, ui}) => {
+	const {memberships} = ui;
+	const result = await invoke();
+	if (!result.ok) return result;
+	const {actor, community_id} = params;
+	const otherMemberships = memberships
+		.get()
+		.value.filter((m) => m.get().community_id === community_id && m.get().persona_id !== actor);
+
+	if (otherMemberships.length > 0) {
```
greatbacon, line 97: This is maybe the most complex part (and it's not that complex). Basically we just check to see if there were any non-persona memberships in the community, and if so just evict the acting personas memberships, otherwise fully evict the community and it's data.

greatbacon:
Also unsure why prettier is kicking up a fuss, since it's working locally for me :(

ryanatkn:
Probably deps from the recent PR, try running `npm i` and then `gro format`, or saving `src/routes/+layout.svelte` directly.

src/lib/vocab/community/communityEvents.ts
```diff
@@ -152,3 +152,28 @@ export const DeleteCommunity: ServiceEventInfo = {
...
+		additionalProperties: false,
+	},
+	response: {
+		$id: '/schemas/LeaveCommunityResponse.json',
+		type: 'null',
+	},
+	returns: 'Promise<LeaveCommunityResponseResult>',
+	route: {
+		path: '/api/v1/community/leave',
+		method: 'DELETE',
```
ryanatkn, line 177:
```suggestion
		method: 'POST',
```

I think this is better?

ryanatkn (approved):

src/lib/vocab/community/communityEvents.ts
greatbacon: probably lol, I'm never sure when deviating from REST

PR #524, opened 2022-10-20
rename Login and Logout to SignIn and SignOut

ryanatkn:
Renames `Login` to `SignIn` and `Logout` to `SignOut` in preparation for the `SignUp` flow changes.


src/lib/session/sessionEvents.ts
```diff
@@ -38,40 +38,39 @@ export const Login: ServiceEventInfo = {
...
 			session: {type: 'object', tsType: 'ClientAccountSession'},
 		},
 		required: ['session'],
 		additionalProperties: false,
 	},
-	returns: 'Promise<LoginResponseResult>',
+	returns: 'Promise<SignInResponseResult>',
 	route: {
-		path: '/api/v1/login',
+		path: '/api/v1/signin',
```
ryanatkn, line 53: Should we use dash-case for these? e.g. `sign-in`

greatbacon: I think it's fine the way it is

greatbacon (approved):

PR #525, opened 2022-10-20
improve signup flow

ryanatkn:
Adds the `SignUp` event to complement `SignIn` and `SignOut`, making account creation explicit instead of automatic.

Account names must now be emails. Instead of `a` and `b` being seeded, they're now `a@a.a` and `b@b.b`.

We disallow `@` and `.` symbols in persona names in #526, so there's less chance for confusion between persona and account names.


src/lib/ui/SignInForm.svelte
```diff
@@ -18,11 +17,12 @@
 
 	$: disabled = !!submitting;
 
-	const login = async () => {
+	const signIn = async () => {
 		if (submitting) return;
 		if (!username) {
+			// TODO would be better UX to make the above check `!validateAccountName(username)` but staging accounts aren't emails
```
ryanatkn, line 23: This is a minor inconsistency with our staging data. Maybe we should fix? Migrate everyone's login?

src/lib/session/sessionServices.ts
```diff
@@ -1,47 +1,28 @@
...
+			const {params, session} = serviceRequest;
+
+			const username = scrubAccountName(params.username);
+			if (!username) {
+				return {ok: false, status: 400, message: 'invalid username'};
+			}
+
+			const account = unwrap(await repos.account.findByName(username));
+			if (!account) {
+				return {ok: false, status: 400, message: 'account does not exist'};
```
ryanatkn, line 21: This and signup leak information about which accounts are in the system, and it seems that we could fix the situation: <https://security.stackexchange.com/questions/13079/is-there-any-reason-to-show-the-same-message-for-invalid-username-as-password>

The main idea I took from this is that on signup, we could return a fake successful error message if the account already exists.

src/lib/util/randomEventParams.ts
```diff
@@ -29,17 +30,24 @@ export const randomEventParams: RandomEventParams = {
...
+	SetSession: async () => {
+		return {session: {guest: true}};
+	},
+	SignIn: async (_random, {account} = {}) => {
+		if (!account) throw Error('testing SignIn requires an account');
 		return {
-			username: randomString(),
-			password: randomString(),
+			username: account.name,
+			password: (account as any).__testPlaintextPassword,
```
ryanatkn, line 40: This hack is used in the generic tests for services and events. It's only added in those tests, no chance of making it to production.

src/lib/session/sessionServices.ts
```diff
@@ -1,47 +1,28 @@
...
-					toDefaultAccountSettings(),
-				);
-				if (createAccountResult.ok) {
-					account = createAccountResult.value;
-				} else {
-					// Failed to create the account.
-					return {ok: false, status: 500, message: 'failed to create account'};
-				}
+			if (!(await verifyPassword(params.password, account.password))) {
+				return {ok: false, status: 400, message: 'incorrect password'};
```
ryanatkn, line 25: It'd probably be better to avoid leaking account existence here, maybe following <https://security.stackexchange.com/questions/13079/is-there-any-reason-to-show-the-same-message-for-invalid-username-as-password>

A problem is that if the signup flow stops you from signing up, it leaks the same info. I'm not sure how to fake a success given that we don't force email verification to complete signing up.

src/lib/ui/SignUpForm.svelte
```diff
@@ -0,0 +1,129 @@
+<script lang="ts">
```
ryanatkn, line 1: This is a copy-paste of the SignInForm with some minor changes.

src/lib/ui/SignInForm.svelte
```diff
@@ -20,9 +20,11 @@
 
 	const signIn = async () => {
 		if (submitting) return;
+		username = scrubAccountName(username);
 		if (!username) {
+			// TODO would be better UX to make the above check `!checkAccountName(username)` but staging accounts aren't emails
```
ryanatkn, line 25: The UX is made slightly worse here because our staging data has account names that aren't emails. Maybe we should fix? Migrate everyone's login?

src/lib/session/sessionServices.ts
```diff
@@ -1,47 +1,29 @@
...
-			if ('account_id' in serviceRequest) {
-				session.signOut();
-				return {ok: false, status: 400, message: 'something went wrong, please try again'};
+			const {params, session} = serviceRequest;
+
+			const username = scrubAccountName(params.username);
+			if (!username) {
+				return {ok: false, status: 400, message: 'invalid username'};
+			}
+			// TODO could return a good error here for usernames that fail `checkAccountName` if we migrate production
```
ryanatkn, line 18: See also this comment: https://github.com/feltcoop/felt-server/pull/525/files#r1002325477

src/lib/util/randomVocab.ts
```diff
@@ -114,15 +114,18 @@ export interface RandomVocab {
...
 			await this.db.repos.account.create(
 				params.username,
 				params.password,
 				toDefaultAccountSettings(),
 			),
 		);
+		// This makes the unencrypted password available for tests,
+		// so things like `SignIn` can be tested with existing accounts.
+		(account as any).__testPlaintextPassword = params.password;
+		return account;
```
ryanatkn, line 128: This hack is used in the generic tests for services and events so our test code has access to the plaintext password, necessary for signing in to an existing account. It's only added in those tests, no chance of making it to production.

We could also reverse the account check logic here (instead of checking `if (!account)` to throw, checking `if (account)`), and force it to create the account so we know the password to sign in with, but that would require special casing the events and services tests for just the `SignIn` event. It seemed simpler to make it available everywhere.

src/lib/ui/style.css
```diff
@@ -128,3 +128,9 @@ maybe upstream to Felt? think about how to upstream without bloating bundles tha
 	/* TODO probably want `--button_color_selected` so it's different than active */
 	background-color: var(--button_color_active);
 }
+
+/* a minimal alternative to `<Message status="error">some error text</Message>` */
+.error-text {
+	font-weight: bold;
+	color: var(--error_color);
+}
```
ryanatkn, line 136: This seems broadly useful as an alternative to the more visually heavyweight `<Message status="error">` -- thinking we should upstream to Felt?

src/lib/ui/SignInForm.svelte
ryanatkn: We talked about this -- we're going to clear staging data and give good error messages.

src/lib/ui/style.css
greatbacon: I think the lighterweight class could be useful. I know a design pattern I see occasionally is that client-side validation has small text pop up next to forms when a check fails, and then a heavier weight design element (toast/dialog/etc) is used when the error came from the server.

greatbacon (approved):

PR #526, opened 2022-10-21
scrub and check persona and community names

ryanatkn:
Adds a new pattern for property scrubbing and checking (aka sanitizing/normalizing and validating) optimized for user-friendly error messages. `ajv` is used for validation, and this is currently a separate step, so I went with `scrub` and `check` to emphasize their different semantics. `check` is a bit weird in that it returns `null` when successful or a user-friendly error message string -- this is for efficiency. Maybe a better name would clarify how it works?

Implements the pattern for `persona.name` and reuses it for `community.name`, and includes tests. We discussed it and we'll use GitHub's restrictions of 39 alphanumeric characters with optional hyphens in between them, and a minimum of 3 characters.

Also upgrades Felt with a small styling fix.


src/lib/vocab/persona/personaHelpers.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Used for both persona and community names.
+ * @param name
+ * @returns The scrubbed value.
+ */
+export const scrubPersonaName = (name: string): string => name.trim();
+
+const PERSONA_NAME_CHARACTER_MATCHER = /^[a-z\d-]+$/iu;
+// TODO maybe add these to the schema? problem is ajv will return less friendly errors from service calls
+const PERSONA_NAME_MAX_LENGTH = 39; // same as github
+const PERSONA_NAME_MIN_LENGTH = 1;
```
ryanatkn, line 14: TODO change to 3

greatbacon (approved):

PR #527, opened 2022-10-24
add `UpdateAccountPassword`

ryanatkn:
Adds the `UpdateAccountPassword` event and related functionality. The `"Account settings"` accessible through the contextmenu now mount `UpdateAccountPasswordForm.svelte`.


src/lib/vocab/account/accountEvents.ts
```diff
@@ -15,11 +48,38 @@ export const UpdateAccountSettings: ServiceEventInfo = {
 	},
 	response: {
 		$id: '/schemas/UpdateAccountSettingsResponse.json',
-		type: 'null',
+		type: 'object',
+		tsType: 'AccountModel',
```
ryanatkn, line 52: Because `UpdateAccountSettings` and `UpdateAccountPassword` are essentially the same, I updated a view things about the former event. It now returns the `account` from the database with the new `updated` field, and uses it in the mutation to update the object on `ui.account`. You can see everything working by going to `"Account settings"` in the contextmenu, then changing either the password or the settings with the form, and seeing the updated text at the top of the dialog change.

src/lib/vocab/account/AccountRepo.ts
```diff
@@ -33,6 +33,16 @@ export class AccountRepo extends PostgresRepo {
 		return {ok: true, value: data[0]};
 	}
 
+	// TODO this shouldn't exist, should be `findById` with custom fields
+	async findByIdForAuth(account_id: number): Promise<Result<{value: Account | undefined}>> {
```
ryanatkn, line 37: This is the one wart on this PR that I know of. In a followup PR I'll change this to make the columns dynamic. It's not straightforward because of types.

src/lib/vocab/account/accountMutations.ts
```diff
@@ -1,19 +1,24 @@
...
+	return result;
+};
+
+export const UpdateAccountSettings: Mutations['UpdateAccountSettings'] = async ({invoke, ui}) => {
+	const result = await invoke();
+	if (!result.ok) return result;
+	const $account = result.value;
+	ui.account.set($account);
+	return result;
+};
```
ryanatkn, line 16: I changed this to match `UpdateAccountPassword`. The optimistic updates seem more trouble than they're worth in this case.

greatbacon (approved):

PR #532, opened 2022-10-26
Add role to assignment

greatbacon:
This PR has 2 main components
1.) This adds the `role_id` to the assignment table, now allowing for assignments to account for roles. A downstream implication of this is now personas may have multiple assignments within a community (formerly as a `membership` personas were only allowed 1 of these).

2.) This PR also adds front end updates for creating these tertiary assignments & updates the Home view to now dynamically display roles in a community and the personas who have been assigned them.

The ability to remove additional roles from personas will be added in a follow up PR

edit: In retrospect this probably should have been 2 PRS, I can split it apart if you would prefer 🙃


src/lib/ui/AssignmentManager.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+	// TODO speed this up with a better cached data structures
+	$: assignablePersonas = communityPersonas.filter(
+		(p) => !assignments.find((a) => a.get().persona_id === p.get().persona_id),
+	);
+
+	$: console.log(assignablePersonas.length);
+
+	const updateAssignedPersonas = (assignedPersona: Readable<Persona>) => {
+		assignablePersonas = assignablePersonas.filter((ap) => ap !== assignedPersona);
```
greatbacon, line 32: So I'm having some issues related to this store, I'm trying to update it as personas get assigned roles in this flow, but it seems like the array keeps getting repopulated before assigned items get removed and thus never gets any smaller.  Any thoughts?

greatbacon:
I've got a couple of other thoughts I'll leave here related to assignments. I don't think we necessarily need an answer today but we probably need something in the medium term.

1.) We don't currently have a process for "kicking" people from a community. This behavior could be encompassed by simply removing the last assignment from a persona. Or we could always block the last assignment from being removed & then add a dedicated kick button.
2.) If someone is kicked from a community then it's possible that someone down the road (or even immediately) could just invite them back in. We probably will want some mechanism for keeping a record of personas in a community context, even if they've been removed from the community. We might also want a dedicated/reserved `role` like `banned` or `kicked` to provide some handles or block toxic personas from sneaking back in.

src/lib/db/migrations/00036-add-roles-to-assignments.ts
```diff
@@ -0,0 +1,26 @@
...
+	await sql`
+    ALTER TABLE assignments
+    ALTER COLUMN role_id
+    SET NOT NULL
+  `;
+
+	await sql`
+		ALTER TABLE assignments
+		ADD UNIQUE (persona_id, community_id, role_id);
+	`;
```
ryanatkn, line 25: :ok_hand:

src/lib/ui/RoleItem.svelte
```diff
@@ -0,0 +1,42 @@
...
+</script>
+
+<li>
+	<span class="role-name">{$role.name}:</span>
+	<ul class="role-members">
+		{#if assignments.length > 0}
+			{#each assignments as assignment (assignment)}
+				<li><AssignmentItem {assignment} /></li>
+			{/each}
+		{:else}no assignments{/if}
```
ryanatkn, line 21:
```suggestion
		{#each assignments as assignment (assignment)}
			<li><AssignmentItem {assignment} /></li>
		{:else}
			no assignments
		{/each}
```

ryanatkn (approved):

src/lib/ui/AssignmentManager.svelte
```diff
@@ -0,0 +1,51 @@
...
+	export let community: Readable<Community>;
+	export let assignments: Array<Readable<Assignment>>;
+
+	$: selectedPersona = $personaSelection;
+
+	$: communityPersonas = $personasByCommunityId.get($role.community_id)!;
+
+	// TODO speed this up with a better cached data structures
+	$: assignablePersonas = communityPersonas.filter(
+		(p) => !assignments.find((a) => a.get().persona_id === p.get().persona_id),
```
ryanatkn, line 26:
```suggestion		(p) =>
			!assignments.some(
				(a) =>
					a.get().persona_id === p.get().persona_id &&
					a.get().community_id === p.get().community_id,
			),
```

src/lib/ui/AssignmentManager.svelte
ryanatkn: This is caused by the `dispatch.OpenDialog` breaking prop reactivity. Can be fixed here by doing the `$: assignments = $assignmentsByRoleId.get($role.role_id);` lookup in the `AssignmentManager` instead of making it a prop.

PR #534, opened 2022-10-27
add delete persona

ryanatkn:
Adds `DeletePersona` and the option to the acting persona contextmenu.

followup:

- evict related objects on the client -- assignments, spaces, entities/ties
- use `testDbCounts` in `servicesIntegration.test.ts` -- see #536
- cleanup: space cascade doesn't delete directories and orphaned entities, so we need to do that explicitly


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -89,6 +90,16 @@ export class EntityRepo extends PostgresRepo {
 		return OK;
 	}
 
+	async ghost(persona_id: number): Promise<Result<{value: number}>> {
```
ryanatkn, line 93: Is `anonymize` a better name for this method?

greatbacon, line 93: a little nitpicky, but can we make this name an explicit `ghostEntities`?

greatbacon (approved):
Had one nitpick about a name, otherwise great work! 💪

src/lib/vocab/entity/EntityRepo.ts
ryanatkn: good nitpick that's not a good name -- how about `attributeToGhost`? That'll be future-proof as we probably rename the `persona_id` to `actor` and sometimes will have `attributedTo` as the "acting as" secondary persona.

PR #536, opened 2022-10-28
add test helper for db table row counts

ryanatkn:
Adds the test helper `testDbCounts` for checking test cleanup. Usage:

```ts
const assertDbCounts = await testDbCounts(db);`
// ... do some stuff
await assertDbCounts(db); // throws if any table row counts differ than when instantiated
```

Example error (this is the current output from `servicesIntegration.test.ts`):

```
    Database row counts changed unexpectedly:

accounts: +2
personas: +2
communities: +2
roles: +2
assignments: +2
spaces: +12
entities: +13
```

Will use this in `servicesIntegration.test.ts` in a followup. Need #534 for that.


src/lib/vocab/entity/EntityRepo.test.ts
```diff
@@ -12,6 +12,30 @@ const test__EntityRepo = suite<TestDbContext & TestAppContext>('EntityRepo');
...
 test__EntityRepo.after(teardownDb);
 
+test__EntityRepo('create and delete entities', async ({db, random}) => {
+	const {
+		community,
+		account,
+		personas: [persona],
+		spaces: [space],
+	} = await random.community();
+	const assertDbCounts = await testDbCounts(db);
```
ryanatkn, line 22: This was the simplest test I could think of to demonstrate and test `testDbCounts` usage. Mixes concerns with what probably belongs in a new `testHelpers.test.ts` but I don't see a way to cleanly separate concerns, so it was less work to just add it here.

greatbacon: I think we can give ourselves a break in just implicitly testing our test helpers via other tests lol

greatbacon (approved):

PR #538, opened 2022-10-29
change client `assignments` array to a set and treat `Assignment`s as immutable

ryanatkn:
Updates some of the assignment-related patterns on the frontend with stash/evict. Changes the array of assignments to a set, towards more ergonomic+efficient patterns for our frontend data structures.

Fixes numerous eviction bugs.

Another change is simplifying the code under the assumption that `Assignment` objects are immutable. They have no `updated` or data field, and so they're never updated, only created/deleted. (CRD not CRUD?) Similar to `ties` this means we don't need store wrappers on the client.


vite.config.js
```diff
@@ -3,6 +3,7 @@ import {sveltekit} from '@sveltejs/kit/vite';
 /** @type {import('vite').UserConfig} */
 const config = {
 	plugins: [sveltekit()],
+	optimizeDeps: {exclude: ['@feltcoop/felt']},
```
ryanatkn, line 6: same fix as #547

greatbacon (approved):
🦔 💨

PR #542, opened 2022-11-01
improve form accessibility

ryanatkn:
Changes forms to use a `.title` instead of the placeholder for the name.

The layout should be improved with [this Felt update](https://github.com/feltcoop/felt/pull/254) and will need some more tweaking, but it'll help inform the standard form styles (we may have centered forms with a special class)


greatbacon (approved):

ryanatkn:
Made a bunch of changes since you approved, just applying the same pattern to the rest of the forms.

PR #545, opened 2022-11-03
improve client persona types

ryanatkn:
Fixes the client persona types, adding `ClientPersona`, `PublicPersona`, and `ActorPersona`. This touches a lot of code, but luckily it's mostly just types: for the main change see `$lib/vocab/persona/persona.schema.ts` and `$lib/vocab/persona/PersonaRepo.ts` 

Surfaced some issues where services returned data that should be private. The `PersonaRepo` methods should no longer leak private info.

`PublicPersona` includes only the fields that are expected on the client for non-session personas. (see below for a possible name change, "public" doesn't seem quite right)

`ClientPersona` is either a `PublicPersona` or `AccountPersona`, and `AccountPersona` is the same as a "session persona". We currently need no distinction between `AccountPersona` and the client session versions, but we may in the future want a `SessionPersona`. This is because semantically `AccountPersona` includes non-session personas on the server, so it's confusing to reuse that on the client to mean session-only. The fields may also diverge in the future.

`ActorPersona` is either an `AccountPersona` or a `CommunityPersona`. Basically, it just excludes the ghost. In the future this will allow users to perform actions as non-account personas: it was helpful to make the distinction through the types now, because it's a particular part of the API on the server side.

questions

- better name than `PublicPersona`? Isn't exactly public...it's the document that other people in your communities have access to. Maybe `Shared`?

followup:

- `actor` instead of `persona` when it's an `AccountPersona` on the client, as well as in the random helpers


src/lib/vocab/persona/persona.ts
```diff
@@ -32,4 +34,12 @@ export interface GhostPersona {
 }
 export type Persona = AccountPersona | CommunityPersona | GhostPersona;
 
+export interface PublicPersona {
+	persona_id: number;
+	type: 'account' | 'community' | 'ghost';
```
ryanatkn, line 39: This way of doing things won't work if we want to attach `community_id` for community personas. We don't want to add it for account personas, because that's private information. At the moment we can match the community persona name against the community name, but that won't work for future persona types that need to reference their `community_id`.

greatbacon: I think this is fine for now. When we come to an issue that requires a resolution, we can take the time to semantically break out the overloaded `community_id`. Maybe just leave a comment in the code to that effect?

greatbacon:
I think `PublicPersona` is fine; it's maybe not "on the street" in public but certainly like... "at a bar/theatre/club" public. If that makes sense?

If you want some alternatives possibly `GeneralPersona`, `CommonPersona`, or maybe `KnownPersona`?

Also, since this PR is a chonker & touches a lot of files, I recommend we merge & deploy it on it's own so we can really get it tested in staging in isolation before bringing any other changes into the mix.

src/lib/vocab/persona/persona.ts
ryanatkn: Good idea, added a todo linking back to this thread.

ryanatkn:
cool k `PublicPersona` for now, can always split it into more later

greatbacon (approved):

PR #548, opened 2022-11-08
navigate to root on signout

ryanatkn:
When signing out, this redirects the browser to the root path, `/`. I can see this being undesirable in some rare cases -- like you're looking at a particular space but want to switch accounts and look at the same space -- but in most cases, preserving the path will lead to a weird UX, showing accounts spaces they can't access.

Adds a helper `gotoUnlessActive` that's used here and in another place -- it's the same as SvelteKit's `goto` except it does nothing if the target path is already active. Otherwise, a janky double entry is added to the browser history. (kinda seems like this belongs upstream in SvelteKit, didn't see any discussion -- I can see the impl would be straightforward, might open an issue)


greatbacon (approved):

greatbacon:
Users logging out of one account & then into another to view the same space/community is undesired behavior (since they should be using personas) so I'm not going to loose any sleep over the friction in that path.

PR #549, opened 2022-11-09
improve env handling

ryanatkn:
changes:

- change env var `VITE_` prefix to `PUBLIC_` to match the SvelteKit changes -- this is required for the followups, we've been hacking around the SvelteKit behavior (which has changed a lot since I last looked). The env vars are now statically resolved, so we can now do some really cool dev-only stuff like combining dynamic imports and schema checks, so our production runtime wouldn't need the big schema checking lib
- remove some usage of `import.meta.env` for SvelteKit's `$app/environment` (finish in a followup, need changes to Gro)
- init env and sync SvelteKit in our version of `gro typecheck` -- SvelteKit generates types for our env vars, which means `svelte-kit sync` needs to run before any typechecking, else CI and typechecking a fresh repo fail -- luckily Gro's tasks make this easy
- remove unnecessary usage of `dotenv`
 
followup:

- remove remaining usage of `import.meta.env`
- use the SvelteKit `$env` imports and clean up `src/config.ts` (can we delete it altogether?)
- should we no longer use `process.env.NODE_ENV` on the server anywhere? `dev` from `$app/environment` instead?
- can we remove the remaining usage of `dotenv`?


src/.env.development
```diff
@@ -0,0 +1,3 @@
+COOKIE_KEYS=latestSecret__olderSecret__oldestSecret
```
greatbacon, line 1: Why do we have these new .env files in src? Does the server no longer read them from the root?

PR #552, opened 2022-11-09
Fix alter type issue

greatbacon:
This PR fixes an issue related to a Postgres constraint preventing the alteration of enumerated types inside of transaction blocks.

Fix is based off of this Stack Overflow post: https://stackoverflow.com/questions/53149484/error-alter-type-add-cannot-run-inside-a-transaction-block


ryanatkn (approved):
Yikes, do you know why I wouldn't have seen this while developing locally?

greatbacon:
You know, on second glance, this may have just been a me problem. That same stackoverflow mentions there was a fix in Postgres v12 (which I thought I had) but I think it's just my `psql` that's v12 & my actual DB is still on v11 🙃 

Let me check some other things real quick!

greatbacon:
Yup, that's exactly what it was, sorry for the panic pull request! 😞

PR #554, opened 2022-11-09
Split ghost persona transactions

greatbacon:
Okay, so different problem popped up when I tried to run the ghost migration in prod.

Got the error `unsafe use of new value "ghost" of enum type persona_type` which seems to be related to the previous hullabullaloo re: ALTER TYPES in TRANSACTIONS.

The solution though is to just commit the new type to the DB before using it in another query so that's what this does (I think...)

I'm not 100% sure but we probably didn't see it pop up locally since when running `lib/db/create` the new type is technically also included in the transaction.


ryanatkn (approved):
Ah ok, I think this fixes it. I see now that `gro lib/db/testMigration` does indeed fail locally but works with this PR, because seeding can't run until the ghost persona type has been added, and this splits things apart correctly.

PR #560, opened 2022-11-11
add @persona mentions

ryanatkn:
Bolds `@mentions` to personas and communities, and adds contextmenu functionality. The mentions never resolve to community personas -- they resolve to account personas and communities. (maybe both the community and its persona should be in the contextmenu?)

We may want to make the default action on the mentions -- clicking/tapping -- to open the contextmenu. Currently mentions behave as plain text plus the normal contextmenu.

I can see wanting to add a "Go to Community" entry to the contextmenu for non-active communities, and some equivalent to view a Persona's info. (maybe in a dialog?)


greatbacon (approved):
As a proof-of-concept I think this works fine as is. We should probably do a little planning about what behavior we actually want out of this, before futzing around with it more.

PR #561, opened 2022-11-12
add `Lists` view

ryanatkn:
questions:

- `mod` or `plugin` or `package` or `bundle` etc? (npm equivalent is "packages", cargo (Rust) is "crates") The word "mod" may be something we want to reserve for referring to moderators, to avoid confusion. (sadly, I like the aesthetics of mod)
- `$lib/mods/:user_name/:mod_name` namespacing? 
- name? `List` wouldn't have nested lists, would focus on individual items
- mod package interface?

```ts
// /src/lib/mods/ryanatkn/mod.ts
export default {name: '@ryanatkn/lists', views: ['Lists.svelte']};
// or maybe
export default {name: '@ryanatkn/lists', mods: [{type: 'view', path: 'Lists.svelte'}]};
```

followup:

- remove "checked" property
- inherit parent properties like `"checked"`


src/lib/vocab/view/view.ts
```diff
@@ -32,6 +32,7 @@ export const viewTemplates: ViewTemplate[] = [
 	{name: 'Iframe', view: '<Iframe />', icon: '💻'}, // TODO does this need a default `src`?
 	{name: 'EntityExplorer', view: '<EntityExplorer />', icon: '✏️'},
 	{name: 'Todo', view: '<Todo />', icon: '🗒'},
+	{name: 'Lists', view: '<Lists />', icon: '🍔'},
```
ryanatkn, line 35: got a better emoji? lol

ryanatkn:
thinking we should go with `mod` as the temporary name until we decide, since we know for sure we don't want that term

src/lib/vocab/view/view.ts
ryanatkn: I think I like:

# 🔨

changed the code to have it

greatbacon (approved):

PR #562, opened 2022-11-13
fix schema type to include the `name` property that we add

ryanatkn:
Fixes schema types to include the `name` property that we add to them, derived from `$id`, so `'/schemas/Foo.json'` has the name `'Foo'`. The data was already there, but now the types are correct.


src/lib/app/schemas.ts
```diff
@@ -29,6 +29,14 @@ import {RoleSchema} from '$lib/vocab/role/role.schema';
 // but this has the downside of losing the inferred default types,
 // which are handy when the schemas are used directly.
 
+export interface FeltVocabSchema extends VocabSchema {
```
ryanatkn, line 32: Instead of `FeltVocabSchema extends VocabSchema` it may be better to have `VocabSchema extends RawVocabSchema`

greatbacon (approved):

PR #568, opened 2022-11-14
improve entity editor

ryanatkn:
- extract `EditEntityForm`, following the patterns elsewhere
- add `legend` headings to the different `fieldset`s
- add controls to edit all properties of an entity whether or not they currently exist (not all properties supported yet -- started with `type`, `name`, and `content`
- add button to delete properties from an entity


src/lib/ui/PropertyEditor.svelte
```diff
@@ -90,39 +90,45 @@
...
-		<div class="preview markup panel">
-			<p>
-				{#if fieldValue}<pre>{serialized}</pre>{:else}<em>(empty)</em>{/if}
-			</p>
-		</div>
-	{/if}
+{#if value === undefined}
+	<div class="field">{field}</div>
+	<p>undefined</p>
+	<!-- TODO add a button to add/instantiate the field with some value -->
```
ryanatkn, line 96: This module is the same except for this `if` condition, checking for missing values. I think it'd be good to include a button here to instantiate the value, so I left a todo

greatbacon (approved):

PR #570, opened 2022-11-15
Query directory by entity

greatbacon:
Added 2 queries, a generic tie query to find all ties radiating outward from a single destination and a more specialized query that uses the same logic, but then filters for directory entities.

Sorry if this isn't 100% in line with what you were looking for, I couldn't remember the exact use case from our previous discussion.


src/lib/vocab/tie/TieRepo.ts
```diff
@@ -32,7 +32,25 @@ export class TieRepo extends PostgresRepo {
 			)
 			SELECT DISTINCT tie_id, source_id, dest_id, type, created FROM paths;
 		`;
-		log.trace('directory ties', ties);
+		log.trace('ties under source', ties);
+		return {ok: true, value: ties};
+	}
+
+	async filterByDestId(directory_id: number): Promise<Result<{value: Tie[]}>> {
```
ryanatkn, line 39: `directory_id` should be `dest_id` right?

ryanatkn (approved):
super cool, want to write some basic tests for them? I can if you'd prefer, or maybe I don't care enough idk, I might jump straight to using them

greatbacon:
Mostly just wanted to run 'em by you before writing tests lol. I get that 1 comment cleaned up and some tests written before merging.

ryanatkn:
the usecase we have is fixing the way entities update -- we'll use it to send changes for each directory to fix notifications

PR #571, opened 2022-11-15
upgrade deps

ryanatkn:
Upgrades to use `@feltcoop/util`, changes a million imports but that's all


package.json
```diff
@@ -29,16 +29,18 @@
   },
   "devDependencies": {
     "@feltcoop/eslint-config": "^0.2.1",
+    "@feltcoop/felt": "^0.43.0",
+    "@feltcoop/gro": "^0.67.3",
```
ryanatkn, line 33: changes felt and gro back to be dev deps -- the script at `infra/deployed.ts` isn't a task, so we don't need Gro for it, and Felt is no longer needed at runtime, because it'll be bundled by SvelteKit, and otherwise `@feltcoop/util` has all of the runtime deps we previously imported from Felt.

package.json
```diff
@@ -29,16 +29,17 @@
   },
   "devDependencies": {
     "@feltcoop/eslint-config": "^0.2.1",
+    "@feltcoop/gro": "^0.67.3",
```
ryanatkn, line 32: made gro a dev dep again, because we're not using `gro` on the server to run the migrate task, it's just a node script at `dist/infra/deployed.js`

greatbacon (approved):

PR #577, opened 2022-11-18
fix deploy migrations script

ryanatkn:
The post-deploy script tries to run migrations using a `src/`-prefixed path, which only works for development. This fixes the path so it works correctly in production.


src/infra/deploy.task.ts
```diff
@@ -54,9 +54,11 @@ export const task: Task = {
...
 			cd ${deployDirname};
 			tar -xvf ${artifactFilename};
-			echo 'npm ci';
-			npm ci;
-			cd -;
+			echo 'moving node_modules';
+			mv ~/${currentDeploy}/node_modules ./ && echo 'reusing node_modules' || echo 'no node_modules to reuse';
+			echo 'npm i';
+			npm i;
+			cd ~;
```
ryanatkn, line 61: My VPS kept timing out on `npm ci`, this seems to fix it but not sure about it being good long term

src/infra/deployed.ts
```diff
@@ -9,11 +10,21 @@ import {migrate} from '$lib/db/migrate';
...
-	});
+const main = async () => {
+	// TODO remove this after running on production once
+	const data = await db.sql`SELECT id, name FROM migrations`;
+	for (const d of data) {
+		if (d.name.endsWith('.ts')) {
+			// eslint-disable-next-line no-await-in-loop
+			await db.sql`UPDATE migrations SET name=${d.name.slice(0, -3) + '.js'} WHERE id=${d.id}`;
+		}
+	}
```
ryanatkn, line 21: This stuff is only needed because we previously ran production migrations with `.ts` files, and `ley` is sensitive to file extensions. It'd be easy to patch ley to not be, but that would break its compatibility with all existing databases, so it's unlikely to be accepted lol

src/infra/.env.default
```diff
@@ -1 +1 @@
-PUBLIC_HELP_EMAIL_ADDRESS=TODO@TODO.com
+PUBLIC_ADMIN_EMAIL_ADDRESS=TODO@TODO.com
```
ryanatkn, line 1: Be sure to update your env vars locally -- will think about what it would take to automate this

src/infra/deployed.ts
ryanatkn: @greatbacon fixed this directly on prod, so removing it now

src/infra/deployed.ts
```diff
@@ -1,19 +1,21 @@
 import {Logger} from '@feltcoop/util/log.js';
 
 import {migrate} from '$lib/db/migrate';
+import {db} from '$lib/db/db';
 
 // This script is designed to run after updating a production deployment.
 // The filename needs to stay in sync with this constant: `DEPLOYED_SCRIPT_PATH`
 
-// TODO consider making it a task instead (need changes to Gro to run js tasks directly)
```
ryanatkn, line 8: This is not a task -- we've managed to avoid using Gro as a production dependency, which is good, the fewer moving pieces the better

src/infra/deploy.task.ts
greatbacon: I think `--unclean` would be my preferred flag

greatbacon (approved):

src/infra/.env.default
greatbacon: Yea, especially because "us" isn't always going to us lol

PR #579, opened 2022-11-18
add admin community setting `instance.allowedAccountNames`

ryanatkn:
Adds `instance.allowedAccountNames` to the admin community settings. If the property exists, it's used to deny signups as an allowlist.

followup;

- move this data to the entity system so it can be edited and used like any other data


src/lib/vocab/account/accountServices.ts
```diff
@@ -35,6 +36,19 @@ export const SignUpService: ServiceByName['SignUp'] = {
 				return {ok: false, status: 409, message: 'account already exists'};
 			}
 
+			// check `instance.allowedAccountNames`
+			// TODO does this belong in `checkAccountName` above?
+			// TODO consider `const settings = repos.entity.filterByUrl('/instance');` (but scoped to admin?)
+			// should entities be scoped?  or /e/ to reference any url or id?
+			const adminCommunity = unwrap(await repos.community.findById(ADMIN_COMMUNITY_ID));
+			if (!adminCommunity) throw Error();
```
ryanatkn, line 44: Might want to return an `UNKNOWN_ERROR` or something here, something with status `500` and the default unknown message, which is equivalent to throwing like this but more explicit.

greatbacon: We should probably put an error log message here too with a more explicit message.

greatbacon (approved):

PR #581, opened 2022-11-19
show advanced info on edit entity form

ryanatkn:
Adds components to show source and dest entities related to an entity. Currently shows it on the "Edit Entity" form.

Also makes the `PropertyEditor` handle `deletable` fields.

followup:

- show the "advanced" info on the other editor components (and follow the *Form naming pattern)


src/lib/ui/PropertyEditor.svelte
```diff
@@ -17,10 +17,9 @@
 
 	export let value: TValue;
 	export let field: string;
-	export let update: (
-		updated: TValue,
-		field: string,
-	) => Promise<Result<unknown, {message: string}>>;
+	export let update:
+		| ((updated: TValue, field: string) => Promise<Result<unknown, {message: string}>>)
+		| null = null;
```
ryanatkn, line 22: One simple change to this component that causes a lot of apparent churn -- making `update` nullable and default to `null`. This enables showing values without making them editable with a minimal API, and reusing the same component, even though Property*Editor* is now a bit misleading.

src/lib/app/contextmenu/EntityContextmenu.svelte
```diff
@@ -20,7 +20,7 @@
 	<svelte:fragment slot="icon">
 		<UnicodeIcon icon="~" />
 	</svelte:fragment>
-	Entity
+	<small>{$entity.data.type || 'Entity'}</small><code>~{$entity.entity_id}</code><small />
```
ryanatkn, line 23: Not sure about this, mostly playing around to see where it could go.

ryanatkn: k undid those things - https://github.com/feltcoop/felt-server/pull/581/commits/43ea7424f43c7a8fe009d23a1f8ef74c2cd3ba9f

yeah URLs is the starting point, it's an idea we can punt, bringing data/objects at the system level into the namespace with a convenience shorthand (similar to how @ and # can be seen as shorthand for other kinds of objects)

greatbacon (approved):

PR #583, opened 2022-11-20
Read policy UX

greatbacon:
Adds a rough-draft first pass at displaying a given role's policies via the `manage role` dialogue.

This includes stubbing out the first `list` of permissions & using that to generate the display items. Since we currently aren't guaranteed to have a policy for every given permission we have to do some kind of wonky stuff to properly display stuff.


src/lib/ui/PolicyItem.svelte
```diff
@@ -0,0 +1,20 @@
...
+</script>
+
+<!-- TODO possibly remove the `ul` wrapper and change the `li`s to `div`s -->
+<ul>
+	{#if $policy}
+		✅ {$policy.permission}
+	{:else}
+		❌ {permission}
+	{/if}
+</ul>
```
ryanatkn, line 20:
```suggestion
<div>
	{#if $policy}
		✅ {$policy.permission}
	{:else}
		❌ {permission}
	{/if}
</div>
```

ryanatkn, line 20: some cleanup:

```suggestion
<div>
	{#if $policy}
		✅ {$policy.permission}
	{:else}
		❌ {permission}
	{/if}
</div>
```

ryanatkn (approved):

PR #589, opened 2022-11-22
Add policy toggles

greatbacon:
This PR adds UX for toggling a permission associated with a Role.

This is the last piece of "foundational" governance work btw! 🎉 

Once we have this in place, we can start to go through a list of "permissions" and begin implementing/granting them!!!!!!!!

edit: as an aside, this current design (if policy is present, permission is granted // if missing, permission is denied) could result in a lot of DB churn, and I think once we move on to utilizing the more dynamic `data` field just updating a field in that will probably be better in the longer run.


ryanatkn (approved):
wooot, amazing!!

PR #590, opened 2022-11-23
fix delete persona client mutation

ryanatkn:
Deleting a persona works on the backend, but puts the client into a bad state until refreshed.

- properly evict session personas
- change a component to handle null values, because it seems it errors because of a session persona that's mid-eviction (I'm not sure why it's needed, ideally this is what the mutation batching avoids, I'll investigate -- it may be caused by the mutation helpers using promises, and when the UI rerenders during the mutation batch for other reasons it fails to be in a good state -- so my guess is something else is causing a rerender during the mutation, and we may need to go with deferred async effects instead of waiting on promises in our mutation functions; mutations would need to be synchronous)


src/lib/ui/CommunityNavGroup.svelte
```diff
@@ -16,28 +16,30 @@
...
-	/>
-	{#each standardCommunities as community (community)}
+{#if personalCommunity && standardCommunities}
+	<div class="persona-group" use:contextmenu.action={[[ActingPersonaContextmenu, {persona}]]}>
 		<CommunityNavItem
-			{community}
+			community={personalCommunity}
 			{persona}
-			selected={persona === selectedPersona && community === selectedCommunity}
+			selected={persona === selectedPersona && personalCommunity === selectedCommunity}
```
ryanatkn, line 32: this diff is very misleading, all I did was nest things in the `#if` above

src/lib/ui/CommunityNavGroup.svelte
```diff
@@ -16,28 +16,30 @@
...
-		.get(persona)!
-		.find((c) => c.get().type === 'personal')!;
+		.get(persona)
+		?.find((c) => c.get().type === 'personal');
 
 	$: standardCommunities = $communitiesBySessionPersona
-		.get(persona)!
-		.filter((c) => c.get().type !== 'personal')!;
+		.get(persona)
+		?.filter((c) => c.get().type !== 'personal');
```
ryanatkn, line 24: just changing this component to handle nullability

greatbacon (approved):

PR #593, opened 2022-11-27
add client session schema

ryanatkn:
Fills in one of the gaps we had in our schemas. This changes `ClientSession`, the data returned on signup/signin and page load when logged in, to be schema driven, which means the server will now validate response data, so we can't accidentally send private data to clients in the session data.


greatbacon (approved):
🧹 🧹 🧹

PR #595, opened 2022-11-28
add `InviteToCommunity` event

ryanatkn:
Adds the `InviteToCommunity` event to replace the current strategy of creating an assignment for a persona directly.

The problem it solves is that the client previously depended on having all personas loaded to invite anyone -- this change makes inviting depend only on the persona name, not id.

The behavior is otherwise the same as creating an assignment directly, but it gives us a semantic distinction between having a persona added to a community, and having an existing member get a new assignment.

Followup:

- remove `persona.getAll()` (use what instead? `repos.persona.filterByAccountCommunities`?)


src/lib/vocab/community/communityServices.ts
```diff
@@ -185,6 +188,40 @@ export const DeleteCommunityService: ServiceByName['DeleteCommunity'] = {
...
+			const {actor, community_id, name} = params;
+
+			const community = unwrap(await repos.community.findById(community_id));
+			if (!community) {
+				return {ok: false, status: 404, message: 'no community found'};
+			}
+
+			const persona = unwrap(await repos.persona.findByName(name));
+			if (!persona) {
+				return {ok: false, status: 404, message: 'no persona found'};
```
greatbacon, line 206: Should we include the name provided in this error message? I think that could be useful immediate feedback for the user if they fatfinger the input or something.

greatbacon (approved):
Awesome! I think this'll set us up well re: permission work as well (since you might want more people to be able to invite to a community then handing out new assignments)

src/lib/vocab/community/communityServices.ts
ryanatkn: good idea, changed to `cannot find a persona named ${name}`

ryanatkn:
Nice yeah, that's great. Wondering if we want a different name -- currently `InviteToCommunity`, could be `CreateCommunityInvite` or `CreateCommunityInvitation` or ?

PR #596, opened 2022-11-28
Add replies to chat

greatbacon:
This PR adds a button for creating replies to messages in the `Chat` view

It's still somewhat a work in progress; 

- [x] the UX could use some a11y tweaking 
- [x] it looks like there might be a bug in the JSON parser? The Tie `type` is getting mangled in the Websocket client (all the Ties are ending up with the same type).


ryanatkn:
Maybe `ReplyChat` could be a good view name? We may just want to avoid collisions for now to avoid namespacing -- ultimately I think instances will just have their own mapping of component names to implementations, so `Chat` could mean different things on different instances, but always means one thing on a particular instance.

greatbacon:
> Maybe `ReplyChat` or `ChatWithReplies` could be a good view name? I'm thinking we should avoid naming collisions for now to keep things simple. (so "Chat" will always mean `$lib/plugins/feltcoop/chat` aka `@feltcoop/felt-plugin-chat` 

Yeah, I like ReplyChat & then I'll dump it in my plugins dir for now

ryanatkn (approved):

ryanatkn:
# ↳↳↳↳

PR #600, opened 2022-11-28
remove persona.getAll

ryanatkn:
Removes `repos.persona.getAll` and adds `repos.persona.filterByAccountCommunities`, following up to #595 that made it no longer necessary for the client to access all personas on the entire instance.


src/lib/vocab/persona/PersonaRepo.ts
```diff
@@ -70,6 +70,15 @@ export class PersonaRepo extends PostgresRepo {
 		return {ok: true, value: data};
 	}
 
+	async filterByAccountCommunities(account_id: number): Promise<Result<{value: PublicPersona[]}>> {
```
ryanatkn, line 73: not sure of this name

greatbacon (approved):

src/lib/vocab/persona/PersonaRepo.ts
ryanatkn: I'm liking associates right now

PR #603, opened 2022-11-29
Tweak env task

greatbacon:
This PR tweaks the updateEnv task to rely on a  git sub-command for generating the hash instead of reading the file directly, which is a deprecated behavior and was causing some issues on my machine.


ryanatkn (approved):
# 🛠️

src/lib/infra/updateEnv.task.ts
```diff
@@ -1,15 +1,21 @@
...
+
+		let gitHashContents = '';
+		const buildResult = spawnProcess('git', ['show-ref', '-s', `${branch}`], {stdio: 'pipe'});
+		buildResult.child.stdout?.on('data', (data: Buffer) => {
+			gitHashContents = data.toString().trim();
+		});
+		await buildResult.closed;
+
+		const currentEnvContents = await fs.readFile(envFile, 'utf8');
 		const gitHash = gitHashContents.trim().substring(0, 7);
```
ryanatkn, line 19: you may want to handle the case where it's `''` here, not sure what to default to

src/lib/infra/updateEnv.task.ts
```diff
@@ -1,15 +1,21 @@
...
 		const branch = (await fs.readFile('.git/HEAD', 'utf8')).trim().substring(5);
-		const [gitHashContents, currentEnvContents] = await Promise.all([
-			fs.readFile('.git/' + branch, 'utf8'),
-			fs.readFile(envFile, 'utf8'),
-		]);
+
+		let gitHashContents = '';
+		const buildResult = spawnProcess('git', ['show-ref', '-s', `${branch}`], {stdio: 'pipe'});
+		buildResult.child.stdout?.on('data', (data: Buffer) => {
+			gitHashContents = data.toString().trim();
```
ryanatkn, line 14: doubt it'd ever do this, but do you know if this could ever come as multiple buffer chunks? If so I think you want `+=`

src/lib/infra/updateEnv.task.ts
```diff
@@ -1,15 +1,21 @@
...
 	run: async ({fs, log, dev}) => {
 		const envFile = dev ? ENV_FILE_DEV : ENV_FILE_PROD;
 		const branch = (await fs.readFile('.git/HEAD', 'utf8')).trim().substring(5);
-		const [gitHashContents, currentEnvContents] = await Promise.all([
-			fs.readFile('.git/' + branch, 'utf8'),
-			fs.readFile(envFile, 'utf8'),
-		]);
+
+		let gitHashContents = '';
+		const buildResult = spawnProcess('git', ['show-ref', '-s', `${branch}`], {stdio: 'pipe'});
```
ryanatkn, line 12:
```suggestion
		const buildResult = spawnProcess('git', ['show-ref', '-s', branch], {stdio: 'pipe'});
```

I think this is right?

PR #604, opened 2022-11-30
fix bug with missing persona in view context

ryanatkn:
Adds `lookupPersona` which falls back to a valid persona object, currently the ghost.

- should `FALLBACK_PERSONA` always be the ghost? `GHOST_PERSONA`? (answer: yes)


greatbacon:
I think this is related to the changes from getAllPersonas() to the associate filter. There's probably an entity in the pile that references a persona that was removed from the community, I'm having similar issues.

src/lib/vocab/persona/personaHelpers.ts
```diff
@@ -42,3 +46,17 @@ export const checkPersonaName = (name: string): string | null => {
 	}
 	return null;
 };
+
+// TODO get from ui? or have it hardcoded included its `created`?
+export const $FALLBACK_PERSONA: ClientPersona = {
```
greatbacon, line 51: All users should have the ghost persona in their session, this hardcoded response feels like a bug waiting to happen 😬

greatbacon (approved):
I think the fallback persona always being the ghost works well for now. This gives us a place to expand the system if/when we get around to snapshots of left/kicked personas.

ryanatkn:
Thank you yeah, obvious in hindsight: https://github.com/feltcoop/felt-server/pull/604/commits/5534e9e1bde5e88de999a6e8aadc3e577f721cc4

PR #606, opened 2022-11-30
add `direction` prop to `List` view

ryanatkn:
Adds a prop option to the `List` view that makes it flow in the opposite direction.

```svelte
<List />
```

is the same as:

```svelte
<List direction="down" />
```

and this option swaps the input element below the items:

```svelte
<List direction="up" />
```

- [x] might want to think about the API for reversing the order of the message list items before merging (because "direction" might be the wrong name when considering it) (going with `layoutDirection` and `itemsDirection`, which are both `flex-direction` values)
- [x] should we make a default space with this so there's a documented demo? or how else could an example of this live? (yes in a followup)

View for 2 spaces to demo the new stuff:

```svelte
<code>List</code><List /><div><code>List</code> with <code>layoutDirection="column-reverse"</code></div><List layoutDirection="column-reverse" /><div><code>List</code> with <code>layoutDirection="row"</code></div><List layoutDirection="row" /><div><code>List</code> with <code>layoutDirection="row-reverse"</code></div><List layoutDirection="row-reverse" />
```

```svelte
<div><code>List</code> with <code>itemsDirection="column-reverse"</code></div><List itemsDirection="column-reverse" /><div><code>List</code> with <code>itemsDirection="row"</code></div><List itemsDirection="row" /><div><code>List</code> with <code>itemsDirection="row-reverse"</code></div><List itemsDirection="row-reverse" /><div><code>List</code> with <code>layoutDirection="row-reverse" itemsDirection="row-reverse"</code></div><List layoutDirection="row-reverse" itemsDirection="row-reverse" />
```

followup:

- use that code to seed the docs/examples community


greatbacon (approved):
Could "flow" be a better word for the API?

ryanatkn:
yeah hm there's like a "layoutFlow" and an "itemsFlow" I think

PR #610, opened 2022-12-01
Fix new filterAssociate query for ghost

greatbacon:
The new filterAssociates query was missing the system `ghost` persona and could miss the `admin` persona (if a user is not in the admin community). This adds the two to the query plus a quick test to check for it.


ryanatkn (approved):
nice good catch and good tests

PR #614, opened 2022-12-03
add `CreateEventForm` and its contextmenu entry

ryanatkn:
Adds the `CreateEventForm` component and contextmenu entry to open it. The idea is that we should be able to use our event schemas to automatically create a dynamic generic form to instantiate the params for an event -- and then using those params, either dispatch the event or save them off into an entity for later. This is groundwork for the ability to do things like "run this script if the vote passes" - one or more events can be a script, and the data for a dispatchable event is just a name+params (JSON).

- [x] blocked by Felt support for multiple stacking dialogs (pickers don't work, we know we needed this anyway): https://github.com/feltcoop/felt/pull/268

future

- add links to the docs
- work with the response data
- implement the pickers for each data type (and probably don't use the `PropertyEditor`
- create events without dispatching them? save them where? manage them like any other collection, but how should that look?
- figure out if we should include other details besides `params` like entity `name` and `content` and `summary` -- probably so if we create entities to hold the event data


src/lib/app/eventTypes.ts
```diff
@@ -256,42 +265,22 @@ export type SignOutResponse = null;
...
-}
-export interface UpdateAccountSettingsResponse {
-	account_id: number;
-	name: string;
-	settings: {
-		darkmode?: boolean;
-	};
-	created: Date;
-	updated: Date | null;
+	settings: AccountSettings;
```
ryanatkn, line 268: cleaning up some stuff by using `tsType` in the remaining schemas with `$ref`s

greatbacon (approved):
yessssss

PR #615, opened 2022-12-03
paginate chat

ryanatkn:
Adds a new paginated query system and implements it just for the `Chat`.

Future:

- fix the problem detailed in the todos
- implement for other views
- merge behavior with `QueryEntities`
- autoload on scroll? (need to handle the case where you can't scroll because not enough loaded to cause a scrollbar)


src/lib/util/api.ts
```diff
@@ -0,0 +1,25 @@
+import type {Result} from '@feltcoop/util';
```
ryanatkn, line 1: not sure about this module name/location but it seems better than `$lib/server/api`?

src/lib/vocab/entity/entityMutations.ts
```diff
@@ -61,11 +61,27 @@ export const QueryEntities: Mutations['QueryEntities'] = ({ui: {queryByKey}, dis
...
+				} else {
+					query.status.set('failure');
+					query.error.set(result.message);
+				}
+			},
+			() => {
+				if (!query) return;
+				query.status.set('failure');
+				query.error.set('unknown error');
+			},
```
ryanatkn, line 84: this is unrelated to the PR, it fixes error handling of the existing query system that doesn't try to handle pagination

greatbacon (approved):

PR #616, opened 2022-12-04
add `EmptyPath` to let user create a space or an entity

ryanatkn:
This PR has a conceptual step forward in its treatment of empty paths as potentially being filled with a space or entity. We're talking about potentially adding path information to every entity, which would include every directory and therefore space. This PR treats an empty path as a blank page for the user to fill -- so you can now link to things like /foo and immediately see what's there, if anything, and create something if not.

The user was previously presented with only a `CreateSpaceForm.svelte` at these empty paths -- now with `EmptyPath.svelte` you can toggle to create an entity.

- [x] implement `EmptyPath`
- [x] add `CreateEntityForm` to a second tab
- [x] cleanup some stuff

questions

- what about assigning an existing entity this path? maybe a followup, but think it through first

followup:

- prefill entity `path`
- add `type` to `CreateEntityForm`
- query if an entity exists at the path before considering it empty, and if one does exist, show the `EditEntityForm`


src/lib/ui/CreateEntityForm.svelte
```diff
@@ -25,11 +22,6 @@
 	export let type = 'Collection';
 	export let fields: {name?: boolean; content?: boolean} = {name: true, content: true}; // TODO add customization like display names for each field
 
-	$: selectedPersona = $personaSelection;
-	$: persona_id = $selectedPersona!.persona_id;
-	$: selectedSpace = $spaceSelection;
-	$: source_id = $selectedSpace!.directory_id;
```
ryanatkn, line 31: usage of `selected` ui data directly is often sus, we should normally prefer props for flexibility

greatbacon (approved):

PR #617, opened 2022-12-04
Add first policyCheck

greatbacon:
This code introduces the `policyCheck` helper & uses it to check if an actor in a given community has the right policy for UpdateCommunitySettings


src/lib/vocab/policy/permissions.ts
```diff
@@ -1 +1,4 @@
-export const PERMISSIONS: string[] = ['UpdateCommunitySettings', 'DeleteCommunity'];
+export enum PERMISSIONS {
+	UPDATE_COMMUNITY_SETTINGS = 'UpdateCommunitySettings',
+	DELETE_COMMUNITY = 'DeleteCommunity',
+}
```
ryanatkn, line 4:
```suggestion
export type PermissionName = keyof typeof permissions;

export const permissions = {
	UpdateCommunitySettings: 'UpdateCommunitySettings',
	DeleteCommunity: 'DeleteCommunity',
};
```

`enum` is something we've avoided so far and I think we should continue -- it's pretty common to do so in the TS world. There's too many caveats and weird things for a little sugar. (the one exception is `@feltcoop/util`'s `LogLevel`, but that's because we leverage the enum integer for the log check, basically a hack) Instead we can use either:

- objects with data (preferred in this case, we're likely to include schema information in the future, not just a string name)
- or a union of strings, e.g. `export type Permission = 'UpdateCommunitySettings' | 'DeleteCommunity'` (which doesn't make sense in this case because we know we'll need additional data with them)

By choosing the first option we can compute the union of strings, which is what `PermissionName` is.

src/lib/util/randomEventParams.ts
```diff
@@ -67,9 +68,11 @@ export const randomEventParams: RandomEventParams = {
 		if (!persona) ({persona} = await random.persona(account));
 		return randomCommunityParams(persona.persona_id);
 	},
-	UpdateCommunitySettings: async (random, {account, persona, community} = {}) => {
+	UpdateCommunitySettings: async (random, {account, persona, community, authorize} = {}) => {
 		if (!persona) ({persona} = await random.persona(account));
 		if (!community) ({community} = await random.community(persona, account));
+		if (authorize)
+			await random.policy(community, persona, account, PERMISSIONS.UPDATE_COMMUNITY_SETTINGS);
```
ryanatkn, line 75: This is clever, I'm trying to think about what the cleanest solution would be if we could redesign things.

src/lib/util/randomVocab.ts
```diff
@@ -105,10 +106,14 @@ export const randomRoleParams = (actor: number, community_id: number): CreateRol
...
-export const randomPolicyParams = (actor: number, role_id: number): CreatePolicyParams => ({
+export const randomPolicyParams = (
+	actor: number,
+	role_id: number,
+	permission?: PERMISSIONS,
+): CreatePolicyParams => ({
 	actor,
 	role_id,
-	permission: randomPermissionName(),
+	permission: permission || randomPermissionName(),
```
ryanatkn, line 116: you can implement `randomPermissionName` above, like `randomItem(Object.keys(permissions))`

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	const policy = unwrap(
+		await repos.policy.filterByActorCommunityPermission(actor_id, community_id, permission),
+	);
+
+	if (policy.length === 0) {
+		log.trace('no policy present for actor in community', actor_id, community_id);
+		return {ok: false, message: 'Provided actor does not have permission'};
+	}
+	return {ok: true, value: true};
```
ryanatkn, line 31: see above

```suggestion
	return {ok: true};
```

src/lib/util/randomEventParams.ts
greatbacon: I implemented the authorize block just to leave an option for testing fail states down the road, although I suppose assuming permission by default and just manipulating the DB to get to that state would be a better path. 😁 

I agree that having some initial set of permissions to rely on for authorization is the better path forward. I wanted more permissions in place though before digging into "defaults" though.

If I rip out the `authorize` stuff & just leave a //TODO above the policy creation does that work for you for now?

src/lib/util/randomVocab.ts
```diff
@@ -48,7 +50,7 @@ export const randomView = (): string => '<Chat />';
 export const randomEntityData = (): EntityData => ({type: 'Note', content: randomString()});
 export const randomTieType = randomString;
 export const randomRoleName = randomString;
-export const randomPermissionName = randomString;
+export const randomPermissionName = randomItem(Object.keys(permissions));
```
ryanatkn, line 53:
```suggestion
export const randomPermissionName = (): PermissionName => randomItem(Object.keys(permissions));
```

src/lib/util/randomEventParams.ts
ryanatkn: Yep ripping out `authorize` sounds best for now, keeps things simple and we can solve the other problems when we need them

src/lib/vocab/policy/permissions.ts
```diff
@@ -1 +1,5 @@
-export const PERMISSIONS: string[] = ['UpdateCommunitySettings', 'DeleteCommunity'];
+export type PermissionName = keyof typeof permissions;
+export const permissions = {
+	UpdateCommunitySettings: 'UpdateCommunitySettings',
+	DeleteCommunity: 'DeleteCommunity',
+};
```
ryanatkn, line 5:
```suggestion
};
export const permissionNames: PermissionName[] = Object.keys(permissions);
```

src/lib/ui/RoleDetails.svelte
```diff
@@ -87,7 +87,7 @@
 	<h2>Permissions</h2>
 	<div class="policies">
 		{#if policies}
-			{#each PERMISSIONS as permission (permission)}
+			{#each Object.values(PERMISSIONS) as permission (permission)}
 				<PolicyItem actor={persona} {role} {permission} policy={policies.get(permission)} />
```
ryanatkn, line 91:
```suggestion
			{#each permissionNames as permissionName (permissionName)}
				{@const permission = permissions[permissionName]}
				<PolicyItem actor={persona} {role} {permission} policy={policies.get(permission)} />
```

ryanatkn (approved):

ryanatkn:
Had some minor convenience changes and agree, remove the `authorize` hack and punt that issue, assume security in tests

PR #618, opened 2022-12-04
add `OrderedList` view

ryanatkn:
Adds the `OrderedList` by copypasting `$lib/plugins/ryanatkn/list`. It's a usecase that I expect will drive a few prereq PRs.

The idea is that the items are expected to have a settable and persisted order, and it'll need different UI for interacting/viewing that aspect of the data. The collection of items could potentially be viewed with a different sorting algorithm, but the focus of the view is on collections where the order is meaningful, so that's the default experience that it's designed for -- it's related to the `OrderedCollection` in ActivityStreams: https://ryanatkn.github.io/corpus-activity-streams/#OrderedCollection


src/lib/vocab/space/defaultSpaces.ts
```diff
@@ -25,6 +25,7 @@ export const toDefaultSpaces = (actor: number, hub: Hub): CreateSpaceParams[] =>
 				{...toTemplatePartial('Todo'), name: 'todo', path: '/todo'},
 				{...toTemplatePartial('List'), name: 'list', path: '/list'},
 				{...toTemplatePartial('Lists'), name: 'lists', path: '/lists'},
+				{...toTemplatePartial('OrderedList'), name: 'ordered-list', url: '/ordered-list'},
```
ryanatkn, line 28:
```suggestion
				{...toTemplatePartial('OrderedList'), name: 'ordered-list', path: '/ordered-list'},
```

PR #621, opened 2022-12-06
fix `ReplyChat` button size and focus input on button click

ryanatkn:
Tweaks `ReplyChat` in a couple ways:

- fixes the problem we had with setting the reply button size -- it now correctly uses `--icon_size_sm` by setting `--input_height`, which controls the `min-height` and `min-width` for buttons and `.plain-button` -- not the best interface, but it has made sense to standardize input heights across the app including buttons, we now just have a usecase for the smaller version
- when the reply button is clicked, it focuses the main input (whether or not there's not a selected reply -- I originally had it only on open, but on close you probably want to chat most of the time?)


greatbacon (approved):
schweet

PR #622, opened 2022-12-06
rename space path from url

ryanatkn:
`url` was never technically correct as a space property -- we always considered it to be a path relative to a community. This renames `space.url` to `space.path` -- the data is still correct (we don't have validation around it yet, it's on the todo list)


.github/workflows/check.yml
```diff
@@ -42,4 +42,3 @@ jobs:
       - run: npx gro lib/db/create
       - run: npx gro check
       - run: npx gro build
-      - run: npx gro lib/db/testMigration
```
ryanatkn, line 45: had to remove this :[

PR #623, opened 2022-12-06
add mention icons and add inline avatars

ryanatkn:
- add `inline` prop to `Avatar`, `PersonaAvatar` , `CommunityAvatar` , and `EntityIcon`, something we've needed for a while, and this design feels good (defaults to `false`, preserving existing behavior)
- render persona/community icons for `Mention`s when available (this is where we needed `inline`)
- upgrade deps (svelte extension has been crashing in vscode...)


greatbacon (approved):
slick

PR #626, opened 2022-12-06
fix notifications

ryanatkn:
- [x] remove the current strategy
- [x] include changed directories in the response of `CreateEntity`
- [x] test that the first of `CreateEntity` response `entities` is the created one

Not including `UpdateEntity` and `DeleteEntities` in changes right now. Thinking it'll be too noisy to be a good UX (we talked about `UpdateEntity` but I think the same applies to `DeleteEntity`, typically you'll expect new content with a notification, and I think we should lean towards minimizing false positives and accepting more false negatives). We'll want to revisit this UX later.


greatbacon (approved):
Approved pending that test!

PR #627, opened 2022-12-06
add example community seed data

ryanatkn:
Adds the `example` community and two spaces with hardcoded data, `list1` and `list2`, for demonstrating different ways of using `List`. It acts as both docs and manual tests -- we'll commonly want to do this with views as we add props to customize them.

Also improves some of the seed data code, and extracts `$lib/app/templates`. The code is basically fine but I'm not sure where it all belongs and what it's named. `Template` may not even be the right word -- is it the same as the templates we expose to users when they create communities? Or are they different interfaces? The goal was to make them the same, but we'll see if that works.


src/lib/db/seed.ts
```diff
@@ -84,24 +84,32 @@ export const seed = async (db: Database, much = false): Promise<void> => {
...
-		{name: 'felt', actor: mainPersonaCreator.persona_id},
-		{name: 'dev', actor: mainPersonaCreator.persona_id},
-		{name: 'backpackers-anonymous', actor: mainPersonaCreator.persona_id},
-	];
+	let personaIndex = -1;
+	const nextPersona = (): AccountPersona => {
+		personaIndex++;
+		if (personaIndex === personas.length) personaIndex = 0;
+		return personas[personaIndex];
+	};
```
ryanatkn, line 104: thinking this might be worth extracting into a helper, `nexter`?

src/lib/db/seed.ts
```diff
@@ -227,6 +242,48 @@ const SEED_BY_VIEW_NAME: Record<string, (ctx: SeedContext) => Promise<void>> = {
 	},
 };
 
+// TODO extract these interfaces -- where?
+// TODO `Template` may be the wrong word here
+export interface CommunityTemplate {
```
ryanatkn, line 247: Is `Template` the right word to be using here? I'm not sure if this is the same user-facing data structure for defining templates for communities. Maybe it should be, i guess I see no reason to make 2 systems.

src/lib/app/templates.ts
```diff
@@ -0,0 +1,32 @@
+import type {EntityData} from '$lib/vocab/entity/entityData';
+import type {CreateSpaceParams} from '$lib/app/eventTypes';
+
+// TODO where does this belong? vocab?
+
+// TODO `Template` may be the wrong word here
+export interface CommunityTemplate {
+	name: string;
+	spaces?: SpaceTemplate[];
+	// TODO roles: RoleTemplate[{name, creator, default}];
```
ryanatkn, line 10: this is per our other discussion

src/lib/util/nexter.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Returns a function that returns the next item in the `array`
+ * in a linear sequence, looping back to index 0 when it reaches the end.
+ * @param array
+ * @returns A function that returns the next item in the `array`
+ */
+export const nexter = <T>(array: T[]): (() => T) => {
```
greatbacon, line 7: is this worth extracting to a util if it's only being used in 1 place?

ryanatkn: It's something I've needed a lot with procedural generation, hence the effort to document it - I was going to upstream to `@feltcoop/util` -- thought I had a TODO for it, might have been something else.

greatbacon (approved):
Templates seems like the right word for now, if we find it gets confusing we can refactor things out into separate terms.

PR #630, opened 2022-12-07
Add two new permissions

greatbacon:
This PR follows #617 by adding a couple more community level permissions `DeleteCommunity` & `InviteToCommunity`

So far the policy checks are pretty straightforward, but the test hacking is starting to get cumbersome 😅 

I think my next PR will attempt to tackle a "default" collection of roles & policies (i.e. community creator gets `Steward` with all permissions, plus a default `Member` with a minimum set of permissions.)


src/lib/vocab/community/communityServices.ts
```diff
@@ -205,6 +202,7 @@ export const InviteToCommunityService: ServiceByName['InviteToCommunity'] = {
 			const {params} = serviceRequest;
 
 			const {actor, community_id, name} = params;
+			unwrap(await checkPolicy(actor, community_id, permissions.InviteToCommunity, repos));
```
greatbacon, line 205: So I didn't run against it in this PR, but it'll happen pretty soon, which is that this `InviteToCommunityService` makes another service call to `CreateAssignmentService` on line 218. Following the pattern as set so far, there'll eventually be another policyCheck at the top of that call, but it'll be looking for a different permission `CreateAssignment` which a user isn't guaranteed to have. 

My first thought is to pull out some of the common code into a helper & just directly invoke that, but I know you're not keen on that. I'll add an item to our agenda to discuss.

ryanatkn:
re: default roles/policies, see the `Template` work here, is related and includes a placeholder for `roles`: https://github.com/feltcoop/felt-server/pull/627/files#diff-24bd88e0603e57f7780c3e5b7ba7ad9c46d677a5e2cc05e3711167d4be8a9321

ryanatkn (approved):
nice besides the hacks it doesn't take much to implement

Was thinking it'd be a bit more readable to make the permission the first param to `checkPolicy`

PR #631, opened 2022-12-07
drive the UI with view data

ryanatkn:
closes https://github.com/feltcoop/felt.social/issues/2

- [x] change `Workspace` from being a special component to a view plugin
- [x] move the luggage and nav into `Workspace` (making them optional for the app as a whole, we'll probably make them plugins)
- [x] add `ui.mainLayoutView` rendered by `MainLayout` that defaults to `'<Workspace />'`
  - uses `SvastText` like rendering `space.view`
  - to see this in action, go to the console and run `ui.mainLayoutView.set('umm')`

questions:

- what could interacting with `ui.mainLayoutView` look like from a user's perspective? is a similar usecase to `space.view`  - we've talked about a `Layout` or `Profile` collection but it's not clear to me what the first simple+good design would be
- is `Workspace` the same kind of thing as views? Should it use `ui` directly or `getViewContext`?
  - currently uses `getViewContext`, but the problem is when it's mounted, we needed a view context that could contain null values, so we lose type safety for it
  - space views vs layout views?
- `Workspace` may be a more generalized component than what we have so far, which is calling our entire UI "Workspace" -- I'm wondering if instead we have a `<Dashboard />` to represent our current UI - it would do the current `Workspace` stuff, and the workspace could be part of the view manager UX and either replace or be used by the `MainLayout` (in this case, `Workspace` would not be a plugin, but its contents would become a new `Dashboard` plugin view)

followups:

- presets for different layouts besides `Workspace`? I think a centered column and top nav layout is something I'll work on soon.
- source the `view` data from wherever makes sense (account/community/space settings?)


src/lib/ui/MainLayout.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+	import {getApp} from '$lib/ui/app';
+	import SvastText from '$lib/ui/SvastText.svelte';
+	import ViewContext from '$lib/ui/ViewContext.svelte';
+
+	const {
+		ui: {mainLayoutView, personaSelection, communitySelection, spaceSelection},
+	} = getApp();
+
+	// TODO what can we do about this type? making the `viewContext` values nullable causes a ton of type errors
```
ryanatkn, line 10: the problem here is that our main layout view the `Workspace` may not have a space/community/persona, but our `ViewContext` normally provides all 3. This is really just a type safety issue, so I'm ok moving forward, knowing that it's possible to mount views that cause errors in some contexts.

greatbacon (approved):
Definitely don't have answers or thoughts on all questions, but I think at least in the realm of `Workspace` as view or something else I would advocate for presenting/naming/architecting it as something distinct from View.

Views in my understanding, are closely associated with the conception of Spaces, and just looking at the from of `Workspace` (or w/e name it may change to) it seems to float at a level high above Space & even Community. I would say it's almost at the instance level?

I think once we reach a point where we are discussing other forms to slot in & where they live, a better term will come to us.

ryanatkn:
Agreed that would resolve the comment above. We currently treat `Mention` and `Link` as views, which confuse things from the lower level POV like `Workspace` does from the higher level.

src/lib/app/views.ts
```diff
@@ -18,6 +19,7 @@ import List from '$lib/plugins/ryanatkn/list/List.svelte';
 import Lists from '$lib/plugins/ryanatkn/lists/Lists.svelte';
 
 export const views: Record<string, typeof SvelteComponent> = {
+	Workspace,
```
ryanatkn, line 22: - app views
- space views
- widget views

src/lib/ui/MainLayout.svelte
ryanatkn: We're now calling these different kinds/types of views -- app views, space views, and widget views. We'll figure out type safety in the future.

PR #634, opened 2022-12-11
add template for roles and policies

ryanatkn:
Adds roles and policies to the community templates, and adds some initial data for the `example` community. We can now use this to implement basic templates pretty easily in the community creation form.

followup:

- move the existing view templates in combination with these, keeping in mind how these are likely extension points


src/lib/app/templates.ts
```diff
@@ -1,12 +1,15 @@
...
+import type {PermissionName} from '$lib/vocab/policy/permissions';
 
 // TODO where does this belong? vocab?
 
 export interface CommunityTemplate {
 	name: string;
 	spaces?: SpaceTemplate[];
-	// TODO roles: RoleTemplate[{name, creator, default}];
+	roles?: RoleTemplate[];
+	policies?: PolicyTemplate[];
```
greatbacon, line 12: would it make more sense to have the policy templates array explicitly an attribute of each role template, instead of just a big array that sits next to the roles?

ryanatkn: nice yeah did that here: https://github.com/feltcoop/felt-server/pull/634/commits/2268395f0b684a84f0724f1a2cd2899ce5e0d32b

this is much better because `PolicyTemplate` shouldn't actually care what role it's for -- the template itself is just the permission and optional data. The context in which it's used provides the role -- including the role was a misunderstanding of the `Template` responsibility.

greatbacon (approved):

PR #635, opened 2022-12-11
rename some form components and splat `$$restProps` consistently in forms

ryanatkn:
Instead of `SomeInput` we've been using `SomeForm` whenever we have a top-level `form` element exported from a component, and this renames a couple stragglers. Also makes the splatting of `$$restProps` consistent in all `Form` components`


src/lib/ui/CommunityEditor.svelte
```diff
@@ -19,7 +19,7 @@
 </script>
 
 <div class="community-editor column">
-	<form {...$$restProps} class="markup">
+	<form class="markup" {...$$restProps}>
```
ryanatkn, line 22: I made the order of these consistent across all forms. By putting the props splat after class, any `class` prop will override the default one, which is preferred to let the caller customize.

greatbacon (approved):

PR #641, opened 2022-12-13
add stream view

ryanatkn:
**edit: this is an interesting idea but the problem is it locks the item components like `ChatItem` into a view interface. I think this is desirable long term but I don't want to deal with these decisions yet until some things mature. Closing this PR and will likely revisit after alpha is out.**

Adds the `Stream` view, which is a view similar to `List` and `Chat` but more abstract and generic.

It takes an `itemView` string property for looking up an item component used in the `{#each`. (a problem related to but not introduced by this PR: we need to lazy load components instead of loading the entire library of views ahead of time)

It'll be good to develop this alongside `List` and `Chat` not as a full replacement for them, but to see what makes them different, and to upstream as much generic behavior as makes sense to `Stream`. For example, it currently has a hardcoded input component, `StreamInput`, but we could imagine that being a prop as well.

Adds `ChatItem`, `ListItem`, and `ListsItem` to the published views -- but now that I'm looking at this, `ChatItem` doesn't get `persona`, and others like `ForumItem` should be published too under the same pattern.


src/lib/app/views.ts
```diff
@@ -33,5 +38,8 @@ export const views: Record<string, typeof SvelteComponent> = {
 	EntityExplorer,
 	Todo,
 	List,
+	ListItem,
 	Lists,
+	ListsItem,
+	Stream,
```
ryanatkn, line 44: Registering item views here allows components to look them up by name. The motivation was `Stream` taking an `itemView` prop which is indexed into views here, `views[itemView]`, e.g. `views['ListItem']`

PR #639, opened 2022-12-12
Add space id entity

greatbacon:
This PR returns the space_id to entities, plus a foreign key constraint (as well as one for persona_id)

PR #642, opened 2022-12-14
fix login bug

ryanatkn:
- [x] query spaces and directories separately by account
- [x] remove `filterByAccountWithDirectories`
- [x] implement `filterDirectoriesByAccount` -- waiting on #639 to pass CI


greatbacon:
Poking around at this, the issue seems to stem from `SpaceRepo.filterByAccountWithDirectories`.  The query there does an unusual constructed thing where it actually returns 2 objects per row, and it uses `json_build_object` to do so.

It looks like in the process the timestamps that normally get coerced into Dates are instead getting coerced into strings.

If you add the following code to the bottom of the function, it "fixes" things.
```
data.forEach((r) => {
			r.space.created = new Date(r.space.created);
			r.entity.created = new Date(r.entity.created);
			if (r.space.updated) {
				r.space.updated = new Date(r.space.updated);
			}
			if (r.entity.updated) {
				r.entity.updated = new Date(r.entity.updated);
			}
		});
```

Not saying that's the long term solution, but it makes things functional which indicates the issue root cause.

greatbacon (approved):

PR #643, opened 2022-12-15
use community templates in `CreateCommunity`

ryanatkn:
Changes the API to use templates for community creation.

- changes `CreateCommunity` params to accept a template (params changed to nest `name` and `settings` inside a `template` field, and now we can optionally include `spaces` and `roles` there as well)
- call `initTemplateGovernanceForCommunity` not just for standard communities, but for personal and admin communities
- some cleanup and misc improvements around these

followup:

- see comment about template roles and invariants: https://github.com/feltcoop/felt-server/pull/643#discussion_r1058430856
- convert all of the templates to schemas so we can validate - `// TODO add template schemas and use $ref` - we lost schema checking on `settings` in `CreateCommunity` params in this PR
  - also either check for a single `roleTemplate.default` or change the API to a reference
- refactor template data with default spaces, probably splitting default data and template types
- split data from types in `$lib/app/templates` (either to each vocab item or `/vocab/template/template`) - https://github.com/feltcoop/felt-server/pull/643#discussion_r1058445883


src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -30,3 +32,21 @@ export const checkPolicy = async (
 	}
 	return OK;
 };
+
+export const createPolicies = async (
```
ryanatkn, line 36: following the pattern of `createSpaces` with this and `createRoles` -- we'll want to use the same parallelization strategy for all of them when we figure that out (maybe a batch repo method?)

src/lib/vocab/community/communityEvents.ts
```diff
@@ -8,18 +8,18 @@ export const CreateCommunity: ServiceEventInfo = {
 		type: 'object',
 		properties: {
 			actor: {type: 'number'},
-			name: {type: 'string'},
-			settings: {$ref: '/schemas/InitialCommunitySettingsSchema.json'},
+			template: {type: 'object', tsType: 'CommunityTemplate'}, // TODO add template schemas and use $ref
```
ryanatkn, line 11: I think this extra object wrapper for `template` is worth it. The pattern may propagate across the app though..

src/lib/vocab/community/communityHelpers.server.ts
```diff
@@ -99,18 +100,14 @@ export const initAdminCommunity = async (
 export const initDefaultRoleForCommunity = async (
```
ryanatkn, line 100: separating init from create here, so we can create in one place and then conditionally init

src/lib/vocab/community/communityHelpers.server.ts
```diff
@@ -99,18 +100,14 @@ export const initAdminCommunity = async (
...
-
-	const settings = {...community.settings, defaultRoleId: defaultRole.role_id};
-
+	role: Role,
+): Promise<void> => {
+	const settings = {...community.settings, defaultRoleId: role.role_id};
 	unwrap(await repos.community.updateSettings(community.community_id, settings));
 	community.settings = settings;
-
-	return {ok: true, value: defaultRole};
```
ryanatkn, line 111: separating init from create here, so we can create in one place and then conditionally init 

the function doesn't need a return value because we're mutating the `Community`

src/lib/db/seed.ts
```diff
@@ -123,65 +118,23 @@ export const seed = async (db: Database, much = false): Promise<void> => {
...
-		// 							policyTemplate,
-		// 							mainPersonaCreator.persona_id,
-		// 							roleIdByName[roleTemplate.name],
-		// 						),
-		// 					}),
-		// 				);
-		// 			}
-		// 		}
-		// 	}
-		// }
```
ryanatkn, line 184: the `communityTemplate` above includes roles+policies

src/lib/db/seed.ts
```diff
@@ -272,9 +219,25 @@ interface SeedContext {
...
+		// TODO
+	},
+	Home: async () => {
+		// TODO
+	},
+	PersonalHome: async () => {
+		// TODO
+	},
+	InstanceAdmin: async () => {
+		// TODO
```
ryanatkn, line 232: these todos silence the warning during `gro lib/db/create` that there's no entities -- my thinking isn't crystal clear here on what to do or if we should even keep that warning

src/lib/vocab/community/communityHelpers.server.ts
```diff
@@ -63,118 +59,79 @@ export const initAdminCommunity = async (
...
-		);
-	} else {
-		creatorAssignment = unwrap(
-			await repos.assignment.create(actor, community.community_id, roles[0].role_id),
-		);
-	}
-	return {ok: true, value: {roles, creatorAssignment, policies}};
+
+	// TODO in some cases we probably want to add both the `creatorRoleId` and `defaultRoleId` assignment, how to express that in the template data?
+	const creatorAssignmentRoleId = creatorRoleId || defaultRoleId || roles[0].role_id;
```
ryanatkn, line 131: several Qs here:

- sometimes we may want to give a default role to the creator, but other times not (e.g. "member" and "newcomer") - I'm not sure how we could express this -- that's why I changed the API to an array, which I think is the right call, but I balked at changing existing behavior and left the todo
- isn't having a `defaultRoleId` an invariant we expect elsewhere, so we should expect it here too? As written, this helper now gets called for all community types, so it can be safely used for validation
- on the subject of validation, it's a little strange that this silently ignores cases like multiple `creator`/`default` templates, creating only exactly one assignment

src/lib/app/templates.ts
```diff
@@ -1,12 +1,14 @@
 import type {EntityData} from '$lib/vocab/entity/entityData';
 import type {CreatePolicyParams, CreateRoleParams, CreateSpaceParams} from '$lib/app/eventTypes';
 import type {Policy} from '$lib/vocab/policy/policy';
-import type {PermissionName} from '$lib/vocab/policy/permissions';
+import {permissionNames, type PermissionName} from '$lib/vocab/policy/permissions';
+import type {InitialCommunitySettings} from '$lib/vocab/community/community';
 
 // TODO where does this belong? vocab?
```
ryanatkn, line 7: I'm thinking we want to extract types to `src/lib/vocab/template/template.ts` to mirror `$lib/vocab/view/view.ts`, and then move `$lib/vocab/space/defaultSpaces.ts` to `$lib/app` where we store existing app data

greatbacon: That makes sense to me.

src/lib/app/templates.ts
```diff
@@ -62,7 +64,19 @@ export const policyTemplateToCreatePolicyParams = (
 	permission: template.permission,
 });
 
-export const defaultRoles: RoleTemplate[] = [
+// TODO integrate with default space data, `$lib/vocab/space/defaultSpaces.ts`
```
greatbacon, line 67: 🤯

src/lib/app/templates.ts
```diff
@@ -78,3 +92,12 @@ export const defaultRoles: RoleTemplate[] = [
 		policies: [{permission: 'InviteToCommunity'}],
 	},
 ];
+
+export const defaultPersonalCommunityRoles: RoleTemplate[] = [
+	{
+		name: 'Owner',
+		creator: true,
+		default: true,
+		policies: allPolicies,
```
greatbacon, line 101: 🤌

src/lib/vocab/community/communityHelpers.server.ts
greatbacon: Maybe instead of designating creator/default at the Role level, like you say put it at a higher level of the template interface would probably be the most useful, especially if that gives us some type of guarantee of a `defaultRoleId`. Maybe couple that with `creatorRole` actually being `creatorRoles` with an array of roles to assign and if it's empty just use the defaultRole?

greatbacon (approved):

src/lib/vocab/community/communityHelpers.server.ts
ryanatkn: yeah that sounds good, and this PR is so big that I'll get it merged and focus on the followups next

PR #647, opened 2022-12-16
clean orphan communities when deleting personas

ryanatkn:
Oversight in `DeletePersona`, need to check all of their communities to see if they're orphaned.


src/lib/vocab/community/communityHelpers.server.ts
```diff
@@ -17,16 +17,18 @@ const log = new Logger(gray('[') + blue('communityHelpers.server') + gray(']'));
 const DEFAULT_ROLE = 'member';
 
 export const cleanOrphanCommunities = async (
-	community_id: number,
+	communityIds: number | number[],
```
ryanatkn, line 20: We normally don't do APIs like this, but seemed convenient in this case. I don't mind changing it to only accept an array, changes 2 other callsites.

ryanatkn (commented):
accidental review button click, ignore me, can't delete :(

src/lib/vocab/community/communityHelpers.server.ts
```diff
@@ -17,16 +17,18 @@ const log = new Logger(gray('[') + blue('communityHelpers.server') + gray(']'));
...
 	repos: Repos,
 ): Promise<Result> => {
-	log.trace('checking if community is orphaned', community_id);
-	const accountPersonaAssignmentsCount = unwrap(
-		await repos.assignment.countAccountPersonaAssignmentsByCommunityId(community_id),
-	);
-	if (accountPersonaAssignmentsCount === 0) {
-		log.trace('no assignments found for community, cleaning up', community_id);
-		unwrap(await repos.community.deleteById(community_id));
+	if (!Array.isArray(communityIds)) communityIds = [communityIds]; // eslint-disable-line no-param-reassign
```
ryanatkn, line 23: We normally don't do APIs like this, but seemed convenient in this case. I don't mind changing it to only accept an array, changes 2 other callsites.

greatbacon: I think only accepting an array is the right call, since that's how we do it in our other APIs like this.

greatbacon (approved):
🐛 🔨

PR #648, opened 2022-12-20
make repo method names consistent

ryanatkn:
went through all the repos and named things consistently to our current patterns


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -128,17 +128,17 @@ export class EntityRepo extends PostgresRepo {
 	}
 
 	//This function finds all non-directory entities with no ties pointing to them & returns an array of their ids
-	async findOrphanedEntities(): Promise<Result<{value: number[]}>> {
-		const data = await this.sql<Entity[]>`
+	async filterOrphanedEntities(): Promise<Result<{value: Array<{entity_id: number}>}>> {
+		const data = await this.sql<Array<{entity_id: number}>>`
```
ryanatkn, line 132: I'm planning to do a pass at columns for all repo methods, so this is an in-between state.

greatbacon (approved):

PR #653, opened 2022-12-27
add hue buckets for visual consistency

ryanatkn:
Currently we randomize hue to an integer between 0 and 360 which makes the color scheme completely random - we can have pea green next to grass green next to forest green. This changes the hues to be rounded to one of some number of buckets, so the jump from green up and down the rainbow is more distinct and nicer looking.

A designer wouldn't split colors into even buckets because of the way color perception works, so I left a TODO about choosing a color theme.

We still let communities choose a hue of any value, but this at least improves the defaults substantially.


src/lib/ui/color.ts
```diff
@@ -1,5 +1,14 @@
 import {toRandomAlea} from '@feltcoop/util/random-alea.js';
 
+// TODO design a color scheme instead of calculating a number of evenly spaced buckets
+export const HUE_BUCKETS = 7;
```
ryanatkn, line 4: this could probably be increased

greatbacon (approved):

PR #656, opened 2023-01-02
Initial data model diagram

greatbacon:
This PR is for some first draft diagrams to add to our documentation


ryanatkn:
neato, some thoughts:

- add `permissions`
- entities don't point to ties unless I misunderstand what the arrows mean
- what useful info is in the text besides "Here are felt-server's SQL database tables"?
- the text says 8 but there's 9 pictured, I'd omit the count

ryanatkn (approved):

greatbacon:
I think follow up diagrams that do ER for the system vocab & a sequence diagram for say "lifecycle of an event" would be super useful!

ryanatkn:
Ties are the main way we're drawing relationships right now, but the ActivityStreams vocabulary is full of cross references, properties referencing other entities by id. The first usecase will be ordered collections - type `Collection` with an `items` property that's a list of entity ids.

PR #657, opened 2023-01-03
Put 5 new permissions in place, plus updated test

greatbacon:
This PR adds 5 more permissions for policies, all related to managing Roles & Assignments

There is still an issue with unwrapError in the assignementservices tests, can't quite  tell what the issue is though...


src/lib/server/servicesIntegration.test.ts
```diff
@@ -56,7 +56,7 @@ test_servicesIntegration('services integration test', async ({db, random}) => {
 		await CreateAssignmentService.perform({
 			...serviceRequest1, // add `persona2` with `persona1`
 			params: {
-				actor: persona2.persona_id,
+				actor: persona1.persona_id,
```
ryanatkn, line 59: # :tada:

src/lib/vocab/community/communityServices.ts
```diff
@@ -215,18 +214,14 @@ export const InviteToCommunityService: ServiceByName['InviteToCommunity'] = {
...
-						role_id: community.settings.defaultRoleId,
-					},
-				}),
+			const assignment = unwrap(
+				await createAssignment(
+					persona.persona_id,
+					community_id,
+					community.settings.defaultRoleId,
+					repos,
+				),
```
ryanatkn, line 223: much nicer

src/lib/vocab/assignment/assignmentHelpers.server.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const createAssignment = async (
+	persona_id: number,
+	community_id: number,
+	role_id: number,
+	repos: Repos,
+): Promise<ApiResult<Assignment>> => {
+	log.trace('creating assingment for', persona_id, community_id, role_id);
+	// Personal communities disallow assignments as a hard rule.
+	const community = unwrap(await repos.community.findById(community_id));
```
ryanatkn, line 19: We're loading `community` twice in the `InviteToCommunityService` service, I think the function args should change `community_id` to `community`.

ryanatkn:
Doesn't have to be in this PR, I think it's wise for us to add a test every time we add a persona that fails with the wrong actor, so we lock in security as we go.

src/lib/util/randomEventParams.ts
```diff
@@ -128,7 +128,7 @@ export const randomEventParams: RandomEventParams = {
 	},
 	CreateAssignment: async (random, {account, persona, community} = {}) => {
 		if (!persona) ({persona} = await random.persona(account));
-		if (!community) ({community} = await random.community()); // don't forward `persona`/`account` bc that's the service's job
+		if (!community) ({community} = await random.community(persona)); // don't forward `persona`/`account` bc that's the service's job
```
ryanatkn, line 131: this is related to the issue, see the comment, need to think about this a bit

greatbacon: Alright, so I cleared this up & fixed a couple of other tests. At this point it looks like the issues seems to be around the `assignmentHelpers` and error propagation. 

If I change the call in `assignmentServices` to the following:

ryanatkn:
scratch that about tests, looks like we can do them generically: #659

src/lib/vocab/assignment/assignmentHelpers.server.ts
```diff
@@ -0,0 +1,41 @@
+import {unwrap} from '@feltcoop/util';
+import {Logger} from '@feltcoop/util/log.js';
+
+import {blue, gray} from '$lib/server/colors';
+import type {Repos} from '$lib/db/Repos';
+import type {Assignment} from '$lib/vocab/assignment/assignment';
+import type {ApiResult} from '$lib/server/api';
+import type {Community} from '../community/community';
```
ryanatkn, line 8:
```suggestion
import type {Community} from '$lib/vocab/community/community';
```

src/lib/vocab/assignment/assignmentServices.ts
```diff
@@ -15,33 +18,16 @@ export const CreateAssignmentService: ServiceByName['CreateAssignment'] = {
...
-			const {community_id, persona_id, role_id} = params;
-			log.trace('[CreateAssignment] creating assignment', persona_id, community_id);
-
-			// Personal communities disallow assignments as a hard rule.
+			const {actor, community_id, persona_id, role_id} = params;
+			log.trace('[CreateAssignment] creating assignment for persona & role', persona_id, role_id);
+			log.trace('[CreateAssignment] checking policy', actor, community_id);
 			const community = unwrap(await repos.community.findById(community_id));
 			if (!community) {
 				return {ok: false, status: 400, message: 'community not found'};
```
ryanatkn, line 26:
```suggestion
				return {ok: false, status: 404, message: 'community not found'};
```

checked elsewhere, we do 404s for this in a 5 other places

src/lib/util/randomEventParams.ts
ryanatkn: looks like I got the fix here: https://github.com/feltcoop/felt-server/pull/660

src/lib/vocab/assignment/assignmentServices.ts
greatbacon: As a wise person once said, "if you put it off long enough, someone smarter might do it first" :D

I agree though about probably wanting to do it sooner rather than later. I know we have a bunch of spots in the UX where error messages don't even propagate to users, so it might make sense to do some of that planning/refactoring at the same time.

ryanatkn (approved):

PR #659, opened 2023-01-04
add generic tests for unauthorized actors calling authorized services

ryanatkn:
Looking at the permissions checks we're adding in PRs like #657, it seems we can generically test for failed authorization with unauthorized actors. As we add additional checks, we'll remove those events from the list of strings in this PR and tests should fail if we make a mistake with authorization, so it's basically our authorization todo list, and it'll pick up newly created authorized events as test failures unless we add authorization or add them to this temporary list.

It's nice that we get to keep leveraging our data-driven events and services like this, this replaces a *lot* of boilerplate tests


src/lib/server/services.test.ts
```diff
@@ -112,6 +112,68 @@ for (const service of services.values()) {
 			session.signOut(); // sign out after `SignUp` so `SignIn` works (otherwise "already signed in")
 		}
 		assert.is(result.status, 200); // TODO generate invalid data and test those params+responses too
+
+		// Test failure of authorized services with an unauthorized actor.
+		if (
+			event.authorize !== false &&
+			![
+				// TODO when this list is empty, we're fully authorized!
```
greatbacon, line 120: 😮

greatbacon (approved):

PR #663, opened 2023-01-11
rewrite Board view

ryanatkn:
updated `Board`:

- no clicking into a post's details, all posts are shown by default in one list (will add pagination later)
- reverse chronological, both for the top-level posts and replies (we'll make it configurable in the future)
- rewrote the layout to be one inline thing instead of a separate header (so `Board` is compact and cozy by default, not expansive and airy like linear forums)

future

- add a collapser
- fix word-break
- pagination


src/lib/vocab/view/view.ts
```diff
@@ -29,8 +29,8 @@ export const viewTemplates: ViewTemplate[] = [
 	{name: 'AdminHome', view: '<AdminHome />', icon: '🏠', creatable: false, admin: true},
 	{name: 'Chat', view: '<Chat />', icon: '🗨'},
 	{name: 'ReplyChat', view: '<ReplyChat />', icon: '↳'},
-	{name: 'Board', view: '<Board />', icon: '📚'},
-	{name: 'Forum', view: '<Forum />', icon: '📋'},
+	{name: 'Board', view: '<Board />', icon: '📋'},
+	{name: 'Forum', view: '<Forum />', icon: '📚'},
```
ryanatkn, line 33: these icons seemed swapped now that we scoped the `Board` down

greatbacon (approved):
Before this gets merged, would it be possible to port the existing `Board` functionality to the `Forum` view?

ryanatkn:
good catch, I replaced the old `Board` with the current `Forum`, which was under-developed.

PR #664, opened 2023-01-12
support network-relative links prefixed with a double slash

ryanatkn:
linkifies //a.b to https://a.b -- seems like an obvious improvement given we support single slash links

I could see an argument that it should support community names (either in addition to, or instead of the external https: links), but we already support  `@community-name` so I'm not sure if we need the `/`-prefixed version of that path (I can see it both ways right now, and it'd be an easy change)

it's an interesting question that overlaps with some other questions we've been asking, is `@community-name` the same as `$HOST/community-name` or are they 2 separate types (community persona vs community space) from the user's POV

also improves link checks to avoid false positives


greatbacon:
Sorry, for my clarity, this change makes it so that a post containing `//google.com` maps to `https://google.com`?

And the other questions you're contemplating are:
Should `//feltcoop` be able to point to `https://felt.dev/feltcoop` (something technically covered by `@feltcoop`) but then should the `@feltcoop` instead refer to the CommunityPersona and we use the `//` syntax to refer to the commuinty/URL?

ryanatkn:
yep exactly

greatbacon (approved):

PR #666, opened 2023-01-17
fix `initEnv` for projects with no env files

ryanatkn:
started by the discussion here: https://github.com/feltcoop/felt-server/commit/d17abd4647b42b04e7aff17b92db929a3a29c7d9

This isn't totally done but I'm ready to go over it, and talk through the remaining issues. (I think `updateEnv` is still broken like in the above commit)

Fixes `gro dev` and `gro deploy` for projects that haven't yet initialized their `.env` files. To see the problem, delete or rename the `.env` files and run one of those tasks. Incidentally `gro lib/db/create` works because it hardcodes the defaults as fallbacks.

I don't actually know if this is the right fix yet, having difficulty seeing where `gro build` outputs a bad env in the bad commit (https://github.com/feltcoop/felt-server/commit/d17abd4647b42b04e7aff17b92db929a3a29c7d9)

- probably move the vars to `.env.development`/`.env.production` not the base `.env`?
- buggily disallowing 0 values for the PG timeout env vars

be sure to update your local .env files!

followup:

- import directly from SvelteKit instead of `fromEnv` - needs gro support because of tasks


src/lib/infra/updateEnv.task.ts
```diff
@@ -23,5 +25,7 @@ export const task: Task = {
 			log.info(`writing updated git hash ${gitHash} to ${envFile}`);
 			await fs.writeFile(envFile, updatedEnvContents, 'utf8');
 		}
+
+		reloadEnv();
```
ryanatkn, line 29: conceptually this makes sense I think? not sure about the implementation

ryanatkn:
ah env woes, truly a fitting subject for PR #666

src/lib/db/postgres.ts
```diff
@@ -16,21 +17,13 @@ export interface PostgresOptions extends Options<any> {
...
+	host: fromEnv('PGHOST'),
+	port: Number(fromEnv('PGPORT')),
+	database: fromEnv('PGDATABASE'),
+	username: fromEnv('PGUSER'),
+	password: fromEnv('PGPASSWORD'),
+	idle_timeout: Number(fromEnv('PGIDLE_TIMEOUT')) || undefined,
+	connect_timeout: Number(fromEnv('PGCONNECT_TIMEOUT')) || undefined,
 });
 
 export const defaultPostgresOptions = toDefaultPostgresOptions();
```
ryanatkn, line 29: this might be a problem if we're relying on specific ordering in `updateEnv` for example

src/lib/infra/.env.default
```diff
@@ -1,2 +1,9 @@
...
-PUBLIC_ADMIN_ICON=💚
\ No newline at end of file
+PUBLIC_ADMIN_ICON=💚
+PGHOST=localhost
+PGPORT=5432
+PGDATABASE=felt
+PGUSER=postgres
+PGPASSWORD=password
+PGIDLE_TIMEOUT=
+PGCONNECT_TIMEOUT=
```
ryanatkn, line 9: Be sure to add to your local env

greatbacon (approved):

PR #668, opened 2023-01-18
Fix migration issue for posterity

greatbacon:
This bug blew away some data. The issue *shouldn't* impact any other users going forward, but just to be safe...


ryanatkn (approved):
doh good catch! This particular mistake would have been caught if we set the result type of `await sql`, but lots of other things could still slip through

PR #674, opened 2023-01-20
factor out a result helper

ryanatkn:
Factors out a helper we copypasted. More than code de-duplication I wanted this to be documented with types, the terminology can be confusing.

Also resolves a type todo.


src/lib/util/websocket.ts
```diff
@@ -14,8 +12,3 @@ export interface StatusMessage {
 	status: number;
 	message: string;
 }
-
-export type WebsocketResult<T = any> = Result<
-	{status: number; value: T},
-	{status: number} & ErrorResponse
->;
```
ryanatkn, line 21: Was able to delete this and use `ApiResult` everywhere instead, it was an unnecessary duplication.

greatbacon (approved):

PR #677, opened 2023-01-21
add control and ownership diagram

ryanatkn:
This documents some terms we're starting to use more formally. I'm not sure exactly where it belongs. (admin.md didn't seem quite right)


ryanatkn:
ohh that's dope

PR #679, opened 2023-01-22
install deployment deps using `npm ci`

ryanatkn:
Reading the SvelteKit docs on the Node adapter, I saw the recommendation to use `npm ci` over `npm i` on deployment: https://kit.svelte.dev/docs/adapter-node

See also the `npm ci` docs which say it should be used for deployment: https://docs.npmjs.com/cli/v6/commands/npm-ci

I think I finally internalized the point of the command.


src/lib/infra/deploy.task.ts
```diff
@@ -56,8 +56,8 @@ export const task: Task = {
 			tar -xvf ${artifactFilename};
 			echo 'moving node_modules';
 			mv ~/${currentDeploy}/node_modules ./ && echo 'reusing node_modules' || echo 'no node_modules to reuse';
```
ryanatkn, line 58: hmm noticing that my probably too-clever strategy of reusing `node_modules` here (which was only done to speed up deploying) is undone by `npm ci`, which always deletes `node_modules`.

greatbacon: may as well just clean it out then, eh?

greatbacon (approved):

PR #680, opened 2023-01-24
render schema descriptions with rich text

ryanatkn:
In the docs text, changes \`Space\` to `<Vocab name="Space" />` to reference vocabulary items. The new `Vocab` widget view makes it an internal link and provides contextmenu behavior that brings up a dialog with the `SchemaInfo`. 

Also upgrades felt-ui with a tiny patch to fix `.chip` backgrounds, was referencing the old CSS var.

In the future we could make these easier to read and write for both devs and end users by adding a control character to reference system-level objects, so you could write something like `~Space` in any Fuz context, instead of `<Vocab name="Space" />`. An interesting usecase is referencing an action and getting shortcut contextual access to it. (and, without any work on our part, it'll support optional baked params and customizable text, via props and child content, which should be extraordinarily powerful as a way to let users build UI by creating these buttons with behavior)


src/lib/ui/SvastView.svelte
```diff
@@ -16,11 +16,12 @@
...
-			this={components[view.tagName]}
-			{...props}
-			>{#each view.children as child (child)}<svelte:self view={child} />{/each}</svelte:component
-		>{:else}<code class="error-text">[unknown view: {view.tagName}]</code
+{#if view.type === 'svelteComponent'}{#if view.tagName in components}{#if view.children.length}<svelte:component
+				this={components[view.tagName]}
+				{...props}
+				>{#each view.children as child (child)}<svelte:self view={child} />{/each}</svelte:component
+			>{:else}<svelte:component this={components[view.tagName]} {...props} />{/if}{:else}<code
+			class="error-text">[unknown view: {view.tagName}]</code
```
ryanatkn, line 24: This allows views to have default slots, and fixes the warnings we got for default slots. It's hard to read because of Svelte whitespace rules, we don't want this component to add any whitespace.

src/static/schemas/vocab.json
```diff
@@ -650,7 +650,7 @@
 		"Policy": {
 			"$id": "/schemas/Policy",
 			"type": "object",
-			"description": "\n\t\tPolicies are associated with Roles to describe the Actions a Role is able to take in the system.\n\t\tPermissions are the enumeration of the those Actions and are often 1:1.\n\t\t`data` is a currently-unused attribute earmarked for allowing for more complicated governance schemes.\n\t",
+			"description": "\n\t\tEach <Vocab name=\"Policy\" /> associates a <Vocab name=\"Role\" /> with a permission\n\t\tto describe the Actions that <Vocab name=\"Actor\" />s with the <Vocab name=\"Role\" /> are able to perform.\n\t\tPermissions are often 1:1 with Actions, but they don't have to be.\n\t\t`data` is a stub to support more complex governance schemes in the future.\n\t",
```
ryanatkn, line 653: I rewrote this one

greatbacon (approved):
siiiiick

PR #681, opened 2023-01-25
Add 2 more permissions

greatbacon:
Branch name was a little ambitious lol

Anyway, this adds 2 more checks. 

1 is a standard permission check for `Ephemera`

The other is for DeletePersona, which ran up against a corner case in tests which I'll detail below.


src/lib/server/services.test.ts
```diff
@@ -141,23 +141,27 @@ for (const service of services.values()) {
 
 			if (
 				[
+					'DeletePersona', //This event is outside the policy system and requires mismatched actor/persona_id to fail
```
greatbacon, line 144: So, the way the code in this test works, it sets up the `unauthorizedPersona`, but then goes on to use the same persona_id as `actor` and `persona_id` for event params. This however is explicitly the expected behavior for this event, so it fails to fail (lol).

Anyway, I wasn't quite sure about how to go about hacking this edge case, so I just left it in the list for now.

src/lib/vocab/persona/personaServices.test.ts
```diff
@@ -120,5 +121,38 @@ test__personaService('delete a persona and properly clean up', async ({db, rando
...
+			params: {actor: persona1.persona_id, persona_id: persona2.persona_id},
+		}),
+	);
+});
+
+test__personaService(
+	'actors can delete other personas if in the admin community',
+	async ({db, random}) => {
+		const account = await random.account();
+		const {persona} = await random.persona(account);
```
ryanatkn, line 141:
```suggestion
		const {persona} = await random.persona();
```

src/lib/vocab/persona/personaServices.ts
```diff
@@ -129,9 +129,10 @@ export const CreateAccountPersonaService: ServiceByName['CreateAccountPersona']
 export const DeletePersonaService: ServiceByName['DeletePersona'] = {
 	event: DeletePersona,
 	perform: async ({repos, params}) => {
-		const {persona_id} = params;
+		const {actor, persona_id} = params;
 
 		// first check if deleting the persona is allowed
+		//TODO extract to it's own policy helper?
```
ryanatkn, line 135: good todo, wondering if we want a way to refer to these, like "system personas"? so then they helper may be `isSystemPersona`?

ryanatkn (approved):
excellent, wrap up incoming!

src/lib/server/services.test.ts
ryanatkn: lol hm I'll think on it, no harm in leaving for now, a hardcoded hack may be fine

PR #684, opened 2023-01-27
Next batch permissions

greatbacon:
This PR adds the last few permissions related to community actions

edit: decided to do kick from community in a fresh PR


src/lib/server/services.test.ts
```diff
@@ -131,6 +132,15 @@ for (const service of services.values()) {
 				failedParams.actor = unauthorizedPersona.persona_id;
 			}
 
+			// disable community creation on instance for test
+			if (event.name === 'CreateCommunity') {
```
greatbacon, line 136: Sorry for the weird special snowflake blocks; but this was the only way I could think of making sure the appropriate instance setting was in place :(

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -47,3 +48,17 @@ export const checkCommunityAccess = async (
 		return {ok: false, status: 403, message: 'actor does not have permission'};
 	}
 };
+
+export const checkDisabledCommunityCreation = async (repos: Repos): Promise<boolean> => {
```
ryanatkn, line 52: With the current code we've been using `check`-prefixed functions to return `Result` (or `ApiResult`) and `is`-prefix functions return booleans.

If you want to make the naming more consistent with the setting property, how about:

- `isCommunityCreationDisabled`
- `settings.disableCommunityCreation`

I could see wanting to mirror the event name more closely too, but it doesn't read as fluidly:

- `isCreateCommunityDisabled`
- `settings.disableCreateCommunity`

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -47,3 +48,17 @@ export const checkCommunityAccess = async (
...
+
+	if (!result) {
+		throw Error('unable to locate admin community');
+	}
+
+	if (result.settings.instance) {
+		return !!result.settings.instance.disableNewCommunities;
+	} else {
+		return false;
+	}
```
ryanatkn, line 63: stylistic choice, could be shortened:

```suggestion
	return !!result.settings.instance?.disableNewCommunities;
```

src/lib/vocab/community/communityServices.ts
```diff
@@ -234,6 +243,11 @@ export const LeaveCommunityService: ServiceByName['LeaveCommunity'] = {
 			// 	repos.persona.findById(persona_id),
 			// 	repos.community.findById(community_id),
 			// ]);
+
+			if (actor !== persona_id || !(await checkCommunityAccess(actor, community_id, repos))) {
```
ryanatkn, line 247: Two bugs here:

- `checkCommunityAccess` needs to be unwrapped, this would always evaluate to false
- `checkCommunityAccess` doesn't seem right? should be `isPersonaAdmin` right?

```suggestion
			if (actor !== persona_id && !(await isPersonaAdmin(actor, repos))) {
```

src/lib/vocab/community/communityServices.ts
```diff
@@ -234,6 +243,11 @@ export const LeaveCommunityService: ServiceByName['LeaveCommunity'] = {
...
+				return {ok: false, status: 403, message: 'actor does not have permission'};
+			}
```
ryanatkn, line 249: Two bugs here:

- `checkCommunityAccess` needs to be unwrapped, this would always evaluate to false (I checked, we don't make this mistake elsewhere)
- `checkCommunityAccess` should always be checked, not conditionally, but on the `persona_id` not the `actor`
- `isPersonaAdmin` should be checked

```suggestion
			if (actor !== persona_id && !(await isPersonaAdmin(actor, repos))) {
				return {ok: false, status: 403, message: 'actor does not have permission'};
			}
			unwrap(await checkCommunityAccess(persona_id, community_id, repos));
```

src/lib/server/services.test.ts
ryanatkn: Since this depends on special behavior with the instance settings, I think it'd be cleaner to test it separately in `communityServices.test.ts`, not generically. That way we have a test dedicated for that instance property, and this stays simpler.

This test is meant to test failed authorization but not for admin-overriding cases. I think `'CreateCommunity'` is just one of the weird ones, because its purpose is to create the authorized context of a community. So I think it makes sense to change the line above `if (event.authorize !== false) {` to

```ts
if (event.authorize !== false && event.name !== 'CreateCommunity') {
```

and annotate the line with a comment explaining why `'CreateCommunity'` is an exception?

ryanatkn:
Had some things we may need to go back and forth on, `'CreateCommunity'` is indeed a weird one.

src/lib/server/services.test.ts
greatbacon: Okay cool, that sounds good to me. I'm never sure where to make the cut for special exceptions in these general testers lol

src/lib/vocab/community/communityServices.ts
greatbacon: Good catch, thanks!

src/lib/vocab/policy/policyHelpers.server.ts
greatbacon: That makes sense to me, I think the 1:1 mapping of the second form to the action it describes seems like a good pattern to establish.

src/lib/vocab/policy/policyHelpers.server.ts
greatbacon: I pretty much always choose the more verbose form because I'm not used to thinking in truthy/falsey terms so it makes it easier for me to reason about it when I write it. Not opposed to shorter form though lol

src/lib/vocab/community/communityServices.test.ts
```diff
@@ -204,5 +204,44 @@ test_communityServices('deleted communities cleanup after themselves', async ({d
...
+test_communityServices(
+	'when new communities are disabled, only admins should be able to create new ones',
+	async ({db, random}) => {
+		const {persona} = await random.persona();
+
+		const serviceRequest = toServiceRequestMock(db, persona);
+
+		const {settings} = unwrap(await db.repos.community.findById(ADMIN_COMMUNITY_ID))!;
+		const settingValue = settings.instance?.disableCreateCommunity;
+		await db.repos.community.updateSettings(ADMIN_COMMUNITY_ID, {
```
ryanatkn, line 216: should wrap with `unwrap` (not happy with how error prone this is...)

src/lib/vocab/community/communityServices.test.ts
```diff
@@ -204,5 +204,44 @@ test_communityServices('deleted communities cleanup after themselves', async ({d
...
+
+		unwrap(
+			await CreateCommunityService.perform({
+				...serviceRequest,
+				params: randomCommunityParams(actor.persona_id),
+			}),
+		);
+
+		//cleanup from test; do not delete
+		await db.repos.community.updateSettings(ADMIN_COMMUNITY_ID, {
```
ryanatkn, line 239: same as above, `unwrap`

ryanatkn (approved):
awesome that was a bit tricky, I edited my comments like 10 times, had 2 tweaks in the test

PR #688, opened 2023-01-30
Add update/delete/erase entity checks

greatbacon:
This PR adds the same check (entity ownership) and applies to three entity events (update, erase & delete)

The logic is supposed to be as follows:
Only the owner of an entity can take the action (i.e. actor must match persona_id on entity)

UNLESS!

The actor is an admin

OR

The view type of the space is designated as "common" allow anyone to edit any entity within the space.

I think I did it right... 😅


src/lib/vocab/view/view.ts
```diff
@@ -34,15 +35,19 @@ export const viewTemplates: ViewTemplate[] = [
 	{name: 'Notes', view: '<Notes />', icon: '🏷'},
 	{name: 'Iframe', view: '<Iframe />', icon: '💻'}, // TODO does this need a default `src`?
 	{name: 'EntityExplorer', view: '<EntityExplorer />', icon: '✏️', creatable: false},
-	{name: 'Todo', view: '<Todo />', icon: '🗒'},
-	{name: 'List', view: '<List />', icon: '🪜'},
-	{name: 'Lists', view: '<Lists />', icon: '🔨'},
+	{name: 'Todo', view: '<Todo />', icon: '🗒', common: true},
+	{name: 'List', view: '<List />', icon: '🪜', common: true},
+	{name: 'Lists', view: '<Lists />', icon: '🔨', common: true},
```
ryanatkn, line 40: I think we should avoid adding API here because it's probably not the direction we're going: I think this is a space policy concern, not a view concern. (e.g. changing a space's view shouldn't change its governance, not with the real implementation)

Instead I think we should hardcode the checks in `checkEntityOwnership`: `const common = space.view.includes('<Todo') || space.view.includes('<List')` -- the ugliness is a good thing, because we know it's a hacky implementation that can be bypassed by changing `space.view`, until we get around to implementing space policies.

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -58,3 +60,31 @@ export const isCreateCommunityDisabled = async (repos: Repos): Promise<boolean>
...
+	if (await isPersonaAdmin(actor_id, repos)) {
+		return {ok: true, status: 200, value: undefined};
+	}
+
+	const {entities, missing} = unwrap(await repos.entity.filterByIds(entityIds));
+	if (missing && missing.length > 0) {
+		return {ok: false, status: 400, message: 'unable to process non-existing entities'};
+	}
+
+	for (const entity of entities) {
```
ryanatkn, line 78: Let's loop by `space_id` here, e.g. `const spaceIds = new Set(entities.map((e) => e.space_id));`

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -58,3 +60,31 @@ export const isCreateCommunityDisabled = async (repos: Repos): Promise<boolean>
 
 	return !!result.settings.instance?.disableCreateCommunity;
 };
+
+export const checkEntityOwnership = async (
```
ryanatkn, line 64: not sure about this, maybe rename `checkEntityPolicy`? seems more future-aligned

src/lib/vocab/entity/entityServices.ts
```diff
@@ -12,6 +12,7 @@ import {
 import {toTieEntityIds} from '$lib/vocab/tie/tieHelpers';
 import type {Tie} from '$lib/vocab/tie/tie';
 import {cleanOrphanedEntities} from './entityHelpers.server';
+import {checkEntityOwnership} from '../policy/policyHelpers.server';
```
ryanatkn, line 15:
```suggestion
import {cleanOrphanedEntities} from '$lib/vocab/entity//entityHelpers.server';
import {checkEntityOwnership} from '$lib/vocab/policy/policyHelpers.server';
```

src/lib/server/services.test.ts
```diff
@@ -148,10 +148,6 @@ for (const service of services.values()) {
 					'ReadEntitiesPaginated', //TODO add checkCommunityAccess
 					'CreateEntity', //TODO add policy
 
-					'UpdateEntity', //TODO add checkEntityOwnership (actor===persona_id || view type === <> || admin)
-					'EraseEntities', //note about this check, should also add a Space setting to disable check
-					'DeleteEntities', //i.e. set a Space as "common", but make it a one way toggle
-
```
ryanatkn, line 154: # :broom: :dash:

src/lib/vocab/policy/policyHelpers.server.ts
greatbacon: good call

src/lib/vocab/policy/policyHelpers.server.ts
greatbacon: yeah, I think in this weird inbetween state, not using policy is the best course of action.

src/lib/util/randomVocab.ts
```diff
@@ -71,10 +71,14 @@ export const randomCommunityParams = (actor: number): CreateCommunityParams => {
...
-export const randomSpaceParams = (actor: number, community_id: number): CreateSpaceParams => ({
+export const randomSpaceParams = (
+	actor: number,
+	community_id: number,
+	view?: string,
+): CreateSpaceParams => ({
 	actor,
 	community_id,
-	view: randomView(),
+	view: view || randomView(),
```
ryanatkn, line 81: allows the empty string just in case

```suggestion
	view: view ?? randomView(),
```

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -75,12 +74,18 @@ export const checkEntityOwnership = async (
 		return {ok: false, status: 400, message: 'unable to process non-existing entities'};
 	}
 
-	for (const entity of entities) {
-		if (entity.persona_id !== actor_id) {
-			//TODO probably want some caching here if we're gonna have particularly large batches of updates/dels
+	const spaceIds = new Set(
+		entities.map((e) => (e.persona_id === actor_id ? null : e.space_id)).filter(Boolean),
```
ryanatkn, line 78: problem here is TypeScript doesn't understand the `filter(Boolean)`, so adding the `!` is a hack, but it means you can delete the `if (space_id) {` guard below

```suggestion
		entities.map((e) => (e.persona_id === actor_id ? null! : e.space_id)).filter(Boolean),
```

src/lib/vocab/policy/policyHelpers.server.ts
```diff
@@ -75,12 +74,18 @@ export const checkEntityOwnership = async (
...
+	);
+
+	for (const space_id of spaceIds) {
+		if (space_id) {
```
ryanatkn, line 82: can't offer the complete suggestion here, will need to delete the closing `}` below, see the above comment: https://github.com/feltjs/felt-server/pull/688/files/6416faec941f8e35da3864153378ca7de89bec87..e9e0c66fdd9b006c7fb05a1a76261bba627419aa#r1093766855

```suggestion
```

ryanatkn (approved):
looking at the test

src/lib/vocab/entity/entityServices.test.ts
```diff
@@ -232,5 +234,96 @@ test_entityServices('deleting entities and cleaning orphans', async ({random, db
...
+				params: {
+					actor: persona2.persona_id,
+					entity_id: note2.entity_id,
+					data: {type: 'Note', content: 'lol'},
+				},
+			}),
+		);
+
+		//case 4, actor !== persona in regular space | block
+		unwrapError(
```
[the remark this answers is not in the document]
ryanatkn: first thought is do you need a different mock `serviceRequest`? looking at the details

src/lib/vocab/view/view.ts
```diff
@@ -43,6 +43,10 @@ export const viewTemplatesByName: Map<string, ViewTemplate> = new Map(
 	viewTemplates.map((t) => [t.name, t]),
 );
 
+export const viewTemplatesByView: Map<string, ViewTemplate> = new Map(
+	viewTemplates.map((t) => [t.view, t]),
+);
+
```
ryanatkn, line 49:
```suggestion
```

PR #687, opened 2023-01-30
Add create & read entity checks

greatbacon:
This PR adds `checkCommunityAccess` checks to our two read entity services, & adds a policy permission check for the create event.

PR #689, opened 2023-01-31
fix tests for PR #687

ryanatkn:


src/lib/vocab/persona/personaServices.test.ts
```diff
@@ -89,13 +90,27 @@ test__personaService('delete a persona and properly clean up', async ({db, rando
...
+		await CreateAssignmentService.perform({
+			...toServiceRequestMock(db, otherPersona),
+			params: {
+				actor: otherPersona.persona_id,
+				community_id: otherCommunity.community_id,
+				persona_id: persona.persona_id,
+				role_id: otherCommunity.settings.defaultRoleId,
+			},
+		}),
+	);
```
ryanatkn, line 109: The test creates an entirely separate community, then ensures the entity created by the persona gets ghosted. It's more verbose than we'd like because there's no `random.assignment`, so I left a todo for that.

greatbacon: 🙇

greatbacon (approved):

PR #695, opened 2023-02-06
Remove last nested `perform` invocation

greatbacon:
This PR refactors the core CreateSpace logic into its own helper, allowing us to remove the lasted nested `perform` invocation.


src/lib/vocab/space/spaceHelpers.server.ts
```diff
@@ -1,28 +1,85 @@
...
 	}
 	return {ok: true, value: {spaces, directories}};
 };
+
+export const createSpace = async (
+	community_id: number,
+	name: string,
+	view: string,
+	path: string,
+	icon: string,
```
ryanatkn, line 40: I think we should use the params object directly here. Pulling it apart looks more verbose and error prone, like in the future we  may change `icon` to be optional or we may add a property, these call sites know too much about its internals.

```suggestion
	params: CreateSpaceParams,
```

ryanatkn (approved):
looks great, had one requested change about the call signature

PR #697, opened 2023-02-08
Make toServiceRequest calls inline in tests

greatbacon:
This inlines all but 2 `toServiceRequests` in our test suite.

Of the 2 I left in place, one is being dynamically set for every service (so it looked big and worth assigning)

The other was in a test that I think we actually want to deprecate and do the opposite for (allow chaining of perform calls)


src/lib/server/service.test.ts
```diff
@@ -40,6 +39,7 @@ test__service(`roll back the database after a failed transaction`, async ({db, r
 	assert.ok(!unwrap(await db.repos.community.findById(community.community_id)));
 });
 
+//TODO I think we actually want to deprecate this & make it fail eventually?
```
ryanatkn, line 42:
```suggestion
```

I rewrote this test here: https://github.com/feltjs/felt-server/pull/692/

ryanatkn (approved):
awesome, I think this fixes a lot of tests that brea kwith #692

PR #701, opened 2023-02-08
fix checkRemovePersona to fail if persona has no assignments

ryanatkn:
Makes `checkRemovePersona` fail if the persona has no assignments. This incidentally was already checked by `LeaveCommunity` but not `KickFromCommunity`. Adds tests to be sure.


src/lib/vocab/community/communityHelpers.server.ts
```diff
@@ -167,6 +165,7 @@ export const checkRemovePersona = async (
 		const adminAssignmentsCount = unwrap(
 			await repos.assignment.countAccountPersonaAssignmentsByCommunityId(community_id),
 		);
+		// TODO this fails if the persona has multiple roles
```
ryanatkn, line 168: think I found a bug, didn't fix in this PR

src/lib/vocab/persona/personaHelpers.server.ts
```diff
@@ -2,7 +2,15 @@ import {ADMIN_COMMUNITY_ID} from '$lib/app/constants';
...
-export const RESERVED_PERSONA_NAMES = new Set(['admin', 'ghost', 'docs', 'schemas', 'about']);
+export const RESERVED_PERSONA_NAMES = new Set([
+	'about',
+	'admin',
+	'api',
+	'docs',
+	'ghost',
+	'schemas',
+	'vocab',
+]);
```
ryanatkn, line 13: didn't seem worth its own PR, this adds `'api'` and `'vocab'` and alphabetizes

src/lib/vocab/community/communityHelpers.server.ts
greatbacon: I put a bug in the project list, not sure if you've already fixed this in another PR though lol

greatbacon (approved):

PR #703, opened 2023-02-09
simplify mutation destructuring

ryanatkn:
Since we've been refactoring to use mutation helpers, we stopped using the pattern of destructuring mutation `result` objects with a `$` prefix. This cleans up all of our older code to use the newer pattern of either destructuring from `result` without renaming, or using `result.value` directly.


greatbacon (approved):
🧹

PR #706, opened 2023-02-09
simplify env

ryanatkn:
cleans up env usage, continuing the work in #666


src/lib/config.ts
```diff
@@ -1,18 +0,0 @@
...
-
-const PUBLIC_DEPLOY_SERVER_HOST = import.meta.env
-	? import.meta.env.PUBLIC_DEPLOY_SERVER_HOST
-	: process.env.PUBLIC_DEPLOY_SERVER_HOST;
-
-export const API_SERVER_PORT = 3000; // same as in vite.config.js
-export const API_SERVER_HOST = `localhost:${API_SERVER_PORT}`;
-export const WEBSOCKET_URL_DEV = `ws://localhost:${API_SERVER_PORT}/ws`;
-export const WEBSOCKET_URL_PROD = `wss://${PUBLIC_DEPLOY_SERVER_HOST}/ws`;
-export const WEBSOCKET_URL = dev ? WEBSOCKET_URL_DEV : WEBSOCKET_URL_PROD;
```
ryanatkn, line 18: brings me great joy to delete this file

src/lib/infra/.env.production.default
```diff
@@ -1,5 +1,8 @@
 COOKIE_KEYS=latest-secret__older-secret__oldest-secret
 PUBLIC_DEPLOY_SERVER_HOST=server.name
+PUBLIC_WEBSOCKET_URL=wss://server.name:3000/ws
+PUBLIC_API_SERVER_HOSTNAME=localhost
+PUBLIC_API_SERVER_PORT=3000
```
ryanatkn, line 5: I think it's ok to keep these values un-DRY to this degree, where `server.name` and the port are duplicated. There are circumstances where they could be different, and I think it's the least confusing option because all of the values are in one place.

vite.config.js
```diff
@@ -3,11 +3,10 @@ import {sveltekit} from '@sveltejs/kit/vite';
 /** @type {import('vite').UserConfig} */
 const config = {
 	plugins: [sveltekit()],
-	envPrefix: 'PUBLIC_',
```
ryanatkn, line 6: It appears we can remove this now that we're no longer using `import.meta.env`, and instead we're using SvelteKit's env imports.

src/lib/config.ts
greatbacon: 🧹 💨

greatbacon (approved):

PR #713, opened 2023-02-17
refactor query to use events

ryanatkn:
Refactors the paginated query store to use events.

- adds `events` to the `ui`, not `app` like we discussed -- I first put it on `app` but it required threading it through everywhere, for no real reason -- it doesn't complicate `ui` that much, and conceptually it makes sense that the data events are co-located with the data
- adds the dev dependency [eventemitter3](https://github.com/primus/eventemitter3) (the browser doesn't give us a good way to do this with arbitrary data and types)
- handle `addEntity` through the events, not externally in the mutations

I have a big question in a comment below, but I think this PR is gtg.


src/lib/vocab/entity/entityMutations.ts
```diff
@@ -6,12 +6,18 @@ import {stashEntities, evictEntities, stashTies} from '$lib/vocab/entity/entityM
...
 	if (!result.ok) return result;
 	const {entities, ties} = result.value;
 	stashEntities(ui, entities);
 	stashTies(ui, ties);
+	// TODO construct this array during `stash`
+	const {entityById} = ui;
+	events.emit(
+		'stashed_entities',
+		entities.map((e) => entityById.get(e.entity_id)!),
+	);
```
ryanatkn, line 20: This works but it's very hacky, want to think about the best way to fix. My guess is we'll return an array from `stashEntities`.

src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -92,18 +85,6 @@ export const stashTies = (
 					mutated.add(query.data);
 				}
 			}
-
-			// TODO this doesn't handle queries across different actors --
-			// if the user has two actors in the same space but different permissions,
-			// they may need to see a different subset of entities.
-			// We'll need to rethink how this data gets updated in the queries
-			// after implementing entity-level permissions.
-			// See the "query key todo" in multiple places.
```
ryanatkn, line 101: This todo is covered in `createPaginatedQuery`

src/lib/ui/app.ts
```diff
@@ -18,9 +16,6 @@ export const LAST_SEEN_KEY = 'lastseen:';
...
 const KEY = Symbol();
 
-export const getApp = (): AppStores => getContext(KEY);
+export const getApp = (): App => getContext(KEY);
 
-export const setApp = (stores: AppStores): AppStores => {
-	setContext(KEY, stores);
-	return stores;
-};
+export const setApp = (stores: App): App => setContext(KEY, stores);
```
ryanatkn, line 21: I was in this module starting by implementing it as `AppEvents`, and cleaned up some stuff while I was here. Svelte changed `setContext` to return the value hence the shorthand.

src/lib/vocab/entity/entityMutations.ts
greatbacon: Cool, hacky in the leadup to alpha seems like an acceptable risk to me. 

One thought i had in regards to alternative 2 that I've been thinking is about how `stashEntities` & `stashTies` are currently logically distinct operations, but is there an opportunity in bundling them up kind of in the same way we have with our Service API?

Otherwise it sounds like the preferred long term state is the async buffer system?

greatbacon (approved):

PR #714, opened 2023-02-17
batch effects and make mutations synchronous

ryanatkn:
Implements the `batch` refactor described in #713 

- removes the error-prone and tedious `mutated` bookkeeping (in hindsight it was a poor choice I made to implement it like that)
- expects mutation calls into mutation helpers to be wrapped in `ui.mutate(() => ` -- we could hoist this if we change mutations a bit (but they'd lose control over `invoke`, which may be ok, at least as a default)
- makes all mutations synchronous -- any async effects can be added via `ui.addMutationEffect` which get flushed by `ui.mutate`

Future:

- do a store-level batch operation in `ui.mutate` to get atomic updates (so there can't be in-between invalid states, needs to be implemented in our store library, or we need to swap it for something else)


src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -41,20 +43,16 @@ export const stashEntities = (ui: WritableUi, $entities: Entity[]): void => {
 			}
 		}
 	}
+
+	if (added) {
+		ui.addMutationEffect(() => ui.events.emit('stashed_entities', added!));
+	}
```
ryanatkn, line 49: this entire PR led up to this

greatbacon (approved):

greatbacon:
Sorry, I just saw your message in Felt. I like `afterMutation`, I agree that I think it makes it's function clearer.

ryanatkn:
renamed here: https://github.com/feltjs/felt-server/pull/718

PR #725, opened 2023-03-08
rename persona_id field in API params to targetActor

ryanatkn:


ryanatkn:
I don't like the inconsistency with `targetActor` being the only place we use the prefix `target`.

An alternative is we rename `actor`.

greatbacon (approved):
I think having a `target` prefix is fine. It lets us keep our `actor` semantics through the system while provided what I think is an implicitly clearer API!

PR #728, opened 2023-03-15
add broadcast info to the service return type

ryanatkn:
Adds `BroadcastInfo` to the service return values under the `broadcast` key. Currently this is optional, but it'll be required once we migrate all of the services. This PR migrates a single service, `DeletePersona`.

followup:

- use the `BroadcastInfo` and scope `broadcast`


src/lib/vocab/hub/hubHelpers.server.ts
```diff
@@ -14,15 +14,24 @@ import {randomHue} from '$lib/ui/color';
...
 const log = new Logger(gray('[') + blue('hubHelpers.server') + gray(']'));
 
-export const cleanOrphanHubs = async (hubIds: number[], repos: Repos): Promise<void> => {
+/**
+ * Deletes hubs that have no remaining account actors.
+ * Returns the deleted hub ids, if any.
+ * @param hubIds
+ * @param repos
+ */
+export const cleanOrphanHubs = async (hubIds: number[], repos: Repos): Promise<null | number[]> => {
```
greatbacon, line 23: Should we spin up a test for this? We changed the return signature and nothing broke 😅

greatbacon (approved):
Had a question about a test, otherwise lgtm!

src/lib/vocab/hub/hubHelpers.server.ts
ryanatkn: Good idea, I wrote the unit test for the helper then then realized I couldn't test it without hacking around the services, so I expanded a test to ensure the hub gets orphaned: https://github.com/feltjs/felt-server/pull/728/commits/0e0bf634599ca194316202825727c5e17c091fe3

One thing we could do with the returned deleted hub ids is return it from actions like the `LeaveHubService`. Currently it just returns null.

PR #729, opened 2023-03-15
add Whiteboard view

ryanatkn:
Adds the `Whiteboard` view using [the felt-ui whiteboard](https://ui.felt.dev/library/Whiteboard).

copypaste+merge between these two modules:

- https://github.com/feltjs/felt-server/blob/main/src/lib/plugins/feltcoop/iframe/Iframe.svelte
- https://github.com/feltjs/felt-ui/blob/main/src/routes/sketch/whiteboard/%2Bpage.svelte

followup:

- persist the data
- fix bug where ephemera is discarded by actor (should be by tab id, but we don't have a tab id yet)


src/lib/plugins/feltcoop/iframe/Iframe.svelte
```diff
@@ -16,15 +17,19 @@
...
+				postMessage && // wait for init
+				v && // there may be no ephemera
+				v.space_id === $space.space_id && // scope to this space
+				v.actor !== $persona.persona_id // don't forward ephemera created by the user
+			) {
+				// TODO forward `actor: v.actor` if user allows it
+				postMessage({type: 'Ephemera', data: v.data}); // don't forward the space_id
+			}
+		}),
+	);
```
ryanatkn, line 32: this pattern is almost identical, but there's a subtle difference -- it's no longer batched. Previously multiple ephemera on the same animation frame would be batched, so only the last one would be processed. Now, each ephemera is processed, but we have to manually clean up the subscription. I made a Svelte REPL demo of the concept: https://svelte.dev/repl/69e1c9327ce847b0af642ed3163201da?version=3.57.0

greatbacon: Sweet, thanks for including a REPL demo!

greatbacon (approved):

PR #731, opened 2023-03-16
Clarifies some docs & fix some setup stuff.

greatbacon:


ryanatkn:
wondering if it's better to always stop pm2 (which fails on first deploy) and then start instead of trying to restart -- that way the pm2 restart count will be meaningful to see crashes

ryanatkn (approved):

PR #732, opened 2023-03-17
add LayoutContext

ryanatkn:
- fixes the prod bug
- adds `LayoutContext` mirroring `ViewContext` that's used by the main layout to mount the `Workspace`
- clarifies that the layout context and view context are two different things with two different types (the former has everything nullable except the persona) - plugin components can use both `getViewContext` and `getLayoutContext`, so we're starting to distinguish between space views and layout components (widget components being the name we've discussed for the other kind of not-view that includes `Mention` and `Link`)
- renames the `ViewContext` component to `ViewContextProvider` to disambiguate with the interface


src/lib/ui/Marquee.svelte
```diff
@@ -15,7 +15,7 @@
 	} = getApp();
 
 	export let hub: Readable<Hub>;
-	export let space: Readable<Space | null>;
+	export let space: Readable<Space> | null;
```
ryanatkn, line 18: type was wrong here and below

greatbacon (approved):

PR #736, opened 2023-03-20
Fix signup flow

greatbacon:
This PR fixes a bug in the sign up flow for new instances.

It modifies the query looking for adminHub to now return undefined if the hub wasn't found.

Plus an extra helper to do admin hub lookup & check persona counts to see if the instance is new.


src/lib/vocab/account/accountServices.ts
```diff
@@ -39,11 +39,13 @@ export const SignUpService: ServiceByName['SignUp'] = {
 		// TODO does this belong in `checkAccountName` above?
 		// TODO consider `const settings = repos.entity.filterByUrl('/instance');` (but scoped to admin?)
 		// should entities be scoped?  or /e/ to reference any path or id?
-		const adminHub = await repos.hub.loadAdminHub();
-		const allowedAccountNames = adminHub.settings.instance?.allowedAccountNames;
-		if (allowedAccountNames) {
-			if (!allowedAccountNames.includes(username.toLowerCase())) {
-				return {ok: false, status: 400, message: 'cannot create account'};
+		if (!(await isNewInstance(repos))) {
```
ryanatkn, line 42: `repos.hub.hasAdminHub`

PR #738, opened 2023-03-21
remove unwrap from role repo

ryanatkn:
continues the work from #730 for the role repo


src/lib/vocab/account/AccountRepo.ts
```diff
@@ -65,7 +64,7 @@ export class AccountRepo extends PostgresRepo {
 				account,
 				sessionPersonas: sessionPersonasResult.value,
 				hubs: hubsResult.value,
-				roles: rolesResult.value,
+				roles,
```
ryanatkn, line 67: this module shows how removing the wrapper layer is nice syntatically, we're just passing things through instead of dealing with wrappers (now using custom `Error` instances to handle things like API services)

greatbacon (approved):

PR #749, opened 2023-03-24
remove more unwrap calls

ryanatkn:
Removes `unwrap` from the session API - this was the last place we used it outside of tests.


greatbacon (approved):
🎉

PR #771, opened 2023-04-04
forward changed directories on entity changes

ryanatkn:
- adds a new helper `updateDirectories` that gets called by all of the entity-changing services to construct the list of directories that changed, and forwards them to the clients so notifications can use their `updated` field
- change `UpdateEntity` to `UpdateEntities` to match the other batched actions

followup:

- actually fix notifications


src/lib/vocab/entity/EntityRepo.test.ts
```diff
@@ -66,36 +66,6 @@ test__EntityRepo('entites return sorted by descending id', async ({repos, random
 	assert.is(entity0.entity_id, entities[2].entity_id);
 });
 
-test__EntityRepo('disallow mutating tombstones', async ({repos, random}) => {
```
ryanatkn, line 69: I'm pretty sure this was a mistake to block even admins from changing tombstones at the repo level. I think this kind of restriction is more of a governance concern. Removing it simplifies our entity queries.

src/lib/vocab/entity/EntityRepo.test.ts
```diff
@@ -66,34 +66,64 @@ test__EntityRepo('entites return sorted by descending id', async ({repos, random
...
+test__EntityRepo('erase entities', async ({repos, random}) => {
```
ryanatkn, line 69: I'm pretty sure this was a mistake to block even admins from changing tombstones at the level of repo queries. I think this kind of restriction is more of a governance concern. Removing it simplifies our entity queries.

Calling erase multiple times is now a no-op instead of an error, and tombstones can now be modified by admins and others with permission. (the implementation of which we'll add when we do entity-level permissions)

src/lib/vocab/entity/EntityRepo.test.ts
```diff
@@ -66,34 +66,64 @@ test__EntityRepo('entites return sorted by descending id', async ({repos, random
...
+		{data},
+	);
+	const {entity: entity3} = await random.entity(
+		undefined,
+		undefined,
+		undefined,
+		undefined,
+		undefined,
+		{data},
+	);
```
ryanatkn, line 90: these are some gnarly call signatures lol

src/lib/vocab/entity/entityActions.ts
```diff
@@ -229,7 +239,7 @@ export const EraseEntities: ServiceActionData = {
 	},
 	returns: 'Promise<EraseEntitiesResponseResult>',
 	route: {
-		path: '/api/v1/entities/:entity_id/erase',
+		path: '/api/v1/entities/erase',
```
ryanatkn, line 242: review this later to throw an error

greatbacon (approved):

PR #772, opened 2023-04-04
scope broadcasting to joined hubs

ryanatkn:
tests added in #806 and merged back here

scopes broadcasting to actors in hubs with a new `IBroadcast` interface and `Broadcast` class, which exposes an API to all services as `serviceRequest.broadcast`, similar to how `serviceRequest.session` works

adds the `ActorRecord` schema to fix some type issues we had -- there's related cleanup this pattern enables, where we were typecasting due to the type union limitations with `Pick` and repo columns

Followups:

- bug where switching to list after a message returns in the set, but doesn't get included in the query
- detect broken connections from the server side - https://github.com/websockets/ws#how-to-detect-and-close-broken-connections


src/lib/server/Broadcast.ts
```diff
@@ -11,12 +15,20 @@ import type {ApiResult} from '$lib/server/api';
 
 const log = new Logger(gray('[') + blue('broadcast') + gray(']'));
 
+// broadcast.onCreateHub(hub, actor)
+// broadcast.onDeleteHub(hub)
+// broadcast.onAddActor(hub, actor) - invited
+// broadcast.onRemoveActor(hub, actor) - leave+kick, maybe through a helper
+
+// TODO BLOCK if `send` is a broadcast class method, maybe have an interface for services that excludes it
+// broadcast.send()
```
ryanatkn, line 24: don't do this actually, have a separate helper that services do not have access to

src/lib/vocab/hub/hubServices.ts
```diff
@@ -135,6 +135,8 @@ export const CreateHubService: ServiceByName['CreateHub'] = {
 			: toDefaultSpaces(actor, hub);
 		const {spaces, directories} = await createSpaces(createSpacesParams, repos);
 
+		broadcast.createHub(hub_id, account_id, actor);
```
ryanatkn, line 138: probably async

greatbacon (approved):

PR #774, opened 2023-04-06
add optional column selection for filterDirectoriesByEntity

ryanatkn:
This changes `filterDirectoriesByEntity` to have an optional `columns` argument. It then migrates the pattern to the other places we scope columns in queries.

I got the types right to make this as DRY as possible from the caller side, but the implementation pattern casts in 2 places (including a third in one instance). I left TODOs on those -- I'd like to resolve that issue before adding the pattern everywhere. 

There's also an issue some callers have where the types aren't narrowing with `Pick` over a union -- not sure what to do there, I left todos, and I'll try at least once more before merging. Looks like `Extract` may be part of the answer but I don't see how right now - https://stackoverflow.com/questions/73921100/how-to-narrow-a-union-type


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -69,9 +70,12 @@ export class EntityRepo extends PostgresRepo {
 		return {entities, missing};
 	}
 
-	async filterDirectoriesByAccount(account_id: AccountId): Promise<Directory[]> {
-		const data = await this.sql<Directory[]>`
-			SELECT entity_id, data, persona_id, created, updated, space_id, path
+	async filterDirectoriesByAccount<T extends Partial<Directory> = Directory>(
```
ryanatkn, line 73: I did this one on accident at first, the one consumer doesn't use the columns but seems good to keep.

src/lib/vocab/account/accountHelpers.server.ts
```diff
@@ -1,9 +1,10 @@
-import type {AccountSettings} from '$lib/vocab/account/account';
+import type {Account, AccountSettings} from '$lib/vocab/account/account';
 
-// TODO refactor with persona column patterns
+// TODO generate from schema? see also elsewhere
+export type AccountColumns = Array<keyof Account>;
 export const ACCOUNT_COLUMNS = {
 	ClientAccount: ['account_id', 'name', 'settings', 'created', 'updated'],
-};
+} satisfies Record<string, AccountColumns>;
```
ryanatkn, line 7: this `satisfies` pattern was a good step to figure out -- it's a new TS construct that says "make sure it matches, but don't actually cast to the type", so as a consequence we get the `const` values of the types instead of the former `string[]` types, which we'll need for the proper fix -- I think we're close now.

src/lib/ui/EntityCell.svelte
```diff
@@ -6,14 +6,15 @@
...
+	import type {EntityColumn} from '$lib/vocab/entity/entityHelpers.server';
 
 	export let persona: Readable<AccountActor>;
 	export let entity: Readable<Entity>;
 	export let propertyName: string;
 
 	const {actions} = getApp();
 
-	$: value = $entity[propertyName as keyof Entity];
+	$: value = $entity[propertyName as EntityColumn];
```
ryanatkn, line 17: this can be imported from the client because it's a type despite it being in a `.server` file, but I'm wondering if we want to change it to `EntityProperty` and put it on the regular helpers, as the standard pattern.

src/lib/server/authorize.ts
```diff
@@ -36,7 +36,7 @@ export const authorize = async (
 	if (!params.actor) {
 		return {ok: false, status: 400, message: 'actor is required'};
 	}
-	const actor = await repos.persona.findById<Actor>(params.actor, ACTOR_COLUMNS.Persona);
+	const actor = await repos.persona.findById(params.actor, ACTOR_COLUMNS.Persona);
```
ryanatkn, line 39: callers no longer have to specify the type, they just have to pass columns and the return type should be properly scoped to just those columns, but in the runtime value and type

src/lib/vocab/space/spaceHelpers.server.ts
```diff
@@ -5,6 +5,7 @@ import {ApiError} from '$lib/server/api';
 import {Logger} from '@feltjs/util/log.js';
 import {blue, gray} from '$lib/server/colors';
 import type {Repos} from '$lib/db/Repos';
+import {ACTOR_COLUMNS} from '../actor/actorHelpers.server';
```
ryanatkn, line 8:
```suggestion
import {ACTOR_COLUMNS} from '$lib/vocab/actor/actorHelpers.server';
```

src/lib/vocab/hub/hubHelpers.server.ts
```diff
@@ -3,14 +3,15 @@ import {Logger} from '@feltjs/util/log.js';
...
-import type {ActionActor, ActorId, PublicActor} from '$lib/vocab/actor/actor';
+import type {ActorId, PublicActor} from '$lib/vocab/actor/actor';
 import type {Repos} from '$lib/db/Repos';
 import type {Role, RoleId} from '$lib/vocab/role/role';
 import type {Assignment} from '$lib/vocab/assignment/assignment';
 import {defaultAdminHubRoles, type RoleTemplate} from '$lib/app/templates';
 import type {Policy} from '$lib/vocab/policy/policy';
 import {ApiError} from '$lib/server/api';
 import {randomHue} from '$lib/ui/color';
+import {ACTOR_COLUMNS} from '../actor/actorHelpers.server';
```
ryanatkn, line 14:
```suggestion
import {ACTOR_COLUMNS} from '$lib/vocab/actor/actorHelpers.server';
```

greatbacon (approved):

PR #776, opened 2023-04-12
Make `path` in space editor update directory instead of space.

greatbacon:
Was goofing around and noticed a bug in the space editor where update the directory path still tried to update `path` on space, instead of the underlying directory.


ryanatkn (approved):
would be a little clearer to name the helper `updateDirectoryPath` but it's fine as is too

PR #782, opened 2023-04-15
Fix freshness

greatbacon:
This PR reactivates the "freshness" system for Felt! 🎉


src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -49,20 +50,25 @@ export const stashEntities = (ui: WritableUi, $entities: Entity[]): void => {
...
-			if (!freshnessByDirectoryId.get(entity_id)) {
-				setLastSeen(ui, entity_id, ($entity.updated || $entity.created).getTime());
-				setFreshnessByDirectoryId(ui, entity);
-			}
-			upsertFreshnessByHubId(ui, spaceById.get($entity.space_id)!.get().hub_id);
-			// Is the directory's space selected? If so we don't want a notification.
-			if (entity_id === $selectedSpace?.directory_id) {
-				updateLastSeen(ui, entity_id);
-			}
+		if ('directory' in $entity.data) {
```
greatbacon, line 53: This line was the piece that was preventing the system from triggering.

src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -49,20 +50,25 @@ export const stashEntities = (ui: WritableUi, $entities: Entity[]): void => {
...
+			directories.push(entity);
 		}
 	}
 
 	ui.afterMutation(() => ui.events.emit('stashed_entities', stashed));
+
+	for (const entity of directories) {
```
greatbacon, line 60: I had to move this directory logic to post entity mutation stashing because otherwise there's an underlying dependency on the `getEntityById` store inside `upsertFreshnessByHubId` that would fail since no stores are initialized until after the `stashed_entities` event.

I'm not sure if this is the actual right solution, but it is now functional. 🤷

ryanatkn (approved):
wahoo!

I'll follow up with some tweaks

PR #784, opened 2023-04-18
Add a new repo method & test

greatbacon:
This PR adds a new query in advance of broadcast scoping that returns all hub_id:actor_id pairs for a given account i.e. a list of the hubs the various actors of the account are in. Plus a test for good measure!


src/lib/vocab/actor/ActorRepo.ts
```diff
@@ -139,4 +139,12 @@ export class ActorRepo extends PostgresRepo {
 		`;
 		return data[0];
 	}
+
+	async filterHubActorsById(account_id: AccountId): Promise<[{hub_id: HubId; actor_id: ActorId}]> {
```
ryanatkn, line 143: the type `[{hub_id: HubId; actor_id: ActorId}]` should be `Array<{hub_id: HubId; actor_id: ActorId}>`

ryanatkn (approved):
great, thank you -- I'm thinking instead of `filterHubActorsById` maybe `filterHubActorsByAccount`, and insert it right after the existing `filterByAccount`?

PR #786, opened 2023-04-19
refactor directory freshness update

ryanatkn:
makes a couple tweaks to the directory freshness update - I left a todo about possibly using `'stashed_entities'`, doing so would allow us to externalize the freshness system, decoupling it from the mutations, but I'm not sure that's direction we want to go


greatbacon (approved):
🧹

PR #790, opened 2023-04-21
Fix updateFreshness issue

greatbacon:
This PR adds an additional logic check to keep the updateFreshness flow from inadvertently crashing entity view rendering.


src/lib/ui/uiMutationHelpers.ts
```diff
@@ -37,6 +46,7 @@ export const upsertFreshnessByHubId = (ui: WritableUi, hub_id: HubId): void => {
 export const setLastSeen = (ui: WritableUi, directory_id: EntityId, time = Date.now()): void => {
 	const {lastSeenByDirectoryId} = ui;
 	if (lastSeenByDirectoryId.has(directory_id)) {
+		log.error(`lastSeenByDirectoryId has already been set for directory ${directory_id}`);
 		throw Error(`lastSeenByDirectoryId has already been set for directory ${directory_id}`);
```
ryanatkn, line 50: duplicating the messages here and the 2 places above doesn't look right, maybe keep your change and instead of `throw Error(...` change those lines to `return;`?

ryanatkn (approved):

src/lib/ui/uiMutationHelpers.ts
greatbacon: that works for me

PR #797, opened 2023-04-28
fix view bug

ryanatkn:
Fixes the bug that happens when navigating away from certain views. The problem was that `viewContext` was getting updated to components that were mounted from the previous space, triggering queries with stale data.

This changes `ViewContext` to be a non-reactive plain object instead of having a store wrapper.

Views now always get recreated when *any* of the view context changes. This is accomplished with the `{#key` line in `ViewContextProvider`. So if you navigate to a new space, or the same space with a different actor, the view gets recreated. This seems less than ideal for switching actors, but we can try to address that later. At least it won't be a footgun for view authors -- however the core issue of views getting updated could be solved without changing the reactive model, I think.


src/lib/ui/ViewContextProvider.svelte
```diff
@@ -1,16 +1,21 @@
...
+	import type {Readable} from '@feltcoop/svelte-gettable-stores';
 
 	import type {Space} from '$lib/vocab/space/space';
 	import type {Hub} from '$lib/vocab/hub/hub';
 	import type {AccountActor} from '$lib/vocab/actor/actor';
 	import {setViewContext} from '$lib/vocab/view/view';
 	import {getApp} from '$lib/ui/app';
 
+	// TODO `ViewContext` is slightly misnamed compared to `LayoutContext`,
+	// because it's really `SpaceViewContext` -- maybe `SpaceContext`?
```
ryanatkn, line 11: any thoughts on this todo? should we make this `SpaceContext`?

greatbacon: I'm partial to SpaceContext

greatbacon (approved):

PR #799, opened 2023-04-28
change contextmenu to use felt-ui

ryanatkn:
Uses felt-ui's upstreamed contextmenu and deletes the felt-server version.

Uses the helper `toContextmenuParams` to construct contextmenu params with typesafety -- not using this means you don't get safety between Svelte components and their props.

- [x] update imports
- [x] delete the 5 modules that are now in felt-ui
- [x] fix the 3 bugs revealed by the types 
  - [x] add `actor` to `HubAvatar` to fix one of the bugs (causes some churn)
- [x] implement upstream and upgrade felt-ui
  - [x] https://github.com/feltjs/felt-ui/pull/293
  - [x] https://github.com/feltjs/felt-ui/pull/298
- [x] fix the problem caused by `contextmenu.ts` importing a Svelte file, causing our tests to fail - https://github.com/feltjs/felt-ui/pull/302 - this causes some unfortunate bubbling up of the link/text entry component dependencies, but it's fine until we get the correct fix in the tests to handle Svelte files


src/lib/ui/style.css
```diff
@@ -69,60 +71,6 @@ fieldset {
...
-.chevron {
-	position: relative;
-	height: 8px;
-}
-.chevron::before {
-	display: block;
-	content: '';
-	border: 4px solid transparent;
-	border-left-color: var(--text_color_light);
-}
```
ryanatkn, line 124: upstreamed

src/lib/ui/ui.ts
```diff
@@ -123,7 +125,9 @@ export const toUi = (
 	$session: ClientSession,
 	initialMobile: boolean,
 	components: {[key: string]: typeof SvelteComponent},
-	onError: (message: string | undefined) => void,
+	contextmenuLinkComponent: ComponentType<ContextmenuLinkEntry>,
+	contextmenuTextComponent: ComponentType<ContextmenuTextEntry>,
```
ryanatkn, line 129: I don't love that we have to construct `ui` with these two dependencies now, but we can fix during alpha or beta by making our tests handle Svelte files

src/lib/ui/ui.ts
```diff
@@ -354,7 +358,12 @@ export const toUi = (
 	const mainLayoutView = writable('<Dashboard />'); // TODO source this from the hub/space context (so routes can customize the UI)
 	const expandMainNav = locallyStored(writable(!initialMobile), 'expandMainNav');
 	const expandMarquee = locallyStored(writable(!initialMobile), 'expandMarquee');
-	const contextmenu = createContextmenuStore({layout, onError});
+	const contextmenu = createContextmenu({
+		linkComponent: contextmenuLinkComponent,
+		textComponent: contextmenuTextComponent,
```
ryanatkn, line 363: these are the unfortunate required deps as described in other comments

src/routes/+layout.svelte
```diff
@@ -64,9 +65,16 @@
...
 	);
-	const ui = toUi(data, initialMobileValue, components, (errorMessage) => {
-		actions.OpenDialog(toDialogData(ErrorMessage, {text: errorMessage}));
-	});
+	const ui = toUi(
+		data,
+		initialMobileValue,
+		components,
+		ContextmenuLinkEntry,
+		ContextmenuTextEntry,
```
ryanatkn, line 73: these are the unfortunate required deps as described in other comments

PR #800, opened 2023-04-29
add getting started diagram

ryanatkn:
Adds a bare-bones diagram that's an overview of the frontend and backend pieces.

It seems to me right now that we'll want to display each of the main diagrams in at least a couple places in the docs. I'm wondering if we want to use gen to create markdown doc files so these could be embedded in multiple places. Maybe we just copy-paste them? 

future:

- add `Plugin`s
- maybe add `Middleware`
- it'd be cool to link the diagram text to live docs pages on felt.dev


src/docs/getting-started.md
```diff
@@ -14,10 +14,32 @@ See [library-usage](./library-usage.md).
...
+			S[Services] -- function calls --> R[Repos]
+			P <-- data --> R
+			R -- data --> S
+		end
+		subgraph browser frontend
+			A[Actions] -- "API requests over http or websockets" --> S
+      S -- "API response data" --> A
+			C[Svelte components] <-- ui --> A
+		end
+	end
```
greatbacon, line 40:
```suggestion
%%{ init: { 'flowchart': { 'curve': 'monotoneY' } } }%%
flowchart	
        subgraph DB
            P[(Postgres)]
            click P "https://github.com/feltjs/felt-server/blob/main/src/docs/data-model.md"
            end
		subgraph Node backend
			S[Services] -- function calls --> R[Repos]            
			R -- queries --> 
            P -- data --> R
			R -- data --> S
		end
		subgraph browser frontend
			A[Actions] -- "API requests
            (http or websockets)" --> S
            S -- "API response data" --> A
			C[Svelte components] <-- ui --> A
		end	
```
This is a great first take!  

I took a shot at tweaking things; this form gives the diagram a more vertical layout (which I'm partial to for representing call stacks), cleans up the arrows a little, & I think keeps longer sentences from colliding with other things. 

Feel free to take as few or many of the changes as resonate with you!

greatbacon (approved):
I think a generated solution could be cool in the long term. For now I'd probably advocate for still keeping a single source of truth (SoT) on diagrams and maybe just linking to the diagram section of the SoT doc if needed in another doc?

ryanatkn:
That makes sense. Great changes btw

PR #801, opened 2023-05-01
Default hub setting

greatbacon:
This PR adds functionality for setting default hubs to assign new AccountActors to (via Admin settings).

There is currently no actual Admin UX for this, it's just configured using `edit hub` on the Admin hub.


src/lib/vocab/hub/hubMutations.ts
```diff
@@ -63,11 +63,19 @@ export const DeleteHub: Mutations['DeleteHub'] = async ({params, invoke, ui}) =>
...
+export const InviteToHub: Mutations['InviteToHub'] = async ({invoke, actions, ui}) => {
 	const result = await invoke();
 	if (!result.ok) return result;
 
 	const {assignment, actor} = result.value;
+	const {hub_id} = assignment;
+	const {hubById} = ui;
+
+	if (!hubById.has(hub_id)) {
+		const readHubResult = await actions.ReadHub({actor: actor.actor_id, hub_id});
```
greatbacon, line 75: This isn't technically part of the defaultHub invite flow, but I noticed in testing that when you invited an actor to a new hub it wasn't hydrating properly. We already had this logic as part of the `CreateAssignment` mutation, I think I just forgot to port it over properly when we added this action.

src/lib/vocab/actor/ActorRepo.ts
```diff
@@ -149,4 +149,17 @@ export class ActorRepo extends PostgresRepo {
 		`;
 		return data;
 	}
+
+	async filterAssociatesByHub(hub_id: HubId): Promise<PublicActor[]> {
+		const data = await this.sql<PublicActor[]>`
+			SELECT ${this.sql(ACTOR_COLUMNS.PublicActor.map((c) => 'act.' + c))}
```
ryanatkn, line 155: might as well make the columns a function param here, see `findByName` above as an example for getting the types right

src/lib/vocab/actor/actorServices.ts
```diff
@@ -105,6 +106,36 @@ export const CreateAccountActorService: ServiceByName['CreateAccountActor'] = {
 			directories.push(...defaultAdminSpaces.directories);
 		}
 
+		// once all initial hubs & assignments have been created properly
+		// check to see if there are any instance configured default hubs the actor needs assigned to
+		const adminHub = await repos.hub.loadAdminHub();
+		if (adminHub) {
+			const defaultHubIds = adminHub.settings.instance?.newActorDefaultHubs;
```
ryanatkn, line 113: looking at `adminHub.settings.instance.newActorDefaultHubs` and the existing `hub.settings.defaultRoleId`

wondering about consistent naming: `adminHub.settings.instance.initialHubs` and `hub.settings.initialRoles`

greatbacon: I think `initial` & `default` are on the same level of clarity for me.  I'll change the new setting to `defaultHubIds` for consistency though.

src/lib/vocab/hub/hubMutations.ts
ryanatkn: ok cool yeah this may be some of that defensive programming we talked about -- this looks right to me, if we fix it on the payload/broadcast side of things, then it won't hit this code path

ryanatkn (approved):

PR #805, opened 2023-05-04
typesafe dialogs

ryanatkn:


src/lib/app/schemas.ts
```diff
@@ -91,6 +90,4 @@ export const schemas = vocabSchemas.concat(
 		])
 		.filter(Boolean)
 		.map(toFeltVocabSchema),
-	// Include external schema dependencies.
-	toFeltVocabSchema(DialogDataSchema as VocabSchema),
```
ryanatkn, line 95: I didn't realize we were actually using the schema. I'll probably put it back, but it's not a big deal.

greatbacon (approved):

PR #807, opened 2023-05-11
generate a static vocab schema

ryanatkn:
Moves actions to `/src/vocab/actions/` and generates `/static/schemas/vocab.json` by assembling all of our vocab schemas into one. This means we're now publishing a canonical, machine-readable, language-agnostic schema for all of our main data objects and actions with the live server at `$HOST/schemas/vocab.json` and in the git repo at `/src/static/schemas/vocab.json`

Includes `$anchor` - https://json-schema.org/draft/2020-12/json-schema-core.html#anchor so we can reference schemas both via the expected-but-bleh `#/$defs/Actor` and the more convenient `#Actor`

Tweaks the doc page also.

followup:

- switch to `#Anchor` in our schemas, and remove the custom code to support the host-absolute URIs
  - maybe remove `name`
- generate an OpenAPI schema at `/schemas/openapi.json`


src/lib/vocab/vocab.gen.ts
```diff
@@ -0,0 +1,17 @@
...
+ * More: https://json-schema.org/draft/2020-12/json-schema-core.html#name-bundling
+ * @returns
+ */
+export const gen: Gen = async () => {
+	const schema = bundleSchemas(vocabSchemas);
+	return {
+		filename: '../../static/schemas/vocab.json',
+		content: JSON.stringify(schema),
+	};
+};
```
ryanatkn, line 17: usually gen files don't specify a `filename` and are co-located with their output, but in this case we don't want `vocab.gen.ts` to be served in `/static`

src/lib/util/schema.ts
```diff
@@ -1,6 +1,26 @@
...
+import type {VocabSchema} from '@feltjs/gro/dist/utils/schema.js';
 
 // see also $lib/util/ajv.ts
 
-export const toSchemaName = ($id: string): string =>
-	stripEnd(stripStart($id, '/schemas/'), '.json');
+// TODO upstream to gro? schema lib?
+export const bundleSchemas = (schemas: VocabSchema[]): VocabSchema => {
+	const schema: VocabSchema = {
+		$id: 'https://app.felt.dev/schemas/vocab.json',
```
ryanatkn, line 8: This shouldn't be hardcoded to `app.felt.dev` but to do that, we'd have to dynamically generate the bundled schema on build based on environment vars. Not a big deal til beta I think.

src/lib/util/schema.ts
```diff
@@ -1,6 +1,34 @@
...
+		$defs: structuredClone(schemas)
+			.sort((a, b) => parseSchemaName(a.$id)!.localeCompare(parseSchemaName(b.$id)!))
+			.reduce(($defs, schema) => {
+				const {name, $anchor, ...rest} = schema; // re-order properties for readability
+				$defs[schema.name] = {name, $anchor, ...rest};
+				return $defs;
+			}, {} as Record<string, VocabSchema>),
+	};
+	// TODO HACK we're not supposed to mutate schema $ref, upstream this change to our schema defs,
+	// moving from `/schemas/A` to `#A`
```
ryanatkn, line 21: added some followup info the PR text about this

greatbacon (approved):

PR #808, opened 2023-05-17
Clean up admin-home

greatbacon:
So this PR does a couple of light cleanups on the two outstanding home views.

I did have a question about the Query response though.

Currently querying relies on an underlying Set for managing the response set & keeping it up to date in the event of new entities. Nothing wrong there, but all of our view architecture relies on Arrays instead.  The PaginatedQuery also returns those arrays.

We've basically got a function that we've been copy/pasting around our views to do the Set => Array transform, is there a way to just back that into the query response object instead?


ryanatkn (approved):
Good question, I noticed that problem too. I'm trying to think about the optimal solution -- Svelte 4 is going to include iterator support, so Sets and Maps become much nicer to work with. It may be that plain arrays are best for the query data, and then we add additional bookkeeping of ids with a set or something. I made a note to follow up on it soon.

greatbacon:
> Good question, I noticed that problem too. I'm trying to think about the optimal solution -- Svelte 4 is going to include iterator support, so Sets and Maps become much nicer to work with. It may be that plain arrays are best for the query data, and then we add additional bookkeeping of ids with a set or something. I made a note to follow up on it soon.

Okay cool. I think for now just using the helper to at least keep things DRY will pave to road the alpha!

PR #810, opened 2023-05-23
OrderedCollections for lists

greatbacon:
This PR provides a first draft at utilizing OrderCollections for ordering lists. I have only done work on the `Todo` view so far.

Primary changes are as follows:

- Introduce a new EntityData type based on the OrderedCollections type from ActivityStreams.  The `orderedItems` attribute is an array of the entity_ids of the entities that are the dest_id of all `HasItem` ties to the orderedcollection, in the sequential order they are to be rendered in.
- Shifted the data structure of Todo to have a discrete OrderedCollection called `lists` that all other lists are tied to.
- ~Add a callbackFunction param to the EntityCreationDialogue that allows for a post creation function to be invoked on the newly created Entity (such as adding it's id to a `orderedItems` array~
- Introduced logic for adding/removing entities with a `HasItem` source_id relationship from the `orderedItems` array.
- Added buttons to TodoItems for moving an item up or down in the orderedItem array


src/lib/vocab/entity/entityMutationHelpers.ts
```diff
@@ -106,6 +106,8 @@ export const stashTies = (
 			// TODO this lookup is wrong, but what's the best design here?
 			// extract from mutation helpers into the query store?
 			// (currently there's no query store for non-paginated queries, that's a todo)
+			// the issue here is that only entities with a direct tie back to the directory
+			// provided to the query get stored, not the whole entity return set
 			const query = queryByKey.get(source_id);
```
greatbacon, line 111: So in the process of doing this work, I turned this stone over again. I don't think it's mission critical for alpha to get fixed, but we should probably dive into this sooner rather than later since it results in some unexpected behavior.

src/lib/plugins/feltcoop/todo/Todo.svelte
```diff
@@ -87,18 +127,20 @@
...
 						toDialogData(CreateEntityForm, {
 							done: () => actions.CloseDialog(),
-							entityName: 'todo',
+							postCreate: addToLists,
+							entityName: 'todo list',
 							actor,
 							hub,
 							space,
 							fields: {content: true},
+							ties: [{source_id: $listsCollection.entity_id}],
```
greatbacon, line 143: So this line keeps coming up as a "possibly undefined" error, even though I've got that if `$listsCollection` check up above. Is there something here I've forgotten/missed?

ryanatkn: It's a choice of TypeScript for more correctness, because we're in a callback `listsCollection` could have changed despite the `if` guard because some other scope now has that function reference. Due to Svelte's runtime characteristics, I'm pretty sure it's impossible for the user to click the thing when the condition is false (though maybe there's a frame of lag or something), but TS can't know that so it takes the conservative position.

We can't use `!` in Svelte markup, but you can hoist functions to the script to use it. Adding `?` is sometimes also a good fix. In this case I think I'd add `$listsCollection && actions.OpenDialog(`

src/lib/vocab/entity/entityMutationHelpers.ts
ryanatkn: right on I'll be thinking about it, good to figure out as we do a full view pass

src/lib/ui/CreateEntityForm.svelte
```diff
@@ -53,10 +56,13 @@
...
 			space_id: $space.space_id,
 			data: data as EntityData, // TODO avoid typecast, probably validation against type?
-			ties: [{source_id: $space.directory_id}],
+			ties,
 		});
 		status = 'success'; // TODO handle failure (also refactor to be generic)
 		if (result.ok) {
+			if (postCreate) {
+				postCreate(result.value.entities[0]);
+			}
```
ryanatkn, line 65: I think a callback is the right design because we can't use non-generic events for dialogs because we open them not declaratively but via the `OpenDialog` action. 

Seems worth considering if we should use `done` as a standard pattern for dialogs to pass data back to whoever opened them. I think that'd make it more convenient to both define and call. We should be getting type safety for free now, so I think this would work:

```
  export let done: ((entity?: Entity) => void) | undefined = undefined;
```

It may be preferred to have enumerated callbacks with specific names for clarity, but I suspect `done` takes care of most cases. I think `created` would be a better name than `postCreate` if we want a separate callback.

src/lib/ui/CreateEntityForm.svelte
```diff
@@ -11,15 +11,18 @@
...
 	const {actions} = getApp();
 
 	export let done: (() => void) | undefined = undefined;
+	export let postCreate: ((entity: Entity) => void) | undefined = undefined;
 	export let entityName = 'Entity';
 	export let actor: Readable<AccountActor>;
 	export let hub: Readable<Hub>;
 	export let space: Readable<Space>;
 	export let type = 'Collection';
+	export let ties = [{source_id: $space.directory_id}];
```
ryanatkn, line 25: There's a bug here where `ties` isn't reactive to changes in the `space` prop. To help avoid accidentally using `ties` in the future, I think I'd do this:

```ts
	export let ties: EntityId[] | undefined = undefined;

    $: finalTies = ties ?? [{source_id: $space.directory_id}];
```

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -50,3 +51,24 @@ export const checkEntityPath = (path: string): string | null => {
...
+ */
+export const constructOrderedCollection = (
+	orderedCollection: Entity,
+	ui: Ui,
+): Readable<Array<Readable<Entity>>> | null => {
+	//TODO add recovery logic i.e.
+	//if tere are entities in collection, but not orderedIteam array, append
+	//or if there are ids in the array, but not tied to collection; remove them
+	const {entityById} = ui;
+	if (orderedCollection.data.type !== 'OrderedCollection') return null;
```
ryanatkn, line 70: I think it's more future-proof (like if we add a subtype of `OrderedCollection`) and fits patterns elsewhere to duck type here, so this becomes:

```suggestion
	if (!('orderedItems' in orderedCollection.data)) return null;
```

If that doesn't typecheck, nevermind, but I think it does.

src/lib/plugins/feltcoop/todo/Todo.svelte
```diff
@@ -44,6 +57,33 @@
 		}
 	};
 
+	const initListsCollection = async () => {
+		//TODO init with hub actor not user actor
```
ryanatkn, line 61: oh yeah hmm this is an interesting wrinkle to think about, users having ownership of these top-level collections is obviously not what we intended but I'm not sure what the right answer is

src/lib/vocab/entity/entityHelpers.ts
ryanatkn: I was looking at ActivityStreams, something we haven't been considering is that `type` can actually be either a string or an array of strings in the spec. There's more to think through but duck typing feels correct knowing this.

src/lib/ui/CreateEntityForm.svelte
greatbacon: I think having a single `done` with an optional entity makes sense.  I was very excited when I realized I knew enough about JS to realize passing a callback was a solution here, so I'm glad you think it's the right design :D

src/lib/ui/CreateEntityForm.svelte
greatbacon: Looks like the type signature is gonna be a little more gnarly than just EntityId, but it'll fix the bug :D

src/lib/vocab/entity/entityHelpers.ts
greatbacon: Yeah, that makes sense. Duck typing is still new to me as opposed to explicit type checking, but I'll get there eventually :D

src/lib/plugins/feltcoop/todo/Todo.svelte
greatbacon: discussed this today & left some notes in the agenda

src/lib/ui/CreateEntityForm.svelte
```diff
@@ -11,17 +11,31 @@
...
+				| {
+						source_id: EntityId;
+						type?: string;
+				  }
+				| {
+						dest_id: EntityId;
+						type?: string;
+				  }
+		  >
+		| undefined = undefined;
```
ryanatkn, line 35: this can be shortened to use the params type:

```ts
import type {CreateEntityParams} from '$lib/app/actionTypes';
export let ties: CreateEntityParams['ties'] | undefined = undefined;
```

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -50,3 +51,24 @@ export const checkEntityPath = (path: string): string | null => {
...
+export const constructOrderedCollection = (
+	orderedCollection: Entity,
+	ui: Ui,
+): Readable<Array<Readable<Entity>>> | null => {
+	//TODO add recovery logic i.e.
+	//if tere are entities in collection, but not orderedIteam array, append
+	//or if there are ids in the array, but not tied to collection; remove them
+	const {entityById} = ui;
+	if (!('orderedItems' in orderedCollection.data)) return null;
+	if (!orderedCollection.data.orderedItems) return null;
```
ryanatkn, line 71: looks like this is redundant, and isn't needed for type narrowing:

```suggestion
	if (!orderedCollection.data.orderedItems) return null;
```

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -50,3 +51,24 @@ export const checkEntityPath = (path: string): string | null => {
...
 };
+
+/**
+ * Takes an entity with orderedItems in its data
+ * and constructs an ordered array of its items
+ * @param orderedCollection
+ * @param ui
+ * @returns null if invalid, otherwise an array of readable entities
+ */
+export const constructOrderedCollection = (
```
ryanatkn, line 62: Looks like this should return a plain array, not wrapped in a readable.

It also seems to do the same pattern as `lookupActor` and `lookupTies`, so I think `lookupOrderedItems` is a better name.

src/lib/plugins/feltcoop/todo/Todo.svelte
```diff
@@ -87,18 +109,28 @@
...
 					actions.OpenDialog(
 						toDialogData(CreateEntityForm, {
 							done: () => actions.CloseDialog(),
-							entityName: 'todo',
+							entityName: 'todo list',
 							actor,
 							hub,
 							space,
+							type: 'OrderedCollection',
 							fields: {content: true},
```
ryanatkn, line 132: is the default, can also be removed from `AddListButton`

```suggestion
							fields: {content: true},
```

src/lib/vocab/entity/entityServices.ts
```diff
@@ -158,11 +172,15 @@ export const DeleteEntitiesService: ServiceByName['DeleteEntities'] = {
...
+		const collections = await checkRemoveOrderedItems(repos, entityIds);
+
 		const deleted = (await repos.entity.deleteByIds(entityIds)).map((e) => e.entity_id);
 
 		await cleanOrphanedEntities(repos); // TODO probably return the ids here that got orphaned, and scope to the hub in the function
 
-		const entities = await updateDirectories(repos, entityIds);
+		const directories = await updateDirectories(repos, entityIds);
+
+		const entities = collections.concat(directories);
```
ryanatkn, line 183: I'd like to test this in a followup at least

src/lib/vocab/entity/entityHelpers.server.ts
```diff
@@ -47,3 +48,63 @@ export const updateDirectories = async (
...
+): Promise<Entity | void> => {
+	if (tie.type !== 'HasItem') return;
+	const source =
+		createdEntity.entity_id === tie.source_id
+			? createdEntity
+			: await repos.entity.findById(tie.source_id);
+	if (source && 'orderedItems' in source.data) {
+		if (!source.data.orderedItems) {
+			source.data.orderedItems = [];
+		}
```
ryanatkn, line 73: `orderedItems` isn't nullable so I think this is right:

```suggestion
	if (source?.data.orderedItems) {
```

src/lib/vocab/entity/entityServices.ts
```diff
@@ -87,7 +92,16 @@ export const CreateEntityService: ServiceByName['CreateEntity'] = {
 					'source_id' in tieParams
 						? {source_id: tieParams.source_id, dest_id: entity.entity_id}
 						: {source_id: entity.entity_id, dest_id: tieParams.dest_id};
-				ties.push(await repos.tie.create(source_id, dest_id, tieParams.type || 'HasItem')); // eslint-disable-line no-await-in-loop
+				const tie = await repos.tie.create(source_id, dest_id, tieParams.type || 'HasItem'); // eslint-disable-line no-await-in-loop
+				ties.push(tie);
+				const collection = await checkAddOrderedItem(repos, entity, tie); // eslint-disable-line no-await-in-loop
+				if (collection) {
+					if (entity.entity_id === collection.entity_id) {
+						entity.data = collection.data;
```
ryanatkn, line 100: this looks off to me because the `updated` field could have changed too, but does `checkAddOrderedItem` only need to be called if this is a `dest_id`? If so I don't think this condition can be hit, I'll look at this again with fresh eyes

ryanatkn (approved):
💯

ryanatkn:
I think EraseEntities needs the same treatment as DeleteEntities

PR #814, opened 2023-05-31
upgrade deps

ryanatkn:
Upgrades all deps, mainly felt-ui@0.56 which includes dark mode support: https://github.com/feltjs/felt-ui/pull/247

Something's wrong in the build now with the worker module, SvelteKit seems to be doing something I didn't expect with it and it errors. I disabled the password worker here and I'll fix in #816

Future:

- look at `.buttons`
- think about better helpers/patterns for loading from storage


src/lib/server/password.ts
```diff
@@ -16,7 +17,9 @@ const toScrypt = promisify(scrypt);
  * @returns salted and hashed password
  */
 export const toPasswordKey = async (passwordText: string): Promise<string> => {
-	return passwordWorker(passwordText)!;
+	// TODO fix
+	// return passwordWorker(passwordText)!;
+	return hashPassword(passwordText);
```
ryanatkn, line 22: This is the temporary hack, see the PR body.

PR #816, opened 2023-06-01
reuse worker for password hashing

ryanatkn:
Each `Worker` instance is a whole v8 process, overhead that's way too high to pay just to hash a single password. This follows up to #812 to reuse a single instance using a pool. It has to manage the pool's lifecycle, so we add the resource to the `ApiServer` alongside similar things like the db connection.

Future (beta or later):

- customizable pooling
- generic worker helpers
- can we handle errors without shutting down the pool? I don't see how we'd know which call to identify as the failure, I suppose it could return errors for all pending calls, but for now it'll just close everything down


src/lib/server/ApiServer.ts
```diff
@@ -121,11 +125,13 @@ export class ApiServer {
...
 		this.websockets.off('close', this.onWebsocketClose);
 		await Promise.all([
+			promisify(this.app.server.close).call(this.app.server),
 			this.websockets.close(),
 			this.broadcast.close(),
 			this.db.close(),
-			promisify(this.app.server.close).call(this.app.server),
+			this.passwordHasher.close(),
 		]);
+		this.server.close();
```
ryanatkn, line 134:
```suggestion
```

greatbacon (approved):
🧵 🦾

PR #817, opened 2023-06-02
add columns to hub repo methods

ryanatkn:
Adds `columns` to the hub repo methods and scopes usage.

Also refactors the account repo methods to fix type safety. We can default to the full set of columns, but not a subset like we had without losing type safety, and the sensitive `password` field makes it so we don't want to default to the full `Account`.


src/lib/vocab/account/AccountRepo.ts
```diff
@@ -59,7 +64,7 @@ export class AccountRepo extends PostgresRepo {
 
 	async findById<T extends AccountColumn>(
 		account_id: AccountId,
-		columns: T[] = ACCOUNT_COLUMNS.ClientAccount as T[],
+		columns: T[],
```
ryanatkn, line 67: We weren't getting type safety with this pattern whenever the default columns differ from the entire column set, so for the account, because of `password` sensitivity, it seems best to make the columns required. It's possible we'll want to do the same everywhere, but for now I kept the other default columns in other repos.

greatbacon (approved):

PR #819, opened 2023-06-03
refactor `UpdateHub` and forward the updated doc

ryanatkn:
Refactors one of the first actions we implemented - `UpdateHubSettings` becomes `UpdateHub` and now broadcasts the updated data. This ensures the `updated` field is correct and the result is properly broadcast (not relying just on params), and when we add more updatable fields, it'll follow the same pattern as the other tables.

I removed the optimistic update because it was just a proof of concept and isn't necessarily the UX we want.


src/lib/vocab/policy/permissions.ts
```diff
@@ -1,7 +1,7 @@
 export type PermissionName = keyof typeof permissions;
 export const permissions = {
 	Ephemera: 'Ephemera',
-	UpdateHubSettings: 'UpdateHubSettings',
+	UpdateHub: 'UpdateHub',
```
greatbacon, line 4: We're gonna need a db migration too that changes all policy.permission attributes that are `UpdateHubSettings` to `UpdateHub`. That might help with some of the permission issues you've been seeing.

ryanatkn: Ahh good catch I forgot about this

ryanatkn: added the migration

greatbacon (approved):

PR #821, opened 2023-06-06
update query actions and views

ryanatkn:
- delete `ReadEntities` and `QueryEntities` and rename `ReadEntitiesPaginated` to `ReadEntities`
- change old views to the new always-paginated querying pattern
- improve the `EntityExplorer` view
- add `EntityTrees` and `EntityTreeRoot` to complement `EntityTree` and improve it (these are all internal atm, not views)
- create `LoadMoreButton` and use it in all of the views (kinda janky because of `more` having false positives with our first pass implementation, there will be some followup work around this)
- omit `entity.data.type` when the value is `'Note'`, treat it as the default, so now the simplest entity data is just `{}` - this fits with our duck typing, and we'll continue to follow ActivityStreams to include `type` when it's specifically not a `'Note'` (like `'Collection'` and `'OrderedCollection'`)  -- also, we already need a translation step between our data and external ActivityStreams (like contructing `id`), and translating between internal and external representations is common in the ActivityPub implementations I've seen, so I don't see this as a problem, and it doesn't need a migration because it's treated the same if populated

followup

- some of the querying behavior changed for the worse but we need to discuss how related tie querying works, currently it gets the whole world, and everything here was changed to be paginated with a query per leaf, which is not a good solution for performance because of the number of queries and duplicate entity data loading, but we may need to rethink some things (which we expected), and maybe add related tie querying to `ReadEntitiesById`
- broken clear done (sometimes, the ties aren't populated on the client)
- look at these todos: https://github.com/feltjs/felt-server/pull/821/commits/0e83ea9fd763b08a124c357bb859256d1bb1ea68 and delete the stuff around `queryByKey` (and rename `paginatedQueryByKey`)
- `// TODO implement query by path`
- in `EntityTree` add/edit the source ids, enable dest ids to explore upstream, enable query params - maybe add or rename `QueryBuilder` or `QueryExplorer`?
- look at the pattern around `shouldLoadEntities`, should queries internally handle this state so we always create the objects?
- maybe upstream `--button_bg_active` hack?
- `more` should be returned by the server, likely via an n+1 fetch
- think about how to scale better
  - on the board, which currently eagerly loads everything - maybe only load when onscreen by default? also think about throttling, collection ties to a thread root
  - change `EntityTree` so it doesn't eagerly load past a configurable depth?
- collections and/or replies?
- keep the stashed_entities event system? what about ties for the queries? how to keep them live for any changes? see ```// TODO either add `stashed_ties`, add ties to the payload of `stashed_entities`, or rethink this pattern```


src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -53,22 +53,20 @@ export const checkEntityPath = (path: string): string | null => {
...
 	const {entityById} = ui;
-	if (!orderedCollection.data.orderedItems) return null;
-	//TODO assertion is not great, so how best to defensively lookup?
-	const orderdItems = orderedCollection.data.orderedItems.map((id) => entityById.get(id)!);
-	return orderdItems;
+	const entities = [];
+	for (const entity_id of entity.data.orderedItems) {
+		const entity = entityById.get(entity_id);
+		if (entity) {
+			entities.push(entity);
```
greatbacon, line 68: add warning

src/lib/ui/EntityChip.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+	import type {Readable} from '@feltcoop/svelte-gettable-stores';
+
+	import type {Entity} from '$lib/vocab/entity/entity';
+
+	export let entity: Readable<Entity>;
+</script>
```
greatbacon, line 7: add todo about contextmenu

greatbacon (approved):

PR #822, opened 2023-06-07
convert permission name to a schema

ryanatkn:
Converts the `PermissionName` definitions from an object to a schema, and uses it in the policy. Moves the module from `$lib/vocab/policy/` to the new `$lib/vocab/permission/` directory.

Benefits:

- automatic validation of policy data from clients
- automatic validation of data in the database during development (and prod if enabled, which we'll eventually make an option, but currently disable for performance reasons - this will catch migration oversights like this: https://github.com/feltjs/felt-server/pull/819#discussion_r1218261644
- automatic docs like other schemas
- removing the `permission.` usage means users don't need to import the module to reference permissions, which is more convenient and performant

Downsides:

- less ergonomic autocomplete, shows the type instead of a menu - this could always be added back on top of these changes, but I don't think it's worth the code size and extra way to do things


greatbacon (approved):
This looks really cool, thanks for cleaning up my string shortcut lol

I'm not sure if `permissions` should be a top-level vocab directory, I think they pretty squarely nest within the `policy` vocab. i.e. `../vocab/policy/permission/*`

ryanatkn:
I'm not sure either, my reasoning at the time was that it's used outside of policies, but I can see how they're nested in policy and possibly don't make sense outside of them. 

Hmm, maybe `permission` should be `name`, we called the type `PermissionName` which already stuck out as not fitting our normal patterns. The policy permission+data is what makes it unique, `permission` seems to be serving the same role as as a `name` if it's not its own vocabulary-level entity.

PR #825, opened 2023-06-20
flatten plugin directories and merge $lib/app into $lib/ui

ryanatkn:
Restructures some modules:

- removes the username-level plugin directories -- instead we'll publish everything in this repo as first-party that gets versioned with the rest of felt-server
- merges `$lib/app` into `$lib/ui` because of the ambiguity
- moves some modules to `$lib/util` like `url.ts` and `color.ts`


greatbacon (approved):
🥇

PR #826, opened 2023-06-20
Add some top level UX for managing instance settings

greatbacon:
Was doing some digging to confirm what instance settings we had & thought it might be nice to have an instance admin View that spells them out for other operators.

Probably some cool work to be done on dynamically populating the list based on attributes described in the schema & the input forms based on type. But for now this is functional and surfaces the available settings to instance operators lol


src/lib/plugins/feltcoop/instance-admin/InstanceAdmin.svelte
```diff
@@ -1,8 +1,57 @@
...
+			<!--TODO generically populate based on attribue keys/schema?-->
+			<Message status="inform"
+				>to allow any name, delete this setting; otherwise only strings in this array will be
+				allowed for account creation</Message
+			>
+			<PropertyEditor
+				value={instanceSettings.allowedAccountNames}
+				field="allowedAccountNames"
+				update={updateHubSetting}
+				deletable={true}
```
ryanatkn, line 35: looks like this doesn't work without adding ```
				parse={parseJson}
				serialize={serializeJson}```

src/lib/plugins/feltcoop/instance-admin/InstanceAdmin.svelte
```diff
@@ -1,8 +1,57 @@
...
+			/>
+			<PropertyEditor
+				value={instanceSettings.defaultHubIds}
+				field="defaultHubIds"
+				update={updateHubSetting}
+				deletable={true}
```
ryanatkn, line 41: looks like `deletable` isn't working, the `PropertyEditor` looks like it doesn't handle this context, I see a few improves to make that I can follow up on, and I don't think this code should change

ryanatkn (approved):

ryanatkn:
love it, I think there's one bug to fix

src/lib/plugins/feltcoop/instance-admin/InstanceAdmin.svelte
```diff
@@ -1,8 +1,57 @@
...
+		});
 </script>
 
 <div class="prose padded_xl">
 	<h1>instance admin</h1>
+	{#if instanceSettings}
+		<fieldset>
+			<legend>settings</legend>
+			<!--TODO generically populate based on attribue keys/schema?-->
+			<Message status="inform"
```
ryanatkn, line 27: this status is the default

PR #828, opened 2023-06-21
upgrade deps

ryanatkn:
small deps upgrade, svelte 4 was released shortly after I did this, and I'll do a followup soon with that for felt-ui first

also node 20 for us soon


src/routes/+layout.svelte
```diff
@@ -181,7 +181,11 @@
 			<slot />
 		{/if}
 		<DevmodeControls {devmode} />
-		<Dialogs {dialogs} on:close={() => actions.CloseDialog()} />
+		<Dialogs {dialogs} on:close={() => actions.CloseDialog()} let:dialog>
+			<div class="pane width_md">
+				<svelte:component this={dialog.Component} {...dialog.props} />
```
ryanatkn, line 186: felt-ui no longer forces a `.pane` wrapper, giving us more flexibility that we don't try to use yet

greatbacon (approved):

PR #829, opened 2023-06-23
Couple quick fixes to deployment process

greatbacon:
The {browser} import on the entityHelper was crashing in production, so I just set it to log no matter what.

Also I tweaked something in the deployment script to see if it would fix a longstanding issue related to npm crashing during deployment.


ryanatkn (approved):

ryanatkn:
sorry about that, the `browser` usage sneaking into server code isn't good, I'll see about properly fixing it after alpha gets going

PR #830, opened 2023-06-24
add server run check to CI

ryanatkn:
motivated by https://github.com/feltjs/felt-server/pull/829 - CI now runs the built `server.js` as a smoke test to make sure it can init and close correctly

expecting CI to fail until that PR is merged

it should probably detect a timeout on close to prevent that kind of bug causing CI to hang, but I tried for a bit and it started getting convoluted so I stopped


greatbacon (approved):
Looks like tests past, whoohoo!

PR #831, opened 2023-06-25
change forum buttons to links

ryanatkn:
Changes space routing to support the pattern used by e.g. forums, `/hub/space/123`, so they can show the forum-viewed details of entity `123`. Previously everything after `/hub/space/` was ignored.

Implements the pattern for only forums, and will need refactoring to make it more general, like extracting helpers.

followup

- other views, refactor
- support `entity.path`
- somehow persist the routing with the space nav, similar to what the hub nav does


src/lib/ui/TextInput.svelte
```diff
@@ -36,7 +36,7 @@
 <style>
 	.text_input {
 		display: flex;
-		flex: 1;
+		width: 100%;
```
ryanatkn, line 39: might break some things, but fixes chat

greatbacon (approved):

PR #833, opened 2023-06-26
Test http api

greatbacon:
This is a pretty simple MVP for doing HTTP endpoint testing in our suite.

Not really doing any core business logic testing, mostly just making sure routing & schema validation are working.

Figured we could workshop on this file, and then once we get a pattern in place we like, I'd do it for the rest of the actions in a couple of batches.


ryanatkn (commented):
Love this so much. I think it makes sense to follow the same pattern as `services.test.ts` here, looping through all services and constructing params automatically with the test helpers. Like `services.test.ts`  we should be able to get mostly automated smoke tests but with some hacks and extra handling for sessions. One additional hack we'll need here is ignoring GET requests with params until we figure that out. Looks like you got the hard part figured out with the cookie.

I think smoke tests make sense here -- the goal is to specifically test that the http endpoints and related schemas are valid. If specific behavior needs to be tested (like specific returned data or changes in the db), that belongs in repo/service unit tests or `servicesIntegration.test.ts`, there's no need to add the complexity of http on top of that.

I'm thinking the module could be something like `src/server/httpServices.test.ts` or `serviceHttpEndpoints.ts` or something

src/lib/server/serviceHttpEndpoints.test.ts
```diff
@@ -0,0 +1,73 @@
...
+			action.name !== 'SignIn' &&
+			action.name !== 'CreateAccountActor' &&
+			action.name !== 'SignOut'
+		) {
+			const params = await randomActionParams[action.name](random, {account, actor});
+
+			let url = `http://localhost:${TEST_PORT}${path}`;
+
+			//TODO this is pretty hacky, not sure what the more elegant solution is though
+			if (action.name === 'DeleteActor') {
```
greatbacon, line 38: This chunk of `if` gates is pretty hacky, but I wasn't quite sure of the best way to programmatically substitute the `:route_params`

src/lib/server/serviceHttpEndpoints.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+test__ApiServer.before(setupServer);
+test__ApiServer.after(teardownServer);
+
+for (const service of services.values()) {
+	const {action} = service;
+	const {path, method} = action.route;
+
+	if (method === 'GET' || method === 'HEAD') {
+		console.log('skipping GET/HEAD ', action.name);
```
greatbacon, line 20: I can't remember, is there a way in uvu to basically say "there's a test here but we haven't actually implemented it yet?"

ryanatkn (approved):
amaze, gogo checkbots

src/lib/server/serviceHttpEndpoints.test.ts
```diff
@@ -0,0 +1,59 @@
...
+	const path = action.route.path as string;
+	const method = action.route.method as ServiceMethod;
+
+	//
+	if (method === 'GET' || method === 'HEAD') {
+		test__ApiServer.skip(`perform service ${action.name}`, async () => {
+			assert.ok(true);
+		});
+		continue;
+	}
```
ryanatkn, line 28: I think this should be removed after this PR is in: https://github.com/feltjs/felt-server/pull/842

```suggestion
```

greatbacon: I'll just wait for your change to merge & double check the tests work on my end

ryanatkn: k cool, it's merged

PR #841, opened 2023-06-28
upgrade deps including Svelte 4

ryanatkn:
Upgrades deps including Svelte 4.

Includes a similar ESLint override hack that I added to the felt-ui PR. We probably just need to wait for a dependency and it'll go away, and linting will fail once it's fixed.


src/routes/+layout.svelte
```diff
@@ -170,7 +160,7 @@
 
 <SocketConnection {socket} url={PUBLIC_WEBSOCKET_URL} />
 
-<Themed {theme}>
+<Themed>
```
ryanatkn, line 163: the behavior here should be the same

src/lib/ui/About.svelte
```diff
@@ -1,5 +1,8 @@
 <script lang="ts">
 	import {PUBLIC_ADMIN_EMAIL_ADDRESS, PUBLIC_ADMIN_ICON, PUBLIC_GIT_HASH} from '$env/static/public';
+
+	// TODO maybe render this
+	// import {VERSION} from '@sveltejs/kit';
```
ryanatkn, line 5: this was just added to SvelteKit, wasn't sure about including it, the string is `'1.21.0'`

src/routes/+layout.svelte
```diff
@@ -162,15 +152,11 @@
 	]}
 />
 
-<svelte:head>
-	<link rel="shortcut icon" href="/favicon.png" />
-</svelte:head>
-
```
ryanatkn, line 168: noticed this code duplicated what's in `src/app.html`

src/lib/ui/About.svelte
greatbacon: ooh, I like that! Maybe even link to the project too?

greatbacon (approved):
🧹 ✨

src/lib/ui/About.svelte
ryanatkn: sweet yeah added that

PR #842, opened 2023-06-29
change from GET to POST for params-containing endpoints

ryanatkn:
Changes all params-having GET requests to POST and adds the known-issues.md doc. (the only non-params-having GET endpoint is `Ping`)

should make #833 less hacky

wasn't exactly sure where the best places to link the doc are, I added in contributing and getting started


src/lib/vocab/entity/entityActions.ts
```diff
@@ -138,8 +138,8 @@ export const ReadEntities: ServiceActionData = {
 	},
 	returns: 'Promise<ReadEntitiesResponseResult>',
 	route: {
-		path: '/api/v1/entities',
-		method: 'GET',
+		path: '/api/v1/entities/query',
+		method: 'POST',
```
ryanatkn, line 142: I added `/query` to everything to avoid collisions with the `CREATE` endpoints

greatbacon (approved):

PR #846, opened 2023-07-03
Update query params for searching from oldest or newest

greatbacon:
So this PR starts the work of querying entities either by their oldest or their newest (as opposed to our initial default of newest first).

I think I've got the queries on the backend working properly, but their seems to be some logic in the client query class that isn't playing nice & I'm not familiar enough with it to navigate, so I'd like some help :D


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -55,13 +55,14 @@ export class EntityRepo extends PostgresRepo {
...
 		entityIds: EntityId[],
+		orderBy: 'newest' | 'oldest' = 'newest',
 	): Promise<{entities: Entity[]; missing: null | EntityId[]}> {
 		if (entityIds.length === 0) return {entities: [], missing: null};
 		log.debug('[filterByIds]', entityIds);
 		const entities = await this.sql<Entity[]>`
 			SELECT entity_id, space_id, path, data, actor_id, created, updated 
 			FROM entities WHERE entity_id IN ${this.sql(entityIds)}
-			ORDER BY entity_id DESC
+			ORDER BY entity_id ${orderBy === 'newest' ? this.sql`DESC` : this.sql`ASC`}
```
ryanatkn, line 65: `orderBy` would make sense to me as the name if it were specifying the property or whole clause like SQL does, but it's used here more as a modifier to `ORDER BY entity_id`

since we're following JS semantics elsewhere with `find`/`filter`, I think `sort` makes more sense instead of `orderBy`

greatbacon:
> would be good to test this

I didn't want to write any tests until we had agreed on a pattern & api going forward

ryanatkn (approved):

src/lib/vocab/entity/EntityRepo.ts
ryanatkn: nm, rethought this

PR #847, opened 2023-07-03
Add `more` attribute to query response

greatbacon:
This PR adds a `more` attribute to our paginated query. A true indicates that there are more items in the page set to be loaded.


src/lib/vocab/entity/entityServices.test.ts
```diff
@@ -163,6 +167,7 @@ test_entityServices('read paginated entities by source_id', async ({repos, rando
 		}),
 	);
 	assert.is(filtered5.length, 1);
+	assert.not(more5);
```
ryanatkn, line 170: in general the `.not` should be avoided because it doesn't narrow the type, it's why we do `assert.ok(!` instead, just a quirk of the type system

src/lib/vocab/entity/entityServices.ts
```diff
@@ -36,7 +37,11 @@ export const ReadEntitiesService: ServiceByName['ReadEntities'] = {
 		const {hub_id} = await repos.space.findByEntity(source_id);
 		await checkHubAccess(actor, hub_id, repos);
 
-		const pageTies = await repos.tie.filterBySourceIdPaginated(source_id, pageSize, pageKey);
+		const extraPageSize = pageSize ? pageSize + 1 : DEFAULT_PAGE_SIZE + 1;
```
ryanatkn, line 40: just a little clearer: 

```suggestion
		const extraPageSize = (pageSize ?? DEFAULT_PAGE_SIZE) + 1;
```

src/lib/vocab/entity/entityServices.ts
```diff
@@ -36,7 +37,11 @@ export const ReadEntitiesService: ServiceByName['ReadEntities'] = {
...
+		const pageTies = await repos.tie.filterBySourceIdPaginated(source_id, extraPageSize, pageKey);
+		const more = pageSize ? pageSize < pageTies.length : DEFAULT_PAGE_SIZE < pageTies.length;
```
ryanatkn, line 42: the code here and the line above will be cleaner if you change the line above to `const {actor, source_id, pageSize = DEFAULT_PAGE_SIZE, pageKey, related} = params;`

alternatively, a `const finalPageSize = pageSize ?? DEFAULT_PAGE_SIZE` is another pattern we sometimes use

ryanatkn (approved):
yayayay

PR #848, opened 2023-07-04
Update entity paths with new parent name

greatbacon:
This code adds logic for when a space updates its name, to update all paths on entities owned by the space with the new space name at the `/root` of the path.

Does not update entities with no path


src/lib/vocab/action/actionTypes.ts
```diff
@@ -405,7 +405,6 @@ export interface CreateSpaceParams {
 	actor: ActorId;
 	hub_id: HubId;
 	name: string;
-	path: string;
```
greatbacon, line 408: Removed this attribute since spaces no longer have paths, that's all entities now.

ryanatkn:
this would be good to test

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -195,4 +195,20 @@ export class EntityRepo extends PostgresRepo {
...
+		oldPath: string,
+		newPath: string,
+		columns: T[] = ENTITY_COLUMNS.all as T[],
+	): Promise<Array<Pick<Entity, T>>> {
+		const oldPathQuery = `/${oldPath}%`;
+		const entities = await this.sql<Array<Pick<Entity, T>>>`		
+			UPDATE entities e
+			SET path = regexp_replace(path,${oldPath},${newPath},1,1)
+			WHERE space_id=${space_id} and path LIKE ${oldPathQuery}
+			RETURNING ${this.sql(columns.map((e) => 'e.' + e))};
```
ryanatkn, line 210: This isn't going to scale and it'll sometimes forward clients a lot of entities they didn't ask for, but I don't see how we're going to patch the `updated` field specifically on the client without some special care. Will Postgres let us set the `updated` field directly without the recalculation? That way we could set the `updated` property of the entities to the directory's, and then just return the directory entity to clients.

greatbacon: Entities with updated paths are also going to need to know those new path values as well, unless you're assuming that's a calculation that is going to also be done client side?

I'm not sure if it would be 'easier', but we could also look into the fact that we're hardcoding a piece of data in the `entity.path` attribute that could potentially be normalized via a `space_id` lookup and just prepending the `space.name` to the root of the path.

ryanatkn: Yeah I think client side is the only tenable option. Some views may add paths to every entity, and forwarding the entire space's contents to everyone isn't going to work. We also don't track entity subscriptions on the server, so that leaves client-side handling as the only option. If we can forward the correct `updated` field that matches the directory and all changed entities it contains, then we can do the full calculation on clients.

That relies on being able to manually set the precise `updated` field on those entities, or somehow ensure that Postgres updates the directory and its entities atomically so they have an identical `updated` calculated automatically. I don't want to rely on "probably works if the db is fast enough", so we need to research this in Postgres.

ryanatkn: update: we talked, will normalize

greatbacon:
Closing this PR in favor of TBD newer one that implements normalization work around entity paths

PR #849, opened 2023-07-04
Update docs

greatbacon:
This PR adds a couple of missing things to our /docs

For schema attributes that utilize $ref, it uses the stripped schema name for the type & then in-line links to the schema doc (previously it was just rendering `unknown` for these attribute types)

For service actions the docs now also show the http route (both method & path)


src/routes/docs/+page.svelte
```diff
@@ -120,6 +120,13 @@
 						<code class="name">{actionData.name}</code>
 						<small class="type">{actionData.type}</small>
 					</div>
+					{#if actionData.type !== 'ClientAction'}
+						<div class="property">
+							<span>path</span>
```
ryanatkn, line 125: what do you think of the label "http" here? since it's path+method rendered

greatbacon: what about clarifying with `http endpoint`?

PR #852, opened 2023-07-05
Cleanup endpoints

greatbacon:
Gets endpoint paths on the same page & utilizes Vocab component in some new doc pieces.


src/lib/vocab/policy/policyActions.ts
```diff
@@ -55,7 +55,7 @@ export const ReadPolicies: ServiceActionData = {
 	},
 	returns: 'Promise<ReadPoliciesResponseResult>',
 	route: {
-		path: '/api/v1/hub/roles/:role_id/policies/query',
+		path: '/api/v1/roles/:role_id/policies/query',
```
greatbacon, line 58: So I ended up having to keep the role in the path here & the hub in the roles query path below because I believe `inject` was then parsing the `query` at the end as a URL param & breaking things.

src/lib/ui/SchemaInfo.svelte
```diff
@@ -24,7 +25,7 @@
 			<small class="type">
 				{#if schema.$ref}
 					{@const name = toSchemaName(schema.$ref)}
-					<a href="#{name}">{name}</a>
+					<Vocab name={name || 'missing name'} />
```
ryanatkn: saw you did this after opening #852

src/lib/vocab/space/spaceActions.ts
```diff
@@ -71,7 +71,7 @@ export const ReadSpaces: ServiceActionData = {
 	},
 	returns: 'Promise<ReadSpacesResponseResult>',
 	route: {
-		path: '/api/v1/hubs/:hub_id/spaces/query',
+		path: '/api/v1/spaces/query',
```
ryanatkn, line 74: the old version matches `ReadRoles` - `path: '/api/v1/hubs/:hub_id/roles/query',` - including the hub_id here makes sense right?

ryanatkn (approved):
# 🧹 

had one q

src/lib/vocab/policy/policyActions.ts
ryanatkn: Weird I don't know what could be going on there. Related to my other comment, I'm wondering if we should just simplify these to `/api/v1/policies/query` so we could potentially have any query params we want, and make them optional in some cases

greatbacon: Yeah, I think ideally trimming out the query param stuff not directly related to the vocab action at hand would be my preferred way of doing it, there's just this bug with `inject`. I think I have an idea for a different work around though so I'll go for that instead & just strip the paths down.

PR #855, opened 2023-07-06
refactor http tests

ryanatkn:
refactors the http tests a bit

- test `SignOut` instead of bypassing
- improve the error message if something fails, include status/message
- flatten the loop
- return the `session` from the init helper because it's there (even though it's not used with current code)
- in the init helper, create the actor before signing in, so the returned session is fully populated


src/lib/server/serviceHttpEndpoints.test.ts
```diff
@@ -14,38 +18,33 @@ const test__serviceHttpEndpoints = suite<TestServerContext>('serviceHttpEndpoint
...
-			const suffix = params ? inject(path, params) : path;
-			const url = `http://localhost:${TEST_PORT}${suffix}`;
-			const body = method === 'GET' || method === 'HEAD' ? null : JSON.stringify(params || {});
-
-			const res = await fetch(url, {
-				method,
-				headers: {
-					'content-type': 'application/json',
-					Cookie: auth,
-					Accept: 'application/json',
```
greatbacon, line 43: This might not break tests, but we need this `Accept` header in the wild for things to behave properly, so I think we should keep it in the test.

ryanatkn: gotcha ty I didn't know that, will revert

PR #864, opened 2023-07-12
update tagline and simplify some docs

ryanatkn:
- updates the tagline
- deletes `src/infra/README.md` and inlines the two sentences on the main readme
- moves `src/db/README.md` to `src/docs/database.md` (this was the only other readme besides the root)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@feltjs/felt-server",
   "version": "0.1.0",
   "license": "MIT",
-  "description": "server for Felt, a tool for building and maintaining communities 💚",
+  "description": "a programmable platform for hobbyists and human-scale communities on the web 💚",
```
greatbacon, line 5: do we want `... on the web`? I know the platform is a web platform, but I think we've discussed before that we don't want the communities themselves to be "just" web communities.

ryanatkn: good point, I wasn't sure about this but added it because we had the characters to spare -- let's look at it tomorrow

greatbacon (approved):

PR #865, opened 2023-07-12
Normalize entity paths

greatbacon:
This PR normalizes path storage behavior and locks down some assumptions related to the path system.

1.) The `space` path is still stored in the directory entity, but should now be kept 1:1 with the space name.
2.) Entities no longer also include the full space + entity path, but instead their path is assumed to be attached to the `space` path
3.) All entity paths must be unique within their owning space.


src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
 		return data[0];
 	}
 
-	async findByPath(hub_id: HubId, path: string): Promise<Entity | undefined> {
-		log.debug('[findByPath]', hub_id, path);
+	/**
+	 * This query looks for all directories in a given hub that might also have the given path
+	 * @param hub_id : the hub to search for directories
+	 * @param path  : the path being checked for collisions
+	 * @returns : an empty set if no collisions, or an entity
```
ryanatkn, line 45: TSDoc syntax is a little weird, it requires a `-` between the `@param` name and summary and has nothing after `@returns` - https://tsdoc.org/

src/lib/plugins/todo/Todo.svelte
```diff
@@ -30,13 +30,13 @@
...
 
-	$: listsPath = $directory.path + '/lists';
+	const listsPath = '/lists';
 	$: listsCollection = $entities?.value.find((e) => e.get().path === listsPath);
 
 	$: ({space_id, directory_id} = $space);
 	$: ({actor_id} = $actor);
 
-	$: if ($entities?.value.length && !listsCollection) {
+	$: if ($entities?.value && !listsCollection) {
```
ryanatkn, line 39:
```suggestion
	$: if ($entities && !listsCollection) {
```

src/lib/vocab/space/spaceServices.ts
```diff
@@ -52,13 +51,29 @@ export const UpdateSpaceService: ServiceByName['UpdateSpace'] = {
...
+		log.debug('[updateSpace] updating space');
 		const {space_id, actor, ...partial} = params;
 		const space = await repos.space.findById(space_id);
 		if (!space) {
 			return {ok: false, status: 404, message: 'no space found'};
 		}
 
 		await checkPolicy('update_space', actor, space.hub_id, repos);
+
+		if (HOME_VIEWS.includes(space.view)) {
```
ryanatkn, line 63: In `spaceHelpers.ts` we have `canDeleteSpace`, I think it'd be helpful for self-documentation to extract a helper `canUpdateSpace`, could also then be used on the client

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
 		return data[0];
 	}
 
-	async findByPath(hub_id: HubId, path: string): Promise<Entity | undefined> {
-		log.debug('[findByPath]', hub_id, path);
+	/**
+	 * This query looks for all directories in a given hub that might also have the given path
+	 * @param hub_id - the hub to search for directories
+	 * @param path  - the path being checked for collisions
```
ryanatkn, line 44:
```suggestion
	 * @param path - the path being checked for collisions
```

src/lib/vocab/entity/entityServices.ts
```diff
@@ -145,6 +153,14 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
 					if (errorMessage) throw new ApiError(400, errorMessage);
 				}
 
+				if (path) {
```
ryanatkn, line 156: This belongs with the block above, your `if (path) {` is better than the `if (typeof path === 'string') {` but the contents should be siblings.

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
...
+	 * @returns an empty set if no collisions, or an entity
+	 */
+	async findByHubPath(hub_id: HubId, path: string): Promise<Entity | undefined> {
```
ryanatkn, line 47: This name is confusing because it's only directories that it finds. Maybe `findDirectoryByPathInHub`?

Related for naming is `findBySpacePath` and `findByEntityPath` which read strangely to me, I think `findByPathInSpace` and `findByPathWithSibling` are clearer?

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
...
 			ON s.directory_id=e.entity_id AND e.path=${path}
 			WHERE s.hub_id=${hub_id}
 		`;
-		log.debug('[findByPath] result', data);
+		log.debug('[findByHubPath] result', data);
+		return data[0];
+	}
+
+	/**
+	 * This query looks for all entities in a given space that might also have the given path
```
ryanatkn, line 61: missing a period on this line and the other leading lines of the other 2 functions

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
...
+			SELECT e.entity_id, e.space_id, e.path, e.data, e.actor_id, e.created, e.updated
+			FROM entities e			
+			WHERE e.space_id=${space_id} AND e.path=${path}
+		`;
+		log.debug('[findBySpacePath] result', data);
+		return data[0];
+	}
+
+	/**
+	 * This query looks for all entities in the same space as the given entity that might also have the given path
```
ryanatkn, line 79: Missing a period on this line and the other leading lines of the other 2 functions' comments, it's also more than 100 characters wide for this one. 

The "this query" feels noisy and the "all entities" reads as misleading to me, "any entity" seems clearer - so instead of "This query looks for all entities", "Finds any entity"

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
...
+	 * If it returns an empty set, no other entities in the hub share that path.
+	 * @param entity_id - the entity to search for other entities
+	 * @param path - the path being checked for collisions
+	 * @returns an empty set if no collisions, or an entity
```
ryanatkn, line 83: these are misleading, there's no set returned, it's `undefined`

src/lib/vocab/entity/entityServices.ts
```diff
@@ -145,6 +153,14 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
...
+					log.debug('[updateEntity] validating entity path uniqueness');
+					const existingEntityWithPath = await repos.entity.findByEntityPath(entity_id, path);
+					if (existingEntityWithPath) {
+						throw new ApiError(409, 'an entity with that path in this space already exists');
+					}
+				}
+
```
ryanatkn, line 163: This will make the space name go out of sync with the path. Let's talk about what the source of truth should be, I'm not sure if the best UX is to only let users change names, and disallow path editing, or if we want bidirectional editing and parse out the space name from the directory path. Also related is name/path normalization, like lowercasing.

src/lib/vocab/space/spaceServices.ts
```diff
@@ -52,13 +51,29 @@ export const UpdateSpaceService: ServiceByName['UpdateSpace'] = {
...
+			throw new ApiError(405, 'Home spaces cannot be updated');
```
ryanatkn, line 64: following the same pattern as below with deleted and in other services:

```suggestion
			throw new ApiError(405, 'cannot update home space');
```

src/lib/vocab/entity/entityServices.ts
```diff
@@ -145,6 +153,14 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
 					if (errorMessage) throw new ApiError(400, errorMessage);
 				}
 
+				if (path) {
+					log.debug('[updateEntity] validating entity path uniqueness');
+					const existingEntityWithPath = await repos.entity.findByEntityPath(entity_id, path);
```
ryanatkn, line 158: I'm thinking to simplify things in this function, we should go ahead and move the line `const entity = await repos.entity.findById(entity_id);` above to the top of the function body after `const {entity_id, data} = doc;`. It's also an error if the entity doesn't exist, which the line with `entity?.data.orderedItems &&` doesn't handle correctly.

We then can delete the more complex `findByEntityPath` repo method because we'll have the space id.

This way we'll have the entity that we're trying to update loaded for whatever future purposes, now that we have 2 cases. We'll definitely have more reasons to load the entity to do validation.

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,60 @@ export class EntityRepo extends PostgresRepo {
 		return data[0];
 	}
 
-	async findByPath(hub_id: HubId, path: string): Promise<Entity | undefined> {
-		log.debug('[findByPath]', hub_id, path);
+	/**
+	 * This query looks for all directories in a given hub that might also have the given path
+	 * @param hub_id - the hub to search for directories
+	 * @param path  - the path being checked for collisions
+	 * @returns an empty set if no collisions, or an entity
```
ryanatkn, line 45: These are misleading, there's no empty set returned, it's `undefined`.

It's also mixing the semantics of "collisions" with "find" functionality. If we just want collisions, we should rename the functions and optimize them to not return data. We may want them for querying in the future and the optimization doesn't seem important, but it seems weird to have docs that reference collisions here - the line `the path being checked for collisions` doesn't make sense if I'm querying. If it's just a find function, then I don't think a docstring adds any useful information, saying `the path to look for` is just noise.

src/lib/vocab/entity/EntityRepo.ts
ryanatkn, line 61: Missing a period on this line and the other leading lines of the other 2 functions' comments.

The "This query" feels noisy, it's like the "in the system" of the vocab docs. Also the "all entities" reads as misleading to me because we don't allow collisions, "any entity" seems clearer - so instead of "This query looks for all entities", "Finds any entity".

src/lib/vocab/space/spaceHelpers.ts
```diff
@@ -20,6 +20,7 @@ export const parseSpaceIcon = (value: string): Result<{value: string}, {message:
 };
 
 export const HOME_PATH = '/home';
+export const HOME_VIEWS = ['<AdminHome />', '<Home />', '<PersonalHome />'];
```
ryanatkn, line 23: Let's talk about trying to remove this, any view inspection logic like this is a hack that we've used sparingly, and in this case there's at least one weird bug where if you add a home to any space, you won't be able to edit it.

We may need to change how the home space name is mirrored to the hub, I think we'll probably want `home` or something instead to do our detection and syncing with directories.

src/lib/vocab/entity/entityServices.ts
ryanatkn, line 158: This is the only place we use `findByEntityPath`, and with the changes in #870 we'll have the `entity` loaded for the common case, which means we'll have `space_id` and can use `repos.entity.findBySpacePath` and delete this more complex repo method.

src/lib/vocab/entity/entityServices.ts
greatbacon: So I know we chatted about possibly making this bi-directional, but I think actually just blocking path updates on directories makes for a simpler system, especially since we'll have to add `spaces` to the response object for this function to keep client values in sync.

src/lib/vocab/space/spaceServices.ts
```diff
@@ -52,14 +51,34 @@ export const UpdateSpaceService: ServiceByName['UpdateSpace'] = {
...
 		const {space_id, actor, ...partial} = params;
 		const space = await repos.space.findById(space_id);
 		if (!space) {
 			return {ok: false, status: 404, message: 'no space found'};
 		}
 
 		await checkPolicy('update_space', actor, space.hub_id, repos);
-		const updatedSpace = await repos.space.update(space_id, partial);
+
+		if (HOME_VIEWS.includes(space.view)) {
```
ryanatkn, line 63: this could become `isHomeSpace(space)` (and rename the existing `isHomeSpace to `isHomeDirectory`)

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,42 @@ export class EntityRepo extends PostgresRepo {
...
-		log.debug('[findByPath]', hub_id, path);
+	/**
+	 * Finds the directory in a given hub with the given path.
+	 * If it returns nothing, no other directories in the hub share that path.
+	 *
+	 * @param hub_id - the hub to search for directories
+	 * @param path - the path being checked for collisions
+	 * @returns the matching directory, if any
+	 */
+	async directoryByHubPath(hub_id: HubId, path: string): Promise<Entity | undefined> {
```
ryanatkn, line 49: should be `findDirectoryByHubPath`

src/lib/vocab/entity/entityServices.ts
ryanatkn: good stuff, yeah I misjudged, our goal was the simplest

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,42 @@ export class EntityRepo extends PostgresRepo {
...
 	}
 
-	async findByPath(hub_id: HubId, path: string): Promise<Entity | undefined> {
-		log.debug('[findByPath]', hub_id, path);
+	/**
+	 * Finds the directory in a given hub with the given path.
+	 * If it returns nothing, no other directories in the hub share that path.
+	 *
+	 * @param hub_id - the hub to search for directories
+	 * @param path - the path being checked for collisions
```
ryanatkn, line 46:
```suggestion
	 * @param path - the path to match
```

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -38,16 +38,42 @@ export class EntityRepo extends PostgresRepo {
...
-		log.debug('[findByPath] result', data);
+		log.debug('[directoryByHubPath] result', data);
+		return data[0];
+	}
+
+	/**
+	 * Finds any entities in a given space that might also have the given path.
+	 * If it returns nothing, no other entities in the space share that path.
+	 * @param space_id - the space to search for entities
+	 * @param path - the path being checked for collisions
```
ryanatkn, line 66:
```suggestion
	 * @param path - the path to match
```

src/lib/vocab/space/spaceMutations.ts
```diff
@@ -18,11 +19,21 @@ export const DeleteSpace: Mutations['DeleteSpace'] = async ({params, invoke, ui}
...
+	let $directory: Entity | undefined = undefined;
+	if (params.name) {
+		const directory = entityById.get(space.directory_id);
+		$directory = directory?.get();
+		if ($directory) {
+			$directory.path = `/${params.name}`;
+			$directory.updated = space.updated;
+		}
+	}
+	ui.mutate(() => stashSpaces(ui, [space], $directory ? [$directory] : undefined));
```
ryanatkn, line 36: I think we should forward the directory in the API instead of doing this

greatbacon: Since we're not always going to have an updated directory from this action, how to you recommend we structure the API?

ryanatkn (approved):

src/lib/vocab/space/spaceMutations.ts
ryanatkn: We talked about this, will shake this out after the entity/directory/tie changes this week.

PR #866, opened 2023-07-13
fix prod routing in polka and sveltekit

ryanatkn:
Works around the tiered middleware issue with the SvelteKit handler in Polka 0.5.2. See #859 for why we downgraded Polka.


src/docs/deploying-production.md
```diff
@@ -46,6 +46,35 @@ To redeploy:
 gro deploy
 ```
 
+## Test production locally
```
ryanatkn, line 49: should this title be different? `Run production locally`?

greatbacon: I like test, since it implies this is not a standard pattern to follow for deployments.

greatbacon (approved):

greatbacon (approved):
Just confirmed the fix works on staging too! :D :D :D

PR #870, opened 2023-07-18
fix bug updating entity orderedItems

ryanatkn:
Fixes a bug updating `entity.orderedItems`. Previously the code allowed updates to clobber `orderedItems` because it only ran its check if the new data had an `orderedItems` property. If the property was omitted but `data` was provided, the existing `orderedItems` would be lost and ties would get out of sync.

Also checks in the same place if it's trying to update directory data or change an entity into a directory. We were previously relying on permission checks to disallow these (hence the updated status codes in the tests), but I think it's a footgun to allow admins or anyone to corrupt data like that. If you agree I'll make a note to do the same thing for erase and delete in a followup. (403s for those data integrity checks are the red flag)

For #865 we may similarly disallow updating `path` for directories, or we'll need to sync the space name here, or we'll need to let them be decoupled - I left comments in #865, needs discussion. (update: we talked, syncing will probably be easiest and best)

`dequal` was already a dev dependency because of Svelte and others, this does make it a runtime dependency on the server, but we'll need it for other things.


src/lib/vocab/entity/entityServices.ts
```diff
@@ -124,16 +126,13 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
 			params.entities.map(async (doc) => {
 				const {entity_id, data} = doc;
 
+				const entity = await repos.entity.findById(entity_id);
+				if (!entity) throw new ApiError(404, 'no entity found');
```
ryanatkn, line 130: Technically this could be loaded only if there's `data`, but that's the hot path, and we can reuse the `entity` for `path` updates below as well to simplify some code in #865

src/lib/vocab/entity/entityServices.ts
```diff
@@ -124,16 +126,12 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
...
-					const newOrderedItems =
-						data.orderedItems && Array.from(new Set(data.orderedItems)).sort((a, b) => a - b);
-					const entity = await repos.entity.findById(entity_id);
-					const oldOrderedItems =
-						entity?.data.orderedItems &&
-						Array.from(new Set(entity.data.orderedItems)).sort((a, b) => a - b);
-					if (newOrderedItems?.toString() !== oldOrderedItems?.toString())
-						throw new ApiError(400, 'cannot update entity with orderedItems directly');
+				if (data && !dequal(new Set(data.orderedItems), new Set(entity.data.orderedItems))) {
+					throw new ApiError(400, 'cannot update entity orderedItems directly');
```
ryanatkn, line 134: dropped the "with" from the error message to be more precise, you can't update entities with orderedItems or add orderedItems to entities without orderedItems

src/lib/vocab/entity/entityData.ts
```diff
@@ -14,6 +14,10 @@ export interface BaseEntityData {
 	name?: string;
 	checked?: boolean;
 	orderedItems?: EntityId[];
+	directory?: true;
+	formerType?: string;
+	deleted?: string;
+	previousType?: string;
```
ryanatkn, line 20: Adding these like we did the others allows us to avoid the `'foo' in data` checks that make things more verbose. I'll do another pass on these types during alpha to make the ergonomics better.

src/lib/vocab/entity/entityServices.ts
```diff
@@ -124,16 +126,20 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
...
 				//TODO revist this, should data drive ties or vice versa?
-				if (data && 'orderedItems' in data) {
-					const newOrderedItems =
-						data.orderedItems && Array.from(new Set(data.orderedItems)).sort((a, b) => a - b);
-					const entity = await repos.entity.findById(entity_id);
-					const oldOrderedItems =
-						entity?.data.orderedItems &&
-						Array.from(new Set(entity.data.orderedItems)).sort((a, b) => a - b);
-					if (newOrderedItems?.toString() !== oldOrderedItems?.toString())
-						throw new ApiError(400, 'cannot update entity with orderedItems directly');
```
ryanatkn, line 136: dropped the "with" from the error message to be more precise, you can't update entities with orderedItems or add orderedItems to entities without orderedItems

src/lib/vocab/entity/entityServices.ts
```diff
@@ -124,16 +126,20 @@ export const UpdateEntitiesService: ServiceByName['UpdateEntities'] = {
 			params.entities.map(async (doc) => {
 				const {entity_id, data} = doc;
 
+				const entity = await repos.entity.findById(entity_id);
+				if (!entity) throw new ApiError(404, 'no entity found');
+
 				//TODO revist this, should data drive ties or vice versa?
-				if (data && 'orderedItems' in data) {
```
ryanatkn, line 128: this was the bug

greatbacon (approved):

PR #879, opened 2023-07-22
publish docs component and change markdown docs to svelte

ryanatkn:
Migrates our markdown docs to Svelte. The downside is that the code is little harder to read and write (VSCode's builtin HTML Emmet snippets are helpful, like `a+tab`), and the UX upside is that our docs are now in one place, we have full control over their presentation to the user, and we have no restrictions on integration with the larger system, so we can use components, link to resources on the website without it being a context switch, get contextmenu support, etc.

One simple first example is vocab link components - they're both links and have contextmenu behavior. A caveat is that the static docs won't have a working server API on the same host, but it will be a good example integrating of an external client when we do that work during alpha - we can just point it to `hub.felt.dev` and if it's down, at least the docs will still work, just not the live API interactions.

We can now republish our docs to `docs.felt.dev` as the canonical source to mirror `hub.felt.dev/docs` and never worry about ops. Once merged I'll import felt-server into that repo: <https://github.com/feltjs/docs>

- Moves the contents of the `/docs/+page.svelte` into `$lib/docs/Docs.svelte` and publishes it.
- Changes `src/docs` markdown files into svelte (eventually we'll move to a Markdown-like lang but maybe not until beta), and moves them to `$lib/docs/`. These are imported by `$lib/docs/Docs.svelte` and map to individual routes at two web locations, one statically hosted by GitHub, `hub.felt.dev/docs/a` and `docs.felt.dev/a`
- Adds `Mermaid.svelte` for the same `mermaid` diagrams as GitHub, adds the new dev dependency, and in the future we could style them with our vars - https://mermaid.js.org/

followup

- syntax highlighting from felt-ui (its next version will publish the needed prism.css)
- think about collapsed state of e.g. headings that gets persisted to localStorage -- also "focus" mode, search
- create felt-docs and import `@feltjs/felt-server/docs/Docs.svelte` into the root page


README.md
```diff
@@ -23,15 +23,15 @@ felt-server is a [Node.js](https://nodejs.org/) server that you can use in many
...
-[here on GitHub](https://github.com/feltjs/felt-server/blob/main/src/docs),
+[here on GitHub](/src/lib/docs),
 including our
-[known issues](https://github.com/feltjs/felt-server/blob/main/src/docs/known-issues.md).
+[known issues](https://github.com/feltjs/felt-server/blob/main/src/lib/docs/known-issues.md).
 
-To deploy to production, see [Deploying production](/src/docs/deploying-production.md)
-and [Managing production](/src/docs/managing-production.md).
+To deploy to production, see [Deploying production](https://docs.felt.dev/#deploying-production)
+and [Managing production](https://docs.felt.dev/#managing-production).
```
ryanatkn, line 34: I'll get these links working soon after merging this

src/lib/docs/Mermaid.svelte
```diff
@@ -0,0 +1,17 @@
...
+</script>
+
+<script lang="ts">
+	import {browser} from '$app/environment';
+
+	export let text: string;
+
+	let el: HTMLElement;
+
+	$: browser && text && el && void mermaid.run({nodes: [el]});
```
greatbacon, line 14: This is such a cool component!!!!!!!!!!!!!!!!

greatbacon (approved):

PR #881, opened 2023-07-25
OrderedCollections for lists

greatbacon:
This PR ports OrderedCollection behavior from the proof-of-concept Todo into Lists & refactors 3 common functions into the EntityHelper, plus fixes a bug in Home.


src/lib/plugins/home/Home.svelte
```diff
@@ -45,7 +45,7 @@
 		});
 	};
 
-	$: if ($entities && !$rules && !$norms) {
+	$: if ($query?.status === 'success' && !$rules && !$norms) {
```
greatbacon, line 48: This isn't part of the orderedcollection work it just fixes a bug in Home views that was re-creating rules & norms everytime the space was accessed, which fixes one of the freshness bugs.

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -75,3 +77,79 @@ export const lookupOrderedItems = (entity: Entity, ui: Ui): Array<Readable<Entit
...
+ * @param item
+ * @param parent
+ * @param actor_id
+ * @param actions
+ */
+export const moveUp = async (
+	item: Readable<Entity>,
+	parent: Readable<Entity>,
+	actor_id: ActorId,
+	actions: Actions,
```
greatbacon, line 92: I wasn't quite sure if this was a pattern we wanted to start setting, since we don't do it anywhere else in the system atm. Being able to extract functions with actions though does help cleanup a lot of View code though!

src/lib/plugins/todo/TodoItem.svelte
```diff
@@ -38,21 +38,15 @@
 	// TODO extract this pattern from 2 places, into the query system?
```
greatbacon, line 38: I ran out of ideas on how to further refactor this chunk of code (from `let orderedEntities...` to the `assignOrderEntities` function). At the moment it's going to get repeated anywhere we need to deal with orderedEntities but it may be small enough that's okay?

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -75,3 +77,79 @@ export const lookupOrderedItems = (entity: Entity, ui: Ui): Array<Readable<Entit
...
+ * @param actor_id
+ * @returns an array of entity stores referenced by the parentList's orderedItems attribute, or null
+ */
+export const loadOrderedEntities = async (
+	list: Entity,
+	orderedItems: EntityId[],
+	actor_id: ActorId,
+	ui: Ui,
+	actions: Actions,
+): Promise<Array<Readable<Entity>> | null> => {
```
ryanatkn, line 144: I've been thinking about this, let's talk about this maybe needing to be nullable everywhere, `Array<Readable<Entity> | null>`

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -75,3 +77,79 @@ export const lookupOrderedItems = (entity: Entity, ui: Ui): Array<Readable<Entit
...
+	let entityIdsToLoad: EntityId[] | null = null;
```
ryanatkn, line 145: let's leave a TODO here to refactor this, , like `// TODO use query helper`

ryanatkn, line 145:
```suggestion
	let entityIdsToLoad: EntityId[] | null = null; // TODO use query helper
```

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -75,3 +77,79 @@ export const lookupOrderedItems = (entity: Entity, ui: Ui): Array<Readable<Entit
...
+/**
+ * Loads entities in the orderedItems of the parentList
+ * @param parentList
+ * @param orderedItems
+ * @param actor_id
+ * @returns an array of entity stores referenced by the parentList's orderedItems attribute, or null
+ */
+export const loadOrderedEntities = async (
+	list: Entity,
+	orderedItems: EntityId[],
```
ryanatkn, line 140: shouldn't need `orderedItems` here?

src/lib/vocab/entity/entityHelpers.ts
ryanatkn: cool yeah this is a good step, I'll refactor with the query system next, anything passing through `actions` and probably `actor_id` like this can be cleaned up

src/lib/plugins/lists/ListItem.svelte
```diff
@@ -74,6 +62,25 @@
 	const toggleExpandItems = () => {
 		expandItems = !expandItems;
 	};
+
+	// TODO extract this pattern from 2 places, into the query system?
```
ryanatkn, line 66: yeah good comment, it'll be straightforward to refactor from this

src/lib/plugins/lists/ListItem.svelte
```diff
@@ -74,6 +62,25 @@
...
+			$entity!,
+			orderedItems!,
+			$actor.actor_id,
+			ui,
+			actions,
+		);
+	};
+
+	$: first = $parentList.data.orderedItems![0] === $entity.entity_id;
+	$: last = $parentList.data.orderedItems!.at(-1) === $entity.entity_id;
```
ryanatkn, line 81: let's not use `.at` because of browser compatibility, I see we also do it in `TodoItem.svelte`, the other places aren't user facing

becomes the equivalent of `$parentList.data.orderedItems[$parentList.data.orderedItems.length - 1] === $entity.entity_id`

ryanatkn (approved):

src/lib/plugins/todo/TodoItem.svelte
ryanatkn: yep we'll get to it soon, the refactoring will be easier in context with the query changes

PR #882, opened 2023-07-26
confirm before reset of changed property

ryanatkn:
Changes the `reset` button on the property editor to warn you before destructively resetting your changes. Same behavior as GitHub but using our `ConfirmDialog` instead of the browser's `confirm`.


greatbacon (approved):
🧹

PR #883, opened 2023-07-26
replace `$$restProps` with explicit attrs

ryanatkn:
same as https://github.com/feltjs/felt-ui/pull/330


src/lib/plugins/list/ListInput.svelte
```diff
@@ -37,9 +37,9 @@
 </script>
 
 <TextInput
-	style="height: var(--input_height)"
-	{actor}
 	placeholder=">"
+	attrs={{style: 'height: var(--input_height)'}}
```
ryanatkn, line 41: we may as a pattern want to accept `style` as a prop

src/lib/ui/SignOutForm.svelte
```diff
@@ -31,7 +33,7 @@
 	};
 </script>
 
-<form {...$$restProps}>
+<form {...formAttrs}>
```
ryanatkn, line 36: thinking I may want to rename `formAttrs` to `attrs` everywhere, with the rule of thumb to use the plain `attrs` if it's obvious

greatbacon (approved):

src/lib/ui/SignOutForm.svelte
ryanatkn: went ahead and did this

PR #887, opened 2023-07-27
add happy path to edit entity content

ryanatkn:
much nicer `Edit Entity` experience

need to improve it for some cases where other fields are used besides `content`, but this is a good start, we could do those as we do views


src/lib/ui/EntityContextmenu.svelte
```diff
@@ -29,13 +30,27 @@
...
+			run={() =>
+				actions.OpenDialog(
+					toDialogParams(
+						EntityEditor,
+						{actor, entity, done: () => actions.CloseDialog()},
+						{layout: 'page'},
+					),
+				)}
+		>
+			More details
```
ryanatkn, line 52: in the docs our contextmenu for vocab says the equivalent of `More about <code>Entity</code>` - maybe I should change this to that?

src/lib/ui/EditEntityContentForm.svelte
```diff
@@ -0,0 +1,73 @@
+<script lang="ts">
```
ryanatkn, line 1: this is a stripped down version of `EditEntityForm`

src/lib/ui/EntityContextmenu.svelte
greatbacon: I think the nesting nature here creates a slightly different presentation context & trying to make the one fit the other could get weird.

greatbacon (approved):

PR #895, opened 2023-08-03
add vocab name schema

ryanatkn:
Adds generated `VocabName` and `SchemaName` enums and generated types, clarifying our source of truth for schemas and vocab data.

Doing this work made me realize "vocab names" include actions like `CreateSpace` and views like `Chat`, but "schema names" in contrast would be `CreateSpaceParams` and `CreateSpaceResponse`. It's possible we'll want to make `CreateSpace` an actual schema with the params and response as static properties, but I'll wait to see if that's the right call, it seems weird to make a schema just for static data.

`VocabSchema` in Gro is now a bit of a misnomer -- I may just change it to `JsonSchema`.

Also updates our raw schema url to use GitHub. I'm thinking now this should be hosted on felt.dev (the website repo with the docs) instead. (still avoiding hub.felt.dev to use GitHub's CDN) To do this, I'll do a followup to export `vocab.json`.

followup

- data.ts todos
- figure out vocab/schemas naming problem where vocab.json has non-vocab schemas


src/lib/vocab/vocab.gen.json.ts
```diff
@@ -12,7 +12,7 @@ export const gen: Gen = async () => {
 	// TODO make these options configurable
 	const schema = bundleSchemas(
 		schemas,
-		'https://hub.felt.dev/schemas/vocab.json',
+		'https://github.com/feltjs/felt-server/blob/main/src/static/schemas/vocab.json',
```
ryanatkn, line 15: this seems nice and stable

greatbacon (approved):

src/lib/vocab/vocab.gen.json.ts
ryanatkn: (I say the day before we decide on a rename)

ryanatkn:
looks like this is breaking the server during dev with a circular dep initialization-order bug

src/lib/vocab/data.ts
```diff
@@ -0,0 +1,47 @@
+import {actionDatas} from '$lib/vocab/action/actionData';
+import {modelSchemas} from '$lib/vocab/schemas';
+import {viewTemplates} from '$lib/vocab/view/view';
+import {toSchemaName} from '$lib/util/schema';
+import type {ClientActionData, ServiceActionData} from '$lib/vocab/action/action';
+import type {VocabName} from '$lib/vocab/vocab';
+
+export const sortedModelSchemas = modelSchemas.slice().sort((a, b) => a.$id.localeCompare(b.$id));
```
ryanatkn, line 8: realizing this code is a little funky, looks like it belongs `$lib/vocab/schemas` and some of those other modules -- it's also causing circular dep errors in two ways, I'm going to go back to draft and see if I can restructure things

PR #900, opened 2023-08-07
improve docs rendering

ryanatkn:
Improves how some docs are displayed.

Also renames `CommunityActor` to `HubActor` which causes some diff noise


src/lib/ui/SchemaInfo.svelte
```diff
@@ -56,6 +59,13 @@
 				</div>
 			{/each}
 		{/if}
+		{#if anyOf}
+			<div class="union">
+				{#each anyOf as unionSchema}
+					<svelte:self schema={unionSchema} />
+				{/each}
+			</div>
+		{/if}
```
ryanatkn, line 68: this isn't perfect but it's better than it was

greatbacon: what a poetic comment

greatbacon (approved):

PR #901, opened 2023-08-07
rename `felt-server` to `felt`

greatbacon:
I didn't touch anything related to the npm package since that seemed like a heavier/higher stakes lift & I wasn't quite sure what the process for that was.


ryanatkn (approved):
# 💚 

running `gro format` will probably fix that

PR #902, opened 2023-08-08
fix docs links

ryanatkn:
fixes some docs.felt.dev links to felt.dev/docs

also upgrades gro - this may improve perf on macbooks, filesystem watching in gro now uses chokidar (which was already a dep through svelte-check) instead of cheap-watch


greatbacon (approved):
🧹

PR #904, opened 2023-08-11
preserve markup characters

ryanatkn:
Changes the rendering of text like `*foo*` from `<strong>foo</strong>` to `<strong>*foo*</strong>`. This is like Discord and unlike most others. edit: I'm mistaken, this is only in the WYSIWYG editor not rendered, I'm closing, LMK if you actually like it

Pros

- non-destructive to user-input, so text can be copy-pasted through Felt without these characters getting stripped -- Felt rendering still modifies input text like HTML/component tags, but I think it's best to do the least? For example the `Mention` and `Link` components preserve input text.
- subjectively looks better?

Cons

- can be surprising
- subjectively looks worse?


greatbacon:
Okay, I was just playing around with the web UI for Discord & was getting confused lol

ryanatkn:
yeah sorry about that, I don't use discord much these days and this was a wrong memory

PR #905, opened 2023-08-11
replace ties with a tree

ryanatkn:
rebooting #878 but keeping directories

rewriting the data system, opting for a collection tree instead of ties


src/lib/db/migrations/00053-delete-ties.ts
```diff
@@ -0,0 +1,20 @@
...
+	await sql`
+		DROP TABLE ties;
+	`;
+
+	// `lists` is not nullable but `items` is
+	await sql`
+		ALTER TABLE entities
+		ADD COLUMN lists int[] NOT NULL DEFAULT '{}'
+	`;
+	await sql`CREATE INDEX ON entities USING gin (lists)`;
```
ryanatkn, line 13: This may be where we want to call it for the existing db and data -- writing the migration with ties here isn't that hard, but it's not easy either, and it seems like wasted effort at this point.

src/lib/db/migrations/00053-delete-ties.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const up = async (sql: Sql<any>): Promise<void> => {
+	await sql`
+		DROP TABLE ties;
+	`;
+
+	// `lists` is not nullable but `items` is
+	await sql`
+		ALTER TABLE entities
+		ADD COLUMN lists int[] NOT NULL DEFAULT '{}'
```
ryanatkn, line 11: This may be where we want to call it for the existing db and data -- writing the migration with ties here isn't that hard, but it's not easy either, and it seems like wasted effort at this point. This doesn't fail with the existing data, but entities will be orphaned.

src/lib/db/seed.ts
```diff
@@ -274,7 +271,7 @@ const SEED_BY_VIEW_NAME: Record<string, (ctx: SeedContext) => Promise<void>> = {
 	ReplyChat: async (ctx) => {
 		const e1 = await generateEntity(ctx, 'Those who know do not speak.');
 		await generateEntity(ctx, 'Those who speak do not know.', undefined, undefined, [
-			{source_id: e1.entities[0].entity_id, type: 'HasReply'},
+			e1.entities[0].entity_id,
```
ryanatkn, line 274: these are nice simplifications

src/routes/+layout.svelte
```diff
@@ -1,7 +1,9 @@
 <script lang="ts" context="module">
 	if (import.meta.env.PROD) {
 		if (!browser) {
-			Logger.prefixes.unshift(() => format(new Date(), 'M/d H:mm:ss.SSS'));
+			// TODO BLOCK I think this is right?
+			Logger.level = 'off';
+			// Logger.prefixes.unshift(() => format(new Date(), 'M/d H:mm:ss.SSS'));
```
ryanatkn, line 6: I know this could lead to issues making debugging harder, but I think the SSR logging is generally not what we intend for the logger, it can be really noisy with objects and such that are intended for the browser console, and I don't want to be careful about it. Maybe we just up the level to error?

src/routes/+layout.svelte
```diff
@@ -1,8 +1,6 @@
 <script lang="ts" context="module">
-	if (import.meta.env.PROD) {
-		if (!browser) {
-			Logger.prefixes.unshift(() => format(new Date(), 'M/d H:mm:ss.SSS'));
-		}
+	if (import.meta.env.PROD && !browser) {
+		Logger.level = 'off';
```
ryanatkn, line 3: I know this could lead to issues making debugging harder, but I think the SSR logging is generally not what we intend for the logger, it can be really noisy with objects and such that are intended for the browser console, and I don't want to be careful about logging personal data from the frontend - it's a privacy problem the way we've been using it.

PR #908, opened 2023-08-14
Add simple password strength checker to account create/pw update

greatbacon:
Added some logic to check password strength on account creation or pw update. It's pretty simple, just checks length for now.

I could also expand it in this PR based off Bitwarden's practices ( at least 1 uppercase, 1 lowercase, 1 number, 1 of `!@#$%^&*`)

I just put the check at the service layer since we need the pre-hash value for comparison & we already have some validation logic at that level, but I could move it into the worker if you think it makes more sense there.


src/lib/db/seed.ts
```diff
@@ -143,6 +144,15 @@ export const seed = async (db: Database, much = false): Promise<void> => {
 		await createDefaultEntities(toMainAccountServiceRequest, spaces, nextActor, repos);
 	}
 
+	const adminHub = await repos.hub.loadAdminHub();
+	await UpdateHubService.perform({
+		...toMainAccountServiceRequest(),
+		params: {
+			actor: mainActorCreator.actor_id,
+			hub_id: ADMIN_HUB_ID,
+			settings: {...adminHub!.settings, instance: {minimumPasswordLength: 3}},
```
ryanatkn, line 153: The passwords we seed are 1 character long, is there a reason to make this anything else?

```suggestion
			settings: {...adminHub!.settings, instance: {minimumPasswordLength: 1}},
```

src/lib/db/seed.ts
```diff
@@ -143,6 +144,15 @@ export const seed = async (db: Database, much = false): Promise<void> => {
...
+		},
+	});
```
ryanatkn, line 155: Two things here - it was clobbering any existing `settings.instance` value, and the passwords we seed are 1 character, I don't see a reason to restrict this in dev.

```suggestion
	const adminHub = await repos.hub.loadAdminHub()!;
	await UpdateHubService.perform({
		...toMainAccountServiceRequest(),
		params: {
			actor: mainActorCreator.actor_id,
			hub_id: ADMIN_HUB_ID,
			settings: {
				...adminHub.settings,
				instance: {...adminHub.settings.instance, minimumPasswordLength: 1},
			},
		},
	});
```

ryanatkn (approved):

PR #910, opened 2023-08-21
update restrictions for actor and hub names

ryanatkn:
Changes our naming restrictions of actors and hubs to match Mastodon. This will ensure future compatibility, so e.g. Mastodon won't reject our longer usernames or ones that have hyphens - we may not have all of the details encoded (we didn't look at Mastodon's source) but the intent is now clear.


src/lib/util/parseSvast.test.ts
```diff
@@ -102,21 +102,21 @@ test__parseSvast('parses absolute links', async () => {
...
+				children: [{type: 'text', value: '/felt/some_hub'}],
 			},
 		],
 	});
 });
 
 test__parseSvast('parses relative links', async () => {
 	const parsed = parseSvast({
-		value: './some-path/some-link',
+		value: './some_hub/some_path', // TODO shouldn't fail with `some-path`, but the parser needs to be rewritten anyway
```
ryanatkn, line 119: just a note, item paths should use underscores for now too -- eventually we will probably want to support anything, but for now this keeps us from mucking around in the parser that will be rewritten anyway

greatbacon (approved):

PR #914, opened 2023-08-23
fix deleteHub issues

greatbacon:
This PR fixes a couple of issues around hub deletion

One is an ordering issue with the DB cascades; fixed by simply deleting the spaces of a hub prior to deleting the hub. Without doing this we end up in a state where the DB tries to remove hub actors first and then fails a constraint because there are still entities in the space (directories mostly) still referencing the actors.

The second removes an assertion in the RoleList component that is no longer true when Roles get evicted in the client after a Hub deletion but before the client navigates away from the View.


src/lib/vocab/space/SpaceRepo.ts
```diff
@@ -108,4 +108,11 @@ export class SpaceRepo extends PostgresRepo {
 		if (!data.length) throw Error('no space found');
 		return data;
 	}
+
+	async deleteByHub(hub_id: HubId): Promise<Space[]> {
+		const data = await this.sql<Space[]>`
+			DELETE FROM spaces WHERE hub_id=${hub_id}
+		`;
+		return data;
```
ryanatkn, line 116: The type here is off, I looked at similar and we're returning `number` and `data.count` for this kind of call.

ryanatkn, line 116: The type here is off, I looked at similar and we're returning `number` and `data.count` for this kind of call.

```suggestion
	async deleteByHub(hub_id: HubId): Promise<number> {
		const data = await this.sql<any[]>`
			DELETE FROM spaces WHERE hub_id=${hub_id}
		`;
		return data.count;
```

ryanatkn (approved):
nice that's a tricky one

src/lib/vocab/space/SpaceRepo.ts
greatbacon: oh yeah, that's a copy pasta error 🙃

PR #916, opened 2023-08-24
Fix for leaving hub & orphaning it

greatbacon:
This is a similar fix as the DeleteHub issue, but this time for when you leave a hub & orphan it.


ryanatkn (approved):
🐛

greatbacon:
> # 🐛
> Since we have this in two places now, what do you think of adding `deleteHub` in `$lib/vocab/hub/hubHelpers.server.ts`?

Good callout!

PR #918, opened 2023-08-25
fix broadcast for removeActor

ryanatkn:


src/lib/server/service.ts
```diff
@@ -112,6 +113,7 @@ export interface NonAuthenticatedServiceRequest<TParams = any> {
 	session: ISessionApi;
 	broadcast: IBroadcastApi;
 	passwordHasher: PasswordHasher;
+	postRequestHooks: PostRequestHook[];
```
ryanatkn, line 118: The API I outlined looks a little strange internally, but it's because we don't want to expose an array like this as API. I think the middleware implementation was what we wanted, and this should be `postRequest: (cb: () => void) => void;

greatbacon: yeah I did the array cause I didn't understand what type needed to be placed on the service interface to take in that callback function. What should I do?

ryanatkn: I think I like this:

```
afterRequest: (cb: () => void | Promise<void>) => void;
```

And then `await cb()` sequentially when flushing them, allowing them to optionally return a promise which will be important for the future for error handling and possible queue processing. (versus fire-and-forget, which would have no failure recourse)

`afterRequestHooks` or `afterMutationCallbacks` makes sense for the local array

greatbacon: Okay, I shifted to `afterRequestCallbacks` for naming stuff & I think I've got the typing in line with what you were thinking. I'm still running into a type issue in the testHelper though, and I'm stuck again. I know callbacks & I know types, but I'm still week on mixing the two. 😅

ryanatkn: I think I got it in #924

greatbacon (approved):
I'm not sure if I need to be the one to approve this or you do lol, but I think we're mostly wrapped here?

ryanatkn:
looks great! I think you should do the honors

PR #919, opened 2023-08-28
upgrade to node v20

ryanatkn:
Upgrades to Node v20, and adds Node upgrade instructions to the docs.

For the time being, install gro with `npm i -g @feltjs/gro@0.80`


greatbacon (approved):
love the instructions; makes it ez pz 😄

PR #920, opened 2023-08-28
add check policy helper

ryanatkn:
Adds two new functions to the service request API, `checkPolicy` and `checkHubAccess`. It cleans up a little boilerplate, but more importantly, it bakes the `actor` prop into the API so you can't accidentally pass the wrong one, which would be a security problem.

You can still call the existing helpers if a custom `actor` is `repos` is needed, but we don't anywhere. The existing module-level helpers are renamed with the suffix `ForActor`.


src/lib/server/service.ts
```diff
@@ -162,7 +166,19 @@ export function toServiceRequest<TParams = any>(
...
+				null,
+				repos,
+				actor.actor_id,
+			);
+			(req as AuthorizedServiceRequest).checkHubAccess = checkHubAccessForActor.bind(
+				null,
+				repos,
+				actor.actor_id,
+			);
+		}
```
ryanatkn, line 181: Instead of this, I was thinking about something like this:

```ts
(req as AuthorizedServiceRequest).check = new CheckApi(repos, actor);
```

Usage in the services:

```ts
// from
checkPolicy('foo', hub_id);
// to
check.policy('foo', hub_id);
```

It'd be a simpler class wrapper around the params that could contain any other pre-baked checks we want to add in the future.

src/check.task.ts
```diff
@@ -5,23 +5,20 @@ import {MIGRATIONS_DIR, toMigrationIndex} from '$lib/db/migration';
...
+		lint: z.boolean({description: ''}).default(true),
+		'no-lint': z.boolean({description: 'opt out of linting'}).default(false),
 		// added by this project
-		migrations: z.boolean({description: ''}).optional().default(true),
-		'no-migrations': z
-			.boolean({description: 'opt out of migrations check'})
-			.optional() // TODO behavior differs now with zod, because of `default` this does nothing (above too)
-			.default(false),
+		migrations: z.boolean({description: ''}).default(true),
+		'no-migrations': z.boolean({description: 'opt out of migrations check'}).default(false),
```
ryanatkn, line 21: I'm just deleting the `.optional()` part and TODO here and in other tasks. We don't need to distinguish between values with and without defaults, so this doesn't cause any issues.

src/lib/vocab/assignment/assignmentServices.test.ts
```diff
@@ -82,17 +82,17 @@ test__assignmentServices('fail to delete a personal assignment', async ({repos,
...
 	const assignment = assignments.find(
 		(a) => a.actor_id === communityActor.actor_id && a.hub_id === hub.hub_id,
 	)!;
 	unwrapError(
 		await DeleteAssignmentService.perform({
-			...toServiceRequestMock(repos, communityActor),
+			...toServiceRequestMock(repos, actor),
 			params: {
-				actor: actor.actor_id,
+				actor: communityActor.actor_id,
```
ryanatkn, line 95: this was a bug in the tests

src/lib/server/servicesIntegration.test.ts
```diff
@@ -145,7 +145,7 @@ test_servicesIntegration('services integration test', async ({repos, random}) =>
 	assert.is((await repos.assignment.filterByHub(hub.hub_id)).length, 3);
 	unwrap(
 		await DeleteAssignmentService.perform({
-			...toServiceRequestMock(repos, actor2),
+			...toServiceRequestMock(repos, actor1),
```
ryanatkn, line 148: tests were failing for these, I think these are bugs?

src/check.task.ts
```diff
@@ -5,23 +5,20 @@ import {MIGRATIONS_DIR, toMigrationIndex} from '$lib/db/migration';
...
-		'no-typecheck': z.boolean({description: 'opt out of typechecking'}).optional().default(false),
-		test: z.boolean({description: ''}).optional().default(true),
-		'no-test': z.boolean({description: 'opt out of running tests'}).optional().default(false),
-		gen: z.boolean({description: ''}).optional().default(true),
-		'no-gen': z.boolean({description: 'opt out of gen check'}).optional().default(false),
-		format: z.boolean({description: ''}).optional().default(true),
-		'no-format': z.boolean({description: 'opt out of format check'}).optional().default(false),
-		lint: z.boolean({description: ''}).optional().default(true),
-		'no-lint': z.boolean({description: 'opt out of linting'}).optional().default(false),
+		// base args -- TODO extend generically, `import {Args as BaseArgs} from '@feltjs/gro/dist/check.task.js';`
```
ryanatkn, line 8: the next version of Gro exports `Args` so we can do this generically

src/lib/server/servicesIntegration.test.ts
greatbacon: Yeah I think you're right, those look like copypasta errors & actor1 should be the actor declared on the serviceRequest.

src/lib/server/service.ts
greatbacon: Maybe leave a TODO in this block of code suggesting a potential future wrapper class? I think until we have a solid third use case to plug in there it's not quite worth adding that extra layer of abstraction yet.

greatbacon (approved):
This PR is smooth like butter

PR #921, opened 2023-08-28
add `directory_id` and `hub_id` to entities

ryanatkn:


src/lib/vocab/entity/entityHelpers.server.ts
```diff
@@ -47,9 +57,7 @@ export const updateDirectories = async (
 	const updatedDirectories = await Promise.all(
 		directoryIds.map(async (directoryId) => repos.entity.update(directoryId) as Promise<Directory>),
 	);
-	const hubIds = (await repos.hub.filterBySpaces(updatedDirectories.map((d) => d.space_id))).map(
-		(h) => h.hub_id,
-	);
+	const hubIds = updatedDirectories.map((d) => d.hub_id);
```
ryanatkn, line 60: this one felt good -- I looked around a bit for simplifications and found a handful, there's definitely more to find

src/lib/vocab/account/AccountRepo.test.ts
```diff
@@ -0,0 +1,19 @@
...
+test__AccountRepo.after(teardownDb);
+
+test__AccountRepo('loadClientSession', async ({repos, random}) => {
+	const {account, space} = await random.space();
+	const {spaces} = await repos.account.loadClientSession(account.account_id);
+	assert.ok(spaces.find((s) => s.space_id === space.space_id));
+});
+
+test__AccountRepo.run();
+/* test__AccountRepo */
```
ryanatkn, line 20: This is more of a smoke test since `loadClientSession` calls so many others, including `filterDirectoriesByAccount` which I modify in this PR with the `hub_id` simplification.

src/lib/vocab/entity/EntityRepo.ts
```diff
@@ -16,23 +16,29 @@ const log = new Logger(gray('[') + blue('EntityRepo') + gray(']'));
...
+		space_id: SpaceId,
+		hub_id: HubId,
 		data: EntityData,
-		space_id: SpaceId | null,
+		directory_id: EntityId | null = null,
 		path: string | null = null,
 	): Promise<Entity> {
+		if (directory_id === null && !data.directory) {
+			throw Error('missing directory_id or data.directory');
+		}
```
ryanatkn, line 27: I was going to write a test for this constraint, but I'm wondering if we want to remove `data.directory` now, and check `e.entity_id === e.directory_id`, which could be a simple helper `isDirectory(entity)`

src/lib/db/migrations/00052-add-entity-directory-id.ts
```diff
@@ -0,0 +1,52 @@
...
+		SET hub_id=spaces.hub_id
+		FROM spaces
+		WHERE entities.space_id=spaces.space_id;
+		`;
+	// after migrating hub_id, set contraints
+	await sql`
+		ALTER TABLE entities
+		ALTER COLUMN hub_id
+		SET NOT NULL
+	`;
```
ryanatkn, line 42: Like we discussed, `entities.directory_id` and `entities.hub_id` have the `NOT NULL` constraint, but `spaces.directory_id` does not.

src/lib/db/migrations/00052-add-entity-directory-id.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	// index some columns
+	await sql`CREATE INDEX ON spaces (directory_id);`;
+	await sql`CREATE INDEX ON entities (directory_id);`;
+	await sql`CREATE INDEX ON entities (hub_id);`;
+	await sql`CREATE INDEX ON entities (space_id);`;
+	await sql`CREATE INDEX ON entities (actor_id);`;
+
+	await sql`ALTER SEQUENCE files_file_id_seq RENAME TO items_item_id_seq`;
```
ryanatkn, line 51: I had to use this sequence to initialize the `directory_id`, so I went ahead and renamed it to the final form.

src/lib/ui/uiMutationHelpers.ts
```diff
@@ -69,10 +69,6 @@ export const updateLastSeen = (ui: WritableUi, directory_id: EntityId, time = Da
 	const directory = entityById.get(directory_id) as Readable<Directory> | undefined;
 
 	if (directory) {
-		upsertFreshnessByHubId(
-			ui,
-			//TODO add directory field to space & vice versa to avoid this mess below
-			spaceById.get(directory.get().space_id)!.get().hub_id,
-		);
+		upsertFreshnessByHubId(ui, directory.get().hub_id);
```
ryanatkn, line 72: :D

src/lib/vocab/entity/EntityRepo.ts
greatbacon: I think if we can move our directory concern out of the data, that makes things cleaner from an ActivityStreams perspective but yeah, I'm not sure if it makes for a better consumer POV.

I guess from a clarity perspective I think it would be best to really only have one mechanism for determining if an entity is a directory, just to keep things clean and consistant, so I think I'm leaning slightly towards removing that this field from data.

greatbacon (approved):
Looks good, some nice query cleanup in here!

Approving pending the final resolution of the `entity.data.directory` question

PR #922, opened 2023-08-29
upgrade deps

ryanatkn:
Upgrades to the latest deps with new orgs.

Changes environment variables to have the same shape in dev and prod, so dev environments need to update by adding empty vars - see the `.env.(development|production).default` file changes. There's some renames as well. If you don't need any custom values in your dev repo, simply deleting the three env files at the root is enough to migrate.

The server build plugin in Gro now outputs all `entry_points` as individually bundled scripts. This means the `server/server.ts` is now ~5300 lines long and `infra/after_deploy.ts` script (previously `infra/deployed.ts`) is 1000+. It bundles our code, not things in `node_modules`. This is a simplification that makes things robust, no longer do our runtime artifacts depend on the code as a library at runtime, with potentially different needs e.g. dev vs prod. Migrations are similarly independently bundled scripts, though most remain small because they have few dependencies on our modules.

It now outputs the server to `dist_server` in the root, and we gitignore the new Gro pattern of `dist_*` directories in the root. (and `gro clean` now includes them if `build===true`)

Changes all code to use `import.meta.env.DEV` and similar - `dev` from `$app/environment` is not what I thought. (it's not guaranteed to match, the SvelteKit docs say, it indicates if the dev server is running)

future

- replace usage of `crypto` with `webcrypto`, importing from `node:crypto` (and adding `node:` prefixes elsewhere)


src/lib/infra/deploy.task.ts
```diff
@@ -26,7 +24,7 @@ export const task: Task = {
 			await spawn('tar', [
 				'-cf',
 				`${artifactFilename}`,
-				DIST_DIRNAME,
+				BUILD_DIST_DIRNAME,
```
ryanatkn, line 27: renamed this because `dist/` is now SvelteKit's, so Gro prefixes its own version with `BUILD` - I don't love it but I can't think of a better term

src/lib/infra/syncEnvGitHash.task.ts
```diff
@@ -1,33 +1,32 @@
...
-		const branch = (await fs.readFile('.git/HEAD', 'utf8')).trim().substring(5);
+	run: async ({log}) => {
+		const branch = readFileSync('.git/HEAD', 'utf8').trim().substring(5);
 
-		let gitHashContents = '';
+		let rawGitHash = '';
 		const buildResult = spawnProcess('git', ['show-ref', '-s', branch], {stdio: 'pipe'});
 		buildResult.child.stdout?.on('data', (data: Buffer) => {
-			gitHashContents += data.toString().trim();
+			rawGitHash += data.toString().trim();
```
ryanatkn, line 16: I may change this to use `execSync`, I think that's how we can make it a one-liner.

src/lib/plugins/forum/ForumItemDetail.svelte
```diff
@@ -26,12 +26,15 @@
...
-	}, [] as Array<Readable<Entity>>);
+	$: items = Array.from($destTies.value).reduce(
+		(acc, tie) => {
+			if (tie.type === 'HasItem') {
+				acc.push(entityById.get(tie.dest_id)!);
+			}
+			return acc;
+		},
+		[] as Array<Readable<Entity>>,
+	);
```
ryanatkn, line 37: prettier 3 reformatted some code like this block

src/lib/plugins/lists/ClearCheckedButton.svelte
```diff
@@ -19,13 +19,16 @@
...
+		Array.from($destTies.value).reduce(
+			(acc, tie) => {
+				if (tie.type === 'HasItem') {
+					const entity = entityById.get(tie.dest_id)!;
+					acc.push(entity);
+				}
+				return acc;
+			},
+			[] as Array<Readable<Entity>>,
+		);
```
ryanatkn, line 31: prettier again

src/lib/server/ApiServer.ts
```diff
@@ -91,8 +92,8 @@ export class ApiServer {
 		}
 
 		// SvelteKit Node adapter, adapted to our production API server
-		if (process.env.NODE_ENV === 'production') {
-			const importPath = '../../../svelte-kit/handler.js';
+		if (!dev) {
+			const importPath = '../../build/handler.js';
```
ryanatkn, line 96: need to think through how to package up our final artifacts now that `dist/` is SvelteKit's, Gro builds to `.gro/dist/server/server.js`, maybe we should just move/copy files there

src/lib/server/Broadcast.ts
```diff
@@ -1,5 +1,6 @@
 import type {WebSocket} from 'ws';
 import {Logger} from '@feltjs/util/log.js';
+import {dev} from '$app/environment';
```
ryanatkn, line 3: Apparently SvelteKit only says this is `true` when we're running the dev server, but Gro assigns it based on if you're running either `gro dev` or `gro build`. (called by deploy and publish) We may need to change to use `import.meta.env.DEV/PROD/MODE`, which Gro also shims.

src/lib/server/api.ts
```diff
@@ -3,7 +3,10 @@ import {ResultError, type Result} from '@feltjs/util/result.js';
 import type {ErrorResponse} from '$lib/util/error';
 
 export class ApiError extends Error {
-	constructor(public readonly status: number, message: string) {
+	constructor(
+		public readonly status: number,
+		message: string,
+	) {
```
ryanatkn, line 9: prettier 3

src/lib/server/env.ts
```diff
@@ -1,9 +1,6 @@
-import dotenv from 'dotenv';
-import {copyFileSync, existsSync} from 'fs';
-
-const dev = process.env.NODE_ENV !== 'production'; // TODO fixme in multiple places to use `$app/environment`
-
-// TODO how to configure this stuff in user projects? felt/gro config?
+import {dev} from '$app/environment';
+import {env as env_dynamic_public} from '$env/dynamic/public';
+import {existsSync, copyFileSync, readFileSync, writeFileSync} from 'node:fs';
```
ryanatkn, line 3: I might change these to async

src/lib/server/env.ts
```diff
@@ -15,62 +12,45 @@ const envs: Array<{file: string; defaultFile: string; load: boolean}> = [
...
-		throw Error(`Missing environment variable: ${key}`);
-	}
-	return value;
-};
-
-let initedEnv = false;
-
 export const initEnv = (): void => {
-	if (initedEnv) return;
-	initedEnv = true;
```
ryanatkn, line 52: better to just avoid caching

src/lib/server/env.ts
```diff
@@ -15,62 +12,45 @@ const envs: Array<{file: string; defaultFile: string; load: boolean}> = [
...
+export const updateEnvGitHash = (file: string, hash: string): boolean => {
+	const contents = readFileSync(file, 'utf8');
+	const updated = updateEnvValue(contents, 'PUBLIC_GIT_HASH', hash);
+
+	// changed?
+	if (updated === contents) return false;
+
+	writeFileSync(file, updated, 'utf8');
+
+	env_dynamic_public.PUBLIC_GIT_HASH = hash;
```
ryanatkn, line 39: we need to use the `dynamic` version of the env here, it's the only variable that has this requirement

src/lib/server/password.worker.ts
```diff
@@ -171,5 +171,5 @@ const passwordWorker = (): void => {
 };
 
 if (!isMainThread) {
-	passwordWorker();
+	main();
```
ryanatkn, line 174: clearer name to signal we're in a file that may be standalone in a process

src/lib/ui/ui.ts
```diff
@@ -6,13 +6,11 @@ import {
...
 	type Mutable,
 } from '@feltcoop/svelte-gettable-stores';
-import {setContext, getContext, type SvelteComponent, type ComponentType} from 'svelte';
+import {setContext, getContext, type SvelteComponent} from 'svelte';
 import type {DialogParams} from '@feltjs/felt-ui/dialog.js';
 import {browser} from '$app/environment';
 import {EventEmitter} from 'eventemitter3';
 import {createContextmenu, type ContextmenuStore} from '@feltjs/felt-ui/contextmenu.js';
-import type ContextmenuLinkEntry from '@feltjs/felt-ui/ContextmenuLinkEntry.svelte';
-import type ContextmenuTextEntry from '@feltjs/felt-ui/ContextmenuTextEntry.svelte';
```
ryanatkn, line 15: able to remove this because we can import Svelte files in tests now

src/lib/util/ajv.ts
```diff
@@ -1,5 +1,5 @@
 import Ajv, {type Options, type ErrorObject, type ValidateFunction, type SchemaObject} from 'ajv';
-import type {VocabSchema} from '@feltjs/gro';
+import type {JsonSchema} from '@feltjs/gro/util/schema.js';
```
ryanatkn, line 2: `Vocab` isn't a concept Gro should know about

src/lib/util/constants.ts
```diff
@@ -6,3 +6,5 @@ export const ADMIN_HUB_ID = 1;
 export const ADMIN_ACTOR_ID = 1;
 export const GHOST_ACTOR_NAME = 'ghost';
 export const GHOST_ACTOR_ID = 2;
+
+export const GUEST_ACTOR_NAME = 'guest';
```
ryanatkn, line 10: deleted the module where this was since it had only 1 export, and it was making Gro fail for some reason while I was developing, it then failed on other modules but it seemed good to move this here

src/lib/vocab/assignment/assignmentServices.ts
```diff
@@ -61,9 +61,8 @@ export const DeleteAssignmentService: ServiceByName['DeleteAssignment'] = {
 			return {ok: false, status: 405, message: 'cannot leave a personal hub'};
 		}
 		if (hub_id === ADMIN_HUB_ID) {
-			const adminAssignmentsCount = await repos.assignment.countAccountActorAssignmentsByHub(
-				hub_id,
-			);
+			const adminAssignmentsCount =
+				await repos.assignment.countAccountActorAssignmentsByHub(hub_id);
```
ryanatkn, line 65: more prettier changes like this below

vite.config.ts
```diff
@@ -1,12 +1,12 @@
...
 
-/** @type {import('vite').UserConfig} */
-const config = {
+const config: UserConfig = {
 	plugins: [sveltekit()],
 	ssr: {noExternal: ['@feltjs/felt-ui']},
 	server: {
 		proxy: {
-			'/api': 'http://localhost:3000', // equal to `PUBLIC_API_SERVER_HOSTNAME + ':' + PUBLIC_API_SERVER_PORT`
+			'/api': 'http://localhost:3000', // equal to `PUBLIC_SERVER_HOSTNAME + ':' + PUBLIC_SERVER_PORT`
```
ryanatkn, line 9: I removed (I think) all usage of the API prefix for the server identifiers, because we have only one server.

src/lib/infra/deploy.task.ts
greatbacon: Would you do `GRO_DIST_DIRNAME` or `GRO_BUILD_DIST_DIRNAME`?

src/lib/server/ApiServer.ts
greatbacon: I think leveraging `.gro/dist/server` to keep things simple for now makes sense

src/lib/server/api.ts
greatbacon: I really appreciated these prettier callouts 🥇

src/lib/server/password.worker.ts
greatbacon: the `.worker` sub-type is nice too

src/lib/ui/ui.ts
```diff
@@ -106,13 +104,11 @@ export interface Ui {
...
 
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export const toUi = (
 	$session: ClientSession,
 	initialMobile: boolean,
-	components: {[key: string]: typeof SvelteComponent<any>},
-	contextmenuLinkComponent: ComponentType<ContextmenuLinkEntry>,
-	contextmenuTextComponent: ComponentType<ContextmenuTextEntry>,
+	components: {[key: string]: typeof SvelteComponent<any>}
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
greatbacon, line 111: What happened here?

src/lib/util/testHelpers.ts
```diff
@@ -1,9 +1,10 @@
-import sourcemapSupport from 'source-map-support';
 import {Logger} from '@feltjs/util/log.js';
 import type {OmitStrict} from '@feltjs/util/types.js';
 import {unwrap} from '@feltjs/util/result.js';
 import * as assert from 'uvu/assert';
 import {noop} from '@feltjs/util/function.js';
+// TODO BLOCK
+// import sourcemapSupport from 'source-map-support';
```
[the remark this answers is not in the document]
greatbacon: I agree 100%; I think sourcemaps as a follow up makes a ton of sense

src/lib/vocab/hub/hubHelpers.server.ts
```diff
@@ -34,17 +34,17 @@ export const HUB_COLUMNS = {
...
-			hub_id,
-		);
-		if (accountActorAssignmentsCount === 0) {
-			log.debug('no assignments found for hub, cleaning up', hub_id);
-			await deleteHub(repos, hub_id); // eslint-disable-line no-await-in-loop
-			(deleted || (deleted = [])).push(hub_id);
-		}
-	}
+	await Promise.all(
+		hubIds.map(async (hub_id) => {
```
greatbacon, line 38: 🚗 💨

greatbacon (commented):
Wowza! Okay, just looked over everything. Lots of files, but a pretty straightforward set of changes. Thanks again for taking on such a heavy lift!!!  

I know CI is broken right now, and I wanted to wait to do final approval until we've had a successful deploy to staging from this branch.

But overall, lgtm! 👏

src/lib/ui/ui.ts
ryanatkn: There's the improvement where the link and text component are no longer required -- that was a hack we needed because our tests couldn't import Svelte files wiithout crashing.

The eslint-disable moved down because of a dependency update, I think it's prettier and eslint interacting.

src/lib/ui/ui.ts
```diff
@@ -162,7 +158,7 @@ export const toUi = (
 				map.set(hub_id, hubSpaces);
 			}
 			return map;
-		},
+		}
```
ryanatkn, line 161: these were a mistake, will be reverted

tsconfig.json
```diff
@@ -11,11 +12,22 @@
...
+	},
+	"include": [
+		"./gro.config.ts",
+		"./.svelte-kit/ambient.d.ts",
+		"./.svelte-kit/types/**/$types.d.ts",
+		"./vite.config.ts",
+		"./src/**/*.js",
+		"./src/**/*.ts",
+		"./src/**/*.svelte"
+	]
```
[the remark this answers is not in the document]
ryanatkn: I added these lines from theirs (was a surprise that the svelte config isn't added on their end):

```
		"./gro.config.ts",
		"./svelte.config.js",
```

vite.config.ts
ryanatkn: So "server" is the server in `$lib/server/server.ts` and "dev server" is Vite.

src/lib/server/password.worker.ts
ryanatkn: glad you like it, it's required by the Gro plugins to bundle correctly!

src/lib/infra/deploy.task.ts
ryanatkn: I think "dist" will be `dist/` owned by SvelteKit and "Gro dist" will be `.gro/dist`, similar to how "server" and "dev server" work - https://github.com/feltjs/felt/pull/922/files/d17cccf9622f016fe72db4aedb5f261a260d3978#r1334905159

src/lib/vocab/hub/hubHelpers.server.ts
ryanatkn: was causing eslint weirdness so I just refactored instead of disabling the rules lol

gro.config.ts
```diff
@@ -0,0 +1,50 @@
...
+			}),
+			// TODO could replace just the server one instead being more forwards compatible, `replace_plugin` helper?
+			gro_plugin_sveltekit_frontend(),
+		],
+	};
+
+	return final_config;
+};
+
+export default config;
```
ryanatkn, line 50: this is where most of the remaining work is -- initial attempts aren't building correctly, may be work needed in gro

src/lib/server/ApiServer.ts
ryanatkn: sounds good, `gro build` already deletes `.gro/dist` before doing anything

src/lib/infra/deploy.task.ts
greatbacon: Cool, I can dig it 👍

src/static/.well-known/package.json
```diff
@@ -0,0 +1,151 @@
+{
```
ryanatkn, line 1: documented this here, definitely open to feedback - https://github.com/grogarden/gro/blob/main/src/lib/docs/config.md#package_json

src/lib/sync.task.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const task: typeof baseTask = {
+	...baseTask,
+	run: async ({invoke_task, args}) => {
+		// TODO add a hook to gro to do this in the config,
+		// and when it returns `true` reload the process
+		await invoke_task('infra/syncEnvGitHash');
+		return invoke_task('gro/sync', args);
+	},
+};
```
ryanatkn, line 11: Previously this was an override of `dev.task.ts`, so it's mostly just a rename -- `gro sync` is a new Gro builtin task that it internally calls during `gro dev` and `gro build` and some others. The `sync` task calls `gro gen` and `gro exports` which are the same kind of thing as this `syncEnvGitHash`.

src/static/.well-known/package.json
greatbacon: Docs seemed sensible to me on a first pass. They answered my big question which was "why are we doing this" so I'd consider that a great success all around 😆

greatbacon (commented):
howdy traveler, 🤠 🔥  feel free to take a little break at this comment with me, treat it like a checkpoint.

ryanatkn:
I think it's gtg, something failing in a git call on CI but not locally, diagnosing

src/lib/infra/setup.task.ts
```diff
@@ -1,9 +1,22 @@
...
+	PGDATABASE,
+	PGUSER,
+	PGPASSWORD,
+} from '$env/static/private';
+import {
+	PUBLIC_DEPLOY_SERVER_HOST,
+	PUBLIC_ADMIN_EMAIL,
+	PUBLIC_SERVER_HOSTNAME,
+	PUBLIC_SERVER_PORT,
+} from '$env/static/public';
```
ryanatkn, line 17: I'm wondering if these environment variables used in tasks should be loaded explicitly from the `.env.production` file instead of imported like this. The idea, which I'm not sure about yet, is that tasks should continue expecting a dev environment, and so any modules they import will be in dev mode too. It's a subtle difference but I could see it making more sense, and it seems way easier to fix.

greatbacon: That makes sense to me. This is kind of a corner case where we're a.) loading vars for a task & b.) needing them to be the prod file. Especially if it's an easier fix that way.

ryanatkn: Cool yeah I think this is correct - https://github.com/feltjs/felt/pull/922/commits/b17b7314c70cf582e0dae57cb45170fd8b67f162

ryanatkn:
Still at least one issue to solve with `gro infra/deploy`.

package.json
```diff
@@ -21,49 +21,50 @@
     "build": "gro build",
     "test": "gro test",
     "deploy": "gro deploy",
-    "start": "node dist/server/lib/server/server.js"
+    "start": "node ./dist_server/server/server.js",
+    "after_deploy": "node ./dist_server/infra/after_deploy.js"
```
ryanatkn, line 25: moving this here seems like a good idea, so tools can just call `npm run after_deploy` instead of needing to know any paths -- I also just realized we could use Gro's package.json automations to keep this derived from a variable

package.json
```diff
@@ -21,49 +21,50 @@
     "build": "gro build",
     "test": "gro test",
     "deploy": "gro deploy",
-    "start": "node dist/server/lib/server/server.js"
+    "start": "node ./dist_server/server/server.js",
+    "after_deploy": "node dist_server/infra/after_deploy.js"
```
ryanatkn, line 25: this script lets tools run `npm run after_deploy`, and it's generated automatically from a variable with Gro's package.json automations

src/lib/infra/.env.default
```diff
@@ -1,2 +1,2 @@
-PUBLIC_ADMIN_EMAIL_ADDRESS=TODO@TODO.com
-PUBLIC_ADMIN_ICON=💚
+PUBLIC_ADMIN_EMAIL=TODO@TODO.com
+PUBLIC_INSTANCE_ICON=💚
```
ryanatkn, line 2: just renamed `PUBLIC_ADMIN_ICON` to `PUBLIC_INSTANCE_ICON`, I think that's strictly better?

src/lib/ui/About.svelte
```diff
@@ -1,44 +1,61 @@
 <script lang="ts">
 	import {base} from '$app/paths';
-	import {PUBLIC_ADMIN_EMAIL_ADDRESS, PUBLIC_ADMIN_ICON, PUBLIC_GIT_HASH} from '$env/static/public';
+	import {PUBLIC_ADMIN_EMAIL, PUBLIC_GIT_HASH} from '$env/static/public';
 	import {VERSION as SVELTEKIT_VERSION} from '@sveltejs/kit';
+
+	import package_json from '../../static/.well-known/package.json';
```
ryanatkn, line 6: using the new public `package.json` feature to get better displayed info/links below, lots more could be done too

src/lib/util/constants.ts
```diff
@@ -6,3 +8,9 @@ export const ADMIN_HUB_ID = 1;
...
+
+export const VERSION = '0.7.0';
```
ryanatkn, line 14: now generating this file to bake in the `package_json.version` without loading the whole JSON file - I could see just paying that cost but this is simple enough, and I think having a generated constants file is going to be useful with little downside.

src/lib/infra/deploy.task.ts
```diff
@@ -1,75 +1,89 @@
...
+			deploy_login,
+			`
+				mkdir ${deploy_dirname};
+				mv ${artifact_filename} ${deploy_dirname}/;
+				cd ${deploy_dirname};
+				tar -xf ${artifact_filename};
+				echo 'stopping old deploy';
+				pm2 stop default;
+				echo 'npm ci';
+				npm ci;
```
ryanatkn, line 63: I'm probably not understanding several things, had a question specifically about `pm2 stop default;` can we do `npm ci` and the copying below while the current server is still running to minimize downtime?

src/lib/infra/deploy.task.ts
```diff
@@ -1,75 +1,89 @@
...
+			deploy_login,
+			`
+				cd ${deploy_dirname};
+				npm run after_deploy;
+				pm2 start npm -- run start --prefix ~/${CURRENT_DEPLOY_DIRNAME};
+			`,
 		]);
-
-		log.debug('starting new server deployment');
-		await spawn('ssh', [deployLogin, `pm2 start npm -- run start --prefix ~/${currentDeploy}`]);
```
ryanatkn, line 73: combined this with ssh call above

gro.config.ts
```diff
@@ -0,0 +1,81 @@
...
+				types: './dist/docs/docs.d.ts',
+			},
+		};
+		return pkg;
+	};
+
+	return cfg;
+};
+
+export default config;
```
ryanatkn, line 81: big fan of this API

src/lib/infra/setup.task.ts
```diff
@@ -18,44 +18,47 @@ type Args = z.infer<typeof Args>;
...
-			apiServerHost,
+			server_host,
 			REMOTE_NGINX_HTML_DIR,
 		);
-		const nginxHtmlSource = render_502_html(PUBLIC_ADMIN_EMAIL_ADDRESS);
+		const nginxHtmlSource = render_502_html(PUBLIC_ADMIN_EMAIL);
 
 		// This file is used to detect if the setup script has already run.
-		const FELT_SETUP_STATE_FILE_PATH = '~/felt_state_setup';
+		const APP_SETUP_STATE_FILE_PATH = '~/app_setup_state';
```
ryanatkn, line 57: apologies for this rename, to migrate:

```bash
ssh ...........
mv ~/felt_state_setup ~/app_state_setup
```

src/lib/server/server.ts
```diff
@@ -3,16 +3,16 @@ import {createServer} from 'http';
...
 import {ApiServer} from '$lib/server/ApiServer';
 import {Websockets} from '$lib/server/Websockets';
 import {services} from '$lib/server/services';
 import {db} from '$lib/db/db';
-import {fromEnv} from '$lib/server/env';
 import {Broadcast} from '$lib/server/Broadcast';
 
 // Global logging setup
-if (process.env.NODE_ENV === 'production') {
+if (!import.meta.env.DEV) {
```
ryanatkn, line 15: there's al;so `import.meta.env.PROD` if we don't want to negate, idk - https://vitejs.dev/guide/env-and-mode.html#env-variables

src/lib/infra/deploy.task.ts
greatbacon: I was noticing issues where `npm ci` would fail if the server was still running, which is why I added the `pm2 stop default` line above it. There might be a way around it, but I haven't really dug into it in that way.

ryanatkn: gotcha I'll add a todo note

greatbacon (approved):
Alright! I just finished smoketesting, so I think we're gtg!!!! 🎉 🎉 🎉  Unless you have any outstanding TODOs for this.

ryanatkn:
omg

PR #924, opened 2023-08-30
fix types for after service hook

ryanatkn:
Fixes some types for #918.

I changed the name from `AfterResponse` to `AfterRequest` - I think it's a better name, but we may be able to do better. Thinking about the lifecycle of a service, we have request -> reponse, and this runs after the response. Future possible `AfterRequest` and/or `BeforeReponse` hooks would run between the service returning a value and the middleware sending it to clients. I imagine we may want this down the road for more sophisticated behavior composition with services -- e.g. sending off an email without changing the internals of the actual service function.

`AfterService` comes to mind but I don't think I like it better. Any thoughts?


src/lib/util/testHelpers.ts
```diff
@@ -75,8 +74,7 @@ export function toServiceRequestFake(
 	account_id = actor?.account_id || undefined,
 	broadcast: IBroadcast = new BroadcastFake(),
 	passwordHasher = passwordHasherFake,
-	afterRequest = (cb: PostRequestCallback) =>
-		(postRequestCallbacks || (postRequestCallbacks = [])).push(cb),
+	afterResponse = noop,
```
ryanatkn, line 77: I think this was the issue you ran into, this code is gnarly with the combination of bypasses for tests and all of the parameterization. The way it is now, a caller can provide their own `afterReponse` argument if they care about testing its usage by the service, but in general it can be ignored.

greatbacon: Yup, this was the last type issue I was running into. I had no idea we even had a `noop` util lol

greatbacon (approved):

PR #927, opened 2023-08-31
remove data.directory

ryanatkn:
Removes `entity.data.directory` in favor of comparing `entity.entity_id` and `entity.directory_id`, if they're the same it's a directory - adds the helper `isDirectory` to do this comparison, mirroring the SQL we now write.

Clarifies that `entity.data` isn't for system-level concerns like this, similar to how `lists`/`items` will be hoisted.

background info: https://github.com/feltjs/felt/pull/921/files#r1310429778


src/lib/vocab/entity/entityData.ts
```diff
@@ -14,30 +12,18 @@ export interface BaseEntityData {
...
-
-export interface NoteEntityData extends BaseEntityData {
-	type?: 'Note' | undefined;
-	content: string;
-}
-
-export interface ArticleEntityData extends BaseEntityData {
-	type: 'Article';
-	content: string;
-	name: string;
```
ryanatkn, line 36: I removed these because they weren't doing anything helpful

src/lib/vocab/entity/entityHelpers.ts
```diff
@@ -6,9 +6,13 @@ import type {Ui} from '$lib/ui/ui';
 import {Logger} from '@feltjs/util/log.js';
 import type {Actions} from '$lib/vocab/action/actionTypes';
 import type {ActorId} from '$lib/vocab/actor/actor';
+import type {Directory} from '$lib/vocab/entity/entityData';
 
 const log = new Logger('[entityHeler]');
 
+export const isDirectory = (entity: Entity): entity is Directory =>
+	entity.entity_id === entity.directory_id;
```
ryanatkn, line 14: this `entity is Directory` syntax is the special TypeScript construct that does type narrowing for us, so `entity.data.type  === 'Collection'` should be typed

greatbacon (approved):

PR #929, opened 2023-09-06
disable node polyfill

ryanatkn:
we use the latest node so we don't need this - https://kit.svelte.dev/docs/adapter-node#options-polyfill


greatbacon (approved):
🦾

PR #930, opened 2023-09-06
improve service request type

ryanatkn:
Improving our gnarliest type `toServiceRequest` along with the related `toServiceRequestFake`. It comes with a LOC increase but simplifies the concepts. They now use extended base types so the objects have the same shapes, simplifying construction code, and it's clearer when reading to see what each subtype does.

Found several small inconsistencies and improvements (like removing typecasts) as a result.

I also rearranged two params of `toServiceRequestFake` to better align with the non-fake function signature, and was able to remove a few `undefined, ` snippets from args lists as a result.


src/lib/server/service.ts
```diff
@@ -104,60 +104,91 @@ export interface AuthorizedService<
...
 	repos: Repos;
 	params: TParams;
+	account_id: undefined | AccountId;
+	actor: undefined | ActionActor;
 	session: ISessionApi;
 	broadcast: IBroadcastApi;
 	passwordHasher: PasswordHasher;
 	afterResponse: AfterResponse;
+	checkPolicy: undefined | CheckPolicy;
+	checkHubAccess: undefined | CheckHubAccess;
```
ryanatkn, line 129: The main change here is to use a `BaseServiceRequest` type and use the same object shame for all cases. The extending types can then declare if they're `undefined` or a value for whichever they apply to.

greatbacon (approved):
Nice, this doesn't seem too much more complex than before, but definitely looks like it makes type stuff a little easier!

PR #934, opened 2023-09-08
add `mutate` to MutationContext API

ryanatkn:
Looks like we weren't using `mutate` in cases where we had a single change, but I think we should use it everywhere?


src/lib/vocab/action/actions.ts
```diff
@@ -38,6 +38,7 @@ export const toActions = (
 			return mutation({
 				actionName,
 				params,
+				mutate: ui.mutate,
```
ryanatkn, line 41: I think `afterMutation` belongs here too? It means threading through the param explicitly elsewhere, but I think that's a good thing? It's possible too that `mutate` and `afterMutation` will no longer belong exported on the `ui` object in general, but I want to do one step at a time I think.

src/lib/ui/mutation.ts
```diff
@@ -12,6 +12,7 @@ export interface MutationContext<
 > {
 	actionName: string;
 	params: TParams;
+	mutate: MutateUi;
```
ryanatkn, line 15: I think `afterMutation` belongs here too? It means threading through the param explicitly elsewhere to some helpers instead of just passing `ui` everywhere, but I think that's a good thing? It's possible too that `mutate` and `afterMutation` will no longer belong exposed on the `ui` object in general, but I want to go one step at a time.

src/lib/vocab/action/actions.ts
```diff
@@ -75,6 +76,7 @@ export const toActionsBroadcastMessage =
 		return mutation({
 			actionName,
 			params,
+			mutate: ui.mutate,
```
ryanatkn, line 79: This module is where we'd put the state for `mutate` and `afterMutation`, instead of `ui`, I think.

src/lib/ui/mutation.ts
greatbacon: Yeah, this stepwise unwinding makes sense to me. We also thread the `afterResponse` mirror object through our services, so it doesn't seem out of place to do that for `afterMutation` here too.

greatbacon (approved):
This is a great iterative approach to shifting the `mutation` concern out of ui! 🙌

PR #936, opened 2023-09-11
Add missing errorMessage to KickFromHubForm

greatbacon:
Hoists error messages from KickActorItem back to KickFromHubForm & displays them


src/lib/ui/KickActorItem.svelte
```diff
@@ -14,15 +14,21 @@
...
+		const result = await actions.KickFromHub({
 			actor: $actor.actor_id,
 			actor_id: $communityActor.actor_id,
 			hub_id: $hub.hub_id,
 		});
+		if (result.ok) {
+			setErrorMessage(null);
+		} else {
+			setErrorMessage(result.message);
+		}
```
ryanatkn, line 31: Instead of adding hoisted API for this, I think it'd be better to just inline the error message underneath the `PendingButton` in this component.

ryanatkn (approved):

PR #937, opened 2023-09-12
add `createMutationContext` to encapsulate state with `mutate` and `afterMutation`

ryanatkn:
Removes `mutate` and `afterMutation` from the `ui` object, clarifying its role as a namespace for data, and hoists them to the `MutationContext`, which now has its state encapsulated in the new `createMutationContext`.

This change means explicitly threading `afterMutation` through helpers that need it. This keeps things simple and explicit in the mutation helpers, if verbose, rather than passing the entire mutation context everywhere, which is too large a surface area to reason about. This also means we'll experience more function signature churn, but the types should keep things tidy.

future

- add `mutate` usage to singular mutations - see `uiMutations.ts`


src/lib/util/mutation.ts
```diff
@@ -0,0 +1,74 @@
...
+	return {
+		actionName,
+		params,
+		ui,
+		actions,
+		invoke,
+		mutate,
+		afterMutation,
+	};
+};
```
ryanatkn, line 74: moves stuff that was previously in `$lib/ui/mutation.ts` into this `$lib/util` module -- I think this is a good pattern, moving the non-Svelte stuff from `ui/` into `util/`

greatbacon: Yeah I can dig it

src/lib/vocab/actor/actorMutationHelpers.ts
```diff
@@ -43,13 +44,11 @@ export const stashActors = (
 	}
 };
 
-export const evictActors = (ui: WritableUi, actorsToEvict: Set<Writable<ClientActor>>): void => {
```
greatbacon, line 46: were we not calling evictActors anywhere?

ryanatkn: apparently not, only the singular version -- we may want this in the future, but a loop at the callsite might be fine too

greatbacon (approved):

PR #938, opened 2023-09-13
Add static 502.html page to nginx for server downtime

greatbacon:
Adds an extra step to the nginx configuration that includes a static 502 page for brief server downtime when restarting/deploying Felt.

I added the ASCII heart cause I thougt it was fun & didn't want to deal with any additional assets, same reason I just inlined the CSS

Theoretically most users shouldn't even see this page, but I wanted something a little nicer than the default NGINX error page for the 30 seconds Felt isn't running on deploys

Follow up:
more dynamic heart scaling with @media css


src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
...
+                      #####################################                     
+                        ################################                        
+                           ##########################                           
+                              #####################                             
+                                ################                                
+                                  ############                                  
+                                    ########                                    
+                                      ####                                      
+  `;
+};
```
ryanatkn, line 95: love the heart!

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
...
 
   `.trim();
 };
+
+// Outputs an simple html file for nginx to server during Felt downtime
+export const to502File = (PUBLIC_ADMIN_EMAIL_ADDRESS: string): string => {
+	return `
+  <html>
+    <head>
+    </head>
```
ryanatkn, line 49: copying from app.html to prevent weirdness in some browsers/devices -

```html
		<meta charset="utf-8" />
		<meta name="viewport" content="width=device-width" />
```

probably isn't worth inlining the favicon

src/lib/infra/nginxConfig.ts
```diff
@@ -4,6 +4,7 @@ import {HEARTBEAT_INTERVAL} from '$lib/ui/socket';
 export const toNginxConfig = (
 	PUBLIC_DEPLOY_SERVER_HOST: string,
 	API_SERVER_HOST_PROD: string,
+	REMOTE_NGINX_502_PATH: string,
```
ryanatkn, line 7: the `REMOTE_NGINX_502_LOCATION` is passed this function but it's renamed to `REMOTE_NGINX_502_PATH` here, and is confused with the real `REMOTE_NGINX_502_PATH`

src/lib/infra/setup.task.ts
```diff
@@ -40,8 +41,15 @@ export const task: Task<Args> = {
 
 		const REMOTE_NGINX_CONFIG_PATH = '/etc/nginx/sites-available/felt.conf';
 		const REMOTE_NGINX_SYMLINK_PATH = '/etc/nginx/sites-enabled/felt.conf';
+		const REMOTE_NGINX_502_LOCATION = '/var/www/html';
```
ryanatkn, line 44: convention here is `DIR` not `LOCATION`

src/lib/infra/setup.task.ts
```diff
@@ -40,8 +41,15 @@ export const task: Task<Args> = {
...
+		const REMOTE_NGINX_502_PATH = `${REMOTE_NGINX_502_LOCATION}/502.html`;
 
-		const nginxConfig = toNginxConfig(PUBLIC_DEPLOY_SERVER_HOST, apiServerHost);
+		const nginxConfig = toNginxConfig(
+			PUBLIC_DEPLOY_SERVER_HOST,
+			apiServerHost,
+			REMOTE_NGINX_502_LOCATION,
+		);
+		const nginx502File = to502File(PUBLIC_ADMIN_EMAIL_ADDRESS);
```
ryanatkn, line 52: elsewhere we use the `File` suffix to mean a file path, I usually avoid the word because of ambiguity altogether but this conflicts with what we have elsewhere -- `source` is the build tooling word, `contents` is also clear

src/lib/infra/setup.task.ts
```diff
@@ -113,6 +121,8 @@ export const task: Task<Args> = {
 				echo '${nginxConfig}' >> ${REMOTE_NGINX_CONFIG_PATH};
 				cat ${REMOTE_NGINX_CONFIG_PATH};
 				ln -s ${REMOTE_NGINX_CONFIG_PATH} ${REMOTE_NGINX_SYMLINK_PATH};
+				touch ${REMOTE_NGINX_502_PATH};
+				echo '${nginx502File}' >> ${REMOTE_NGINX_502_PATH}
```
ryanatkn, line 125:
```suggestion
				echo '${nginx502File}' >> ${REMOTE_NGINX_502_PATH};
```

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
     proxy_send_timeout ${websocketTimeout};
   }
 
+  error_page 502 /502.html;
+  location /502.html {
+      root ${REMOTE_NGINX_502_PATH};
```
ryanatkn, line 32: this seems weird that it hardcodes 502.html here twice, but includes it in the `_PATH` -- I think hardcoding is fine, but then it should be more like `REMOTE_NGINX_HTML_DIR`

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
...
+                   ##########################################                   
+                      #####################################                     
+                        ################################                        
+                           ##########################                           
+                              #####################                             
+                                ################                                
+                                  ############                                  
+                                    ########                                    
+                                      ####                                      
+  `;
```
ryanatkn, line 94:
```suggestion
`;
```

whitespace will make it looks weird when selected

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
...
+	return `
+  <html>
+    <head>
+    </head>
+    <body style="display: flex; flex-direction: column; justify-content: center;  align-items: center;">
+      <h1>Felt is temporarily down</h1>
+      <h2>If this page persists, please contact ${PUBLIC_ADMIN_EMAIL_ADDRESS}</h2>
+      <div style="white-space: pre-wrap; font-family: monospace;">
+${toASCIIFeltLogo()}
+        </div>
```
ryanatkn, line 55:
```suggestion
      <pre>
${toASCIIFeltLogo()}
</pre>
```

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
     proxy_send_timeout ${websocketTimeout};
   }
 
+  error_page 502 /502.html;
```
ryanatkn, line 30: is `503` a better user-facing code? (assuming it works the same here? or maybe it needs to be `502`?

> The server cannot handle the request (because it is overloaded or down for maintenance). Generally, this is a temporary state.

versus 

> The server was acting as a gateway or proxy and received an invalid response from the upstream server.

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,69 @@ server {
...
+      <div style="white-space: pre-wrap; font-family: monospace;">
+${toASCIIFeltLogo()}
+        </div>
+      </body>
+    </html>
+  `.trim();
+};
+
+//TODO would this be fun to have in felt-util?
+export const toASCIIFeltLogo = (): string => {
```
ryanatkn, line 62:
```suggestion
export const toAsciiFeltLogo = (): string => {
```

the convention for acronyms we use is to treat them as words, so `httpThing` and `thingHttp` instead of `hTTPThing` and `thingHTTP`.

src/lib/infra/nginxConfig.ts
greatbacon: Yeah, the browser caches the favicon so I skipped dealing with that, glad you fell the same.

src/lib/infra/nginxConfig.ts
greatbacon: 503 might be semantically better, but 502 is what NGINX is throwing so that's what I'm trying to catch. I'll double check the docs & see if I can maybe just blanket cover 500 errors from NGINX though.

src/lib/infra/nginxConfig.ts
greatbacon: Yeah, I was waffling on which I preferred. I think I'll stick to just hardcoding 502 stuff and only keep the DIR a variable.

src/lib/infra/setup.task.ts
```diff
@@ -40,8 +41,14 @@ export const task: Task<Args> = {
 
 		const REMOTE_NGINX_CONFIG_PATH = '/etc/nginx/sites-available/felt.conf';
 		const REMOTE_NGINX_SYMLINK_PATH = '/etc/nginx/sites-enabled/felt.conf';
+		const REMOTE_NGINX_502_DIR = '/var/www/html';
```
ryanatkn, line 44: this name still doesn't make sense, I had suggested `REMOTE_NGINX_HTML_DIR`

src/lib/infra/nginxConfig.ts
```diff
@@ -26,10 +27,71 @@ server {
...
     proxy_pass http://${API_SERVER_HOST_PROD};
   }
+
 }
 
   `.trim();
 };
+
+// Outputs an simple html file for nginx to server during Felt downtime
+export const to502File = (PUBLIC_ADMIN_EMAIL_ADDRESS: string): string => {
```
ryanatkn, line 45: name hasn't been updated

src/lib/infra/nginxConfig.ts
greatbacon: I decided to just go for a small font-size that renders the heart on both mobile & desktop at a readable size. Since this is a simple page that 99% of users won't see, I'm trying to keep it as simple as possible.

src/lib/infra/setup.task.ts
greatbacon: sorry, I only noticed the DIR had changed, not HTML too.

greatbacon:
also, just tried replacing all the `#`s with `💚`s & it messes with the spacing too much and kind of smears the heart to the right :(

ryanatkn (approved):
gotcha, when we do the scaling we can try fixing it with a hardcoded character width (probably desirable anyway because of line-height differences across browsers - we basically just want each to be a square regardless of content, and rendering perf doesn't matter so we could wrap each in a `span`)

PR #940, opened 2023-09-17
update names related to files

ryanatkn:
Clarifies some of our names around filesystem primitives.

The pattern, which isn't great yet:

- a `path` is typically relative to our project root, but there's a lot of ambiguity here - it's basically a catch-all that could be replaced with a more context-specific name
- a `basepath` is a [`base` in Node](https://nodejs.org/api/path.html#pathparsepath) (this link has a nice little ASCII diagram), returned by `basename` and a property of `path.parse` -- I'm not 100% on this, but `base` seems too ambiguous. It's weird because Node's `basename` returns the `base` when passed a single arg, and a `basename` when the `ext` is passed as the second arg, so this was already a little confusing. I haven't looked at Bun/Deno terminology here, but they have Node-compatible APIs so this is at least a more standardized step forward.
- an `id` is a Rollup term also used in Gro for an absolute path, e.g. `source_id`, or `file_id` for ambiguous contexts, or `build_id`, etc - not used in this PR, but relevant for the future (this is distinct from a `file_url`/`FileUrl` that has the `file://` prefix, and we may extract a `FileId` or `PathId` type helper when needed)
- `contents` means the data of a file - dealing with the string vs buffer is complexity we don't need in this PR
- `file` would be reserved for [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) on the web platform, which we don't use anywhere yet (I didn't update the `envFile` usage here because we upstream that concern to Gro in the upcoming update)
- `filename` is used interchangeably in the Node docs and the Svelte preprocessor API for `path` it seems, as a catch-all to some relative directory. I wanted to use this instead of `basepath` but I don't want to be ambiguous with Svelte and Node.

wew lol. names really matter here, path string handling is super error-prone and the type primitives will be nice

Also updates some identifiers to `snake_case`. (when they don't extend beyond the current module or when I renamed the thing anyway)


src/lib/infra/nginxConfig.ts
```diff
@@ -1,7 +1,8 @@
 import {HEARTBEAT_INTERVAL} from '$lib/ui/socket';
+import {render_ascii_felt_logo} from '$lib/util/logo';
 
 // Outputs an nginx config with configured values.
-export const toNginxConfig = (
+export const render_nginx_config = (
```
ryanatkn, line 5: `render` is a good term that SvelteKit uses to avoid a verbose `contents` suffix

src/lib/infra/nginxConfig.ts
```diff
@@ -42,72 +43,34 @@ server {
...
+  </head>
+  <body>
+    <div class="headers">
+      <h1>Felt is temporarily down</h1>
+      <h2>If this page persists, please contact ${PUBLIC_ADMIN_EMAIL_ADDRESS}</h2>
+    </div>
+    <pre class="heart">${render_ascii_felt_logo()}</pre>
+  </body>
+</html>
+`.trim();
```
ryanatkn, line 76: this was indented and had some trailing whitespace on lines -- we may just want to pass these through the formatter, Gro exports those utils

src/lib/infra/nginxConfig.ts
```diff
@@ -42,72 +43,34 @@ server {
...
-        <h1>Felt is temporarily down</h1>
-        <h2>If this page persists, please contact ${PUBLIC_ADMIN_EMAIL_ADDRESS}</h2>
-      </div>
-      <pre class="heart">${toAsciiFeltLogo()}</pre>
-    </body>
-  </html>
-  `.trim();
-};
-
-//TODO would this be fun to have in felt-util?
```
ryanatkn, line 78: yeah!

src/lib/infra/syncEnvGitHash.task.ts
```diff
@@ -7,25 +7,25 @@ export const task: Task = {
 	run: async ({fs, log, dev}) => {
 		initEnv();
 
-		const envFile = dev ? ENV_FILE_DEV : ENV_FILE_PROD;
+		const env_basepath = dev ? ENV_FILE_DEV : ENV_FILE_PROD;
```
ryanatkn, line 10: I'll probably rename the `ENV_FILE_` identifiers after upgrading Gro, it deals with their behavior.

src/lib/util/logo.ts
```diff
@@ -0,0 +1,33 @@
...
+                      #####################################                     
+                        ################################                        
+                           ##########################                           
+                              #####################                             
+                                ################                                
+                                  ############                                  
+                                    ########                                    
+                                      ####                                      
+                                       ##                                       
+`;
```
ryanatkn, line 33: this is a possibly client-facing util now, not just server-side, so I'm avoiding `.trim()`

src/lib/util/logo.ts
```diff
@@ -0,0 +1,32 @@
...
+                   ##########################################                   
+                      #####################################                     
+                        ################################                        
+                           ##########################                           
+                              #####################                             
+                                ################                                
+                                  ############                                  
+                                    ########                                    
+                                      ####                                      
+                                       ##                                       `;
```
ryanatkn, line 32: this is a possibly client-facing util now, not just server-side, so I'm avoiding `.trim()`, and because it gets inserted in `pre` we don't want any unnecessary newlines

src/lib/util/fs.ts
```diff
@@ -0,0 +1,5 @@
+import type {Flavored} from '@feltjs/util/types';
+
+// TODO probably upstream to `@feltjs/util`
```
ryanatkn, line 3: Gro has some other primitives like `SourceId` here that probably belong upstreamed with it

src/lib/infra/setup.task.ts
```diff
@@ -4,7 +4,7 @@ import {z} from 'zod';
 
 import {green, red} from '$lib/server/colors';
 import {fromEnv} from '$lib/server/env';
-import {to502File as toHtmlSource, toNginxConfig} from '$lib/infra/nginxConfig';
```
ryanatkn, line 7: one of the rename mistakes slipped through here

src/lib/db/migration.ts
```diff
@@ -1,4 +1,7 @@
 import {stripAfter} from '@feltjs/util/string.js';
+import {readdir} from 'fs/promises';
```
greatbacon, line 2: 👀

ryanatkn: oops I forgot a tiny detail, changing to `node:fs/promises` - using the node namespace everywhere ensures less lockin and more clarity going forward

PR #944, opened 2023-09-27
convert mermaid diagrams to png

### LINKED ISSUE #943: optimize Mermaid diagrams at buildtime
update: will be closed by #944 faster than I expected - going with static pngs created by hand for now, automated soonish but probably avoiding a dev dep on Mermaid

I added [Mermaid](https://github.com/mermaid-js/mermaid) diagrams in a hurry (https://github.com/feltjs/felt/pull/879) to get our docs from `.md` on GitHub to `.svelte` on [`felt.dev/docs`](https://www.felt.dev/docs). It resulted in [working diagrams](https://www.felt.dev/docs/guide/admin/getting-started), but it's importing Mermaid at runtime, which has a very large cost and causes yellow text in the SvelteKit build output.

I'm pretty sure this could be done without the extra runtime cost. Compiling to something static is easy with gen if it fits - maybe preferable to Vite plugins? I need to read the Mermaid docs.

ryanatkn:
closes #943 

This removes the huge JS runtime dependency on mermaid in the docs and replaces it with small static pngs. This speeds up the build considerably  (now ~10 seconds from 30+ seconds), and makes the circular dependency warnings during the build go away. Also trims package-lock by 20%.

Going low tech here, using the online Mermaid diagram editor by hand to output pngs - https://mermaid.live/edit

This solution is the worst for storage in the git repo (not a concern? they're 25-50kb pngs, 3 at the moment), but the best for end user UX and minimizing dev complexity. We can automate this eventually and purge old pngs once we start making changes.

I'm including TS modules that will be used to generate the PNGs but they aren't used automatically yet. I'll probably add something around this in Gro to use Mermaid as a global npm dependency, because I don't think I want to include it as a `devDependency`.


package.json
```diff
@@ -46,7 +46,6 @@
     "eslint": "^8.44.0",
     "eslint-plugin-svelte": "^2.32.2",
     "eventemitter3": "^5.0.1",
-    "mermaid": "^10.2.4",
```
ryanatkn, line 49: wew at those package-lock LOCs

greatbacon (approved):

PR #946, opened 2023-09-30
upgrades

ryanatkn:
installs new dev deps and updates their imports, `@fuz.dev/fuz` (formerly `@feltjs/felt-ui`) is no longer a monolith


package.json
```diff
@@ -63,7 +67,6 @@
   },
   "dependencies": {
     "@feltcoop/svelte-gettable-stores": "^0.2.0",
-    "@fuz.dev/fuz": "^0.69.1",
```
ryanatkn, line 66: this may not work but I don't think our server needs to depend at runtime on these

ryanatkn, line 66: moving this to regular deps looks safe, I checked `dist_server/server/server.js`

greatbacon (approved):

PR #947, opened 2023-10-03
upgrade fuz

ryanatkn:
Upgrades Fuz deps. The only change is unifying a naming convention for CSS variables -- there's no more `_xl` or `_xlN` suffixes, `_xl` becomes `_1` and `_xl5` becomes `_5`. It's a quirk that resolves a lot of naming tension across properties.

I think the benefit is more evident when you look at [the full list of variables](https://www.fuz.dev/library/variables) at once. The suffixes `_sm` and `_lg` etc remain as useful semantic handles at the low ends, and then once you hit "XL" you just number them sequentially starting with `_1`. This is a complex set of tradeoffs and it seems like a good step forward. We may want to do away with the `_lg` etc and only use numbers, but that would be a more complex migration than this, so I'd want to be sure. Numbers fit a trend in Fuz's design that I see elsewhere, they give end users a lot of flexibility and avoid overstepping intent.

The `_xlN` properties were also hard to read because the `l` looks similar to a `1` in monospace font.  The `_xsN` properties remain, but they're less common and I'd like to improve them along these lines. Spacing is the main problem here, and I may split the ideas of `spacing` and `distance` in the future.


greatbacon (approved):

greatbacon:
Is there a way to alias one variable to another like `_md` points to `_5` or something like that? Could be a way to keep some of the more human level semantics near the middle of sizing but make the underlying variables all line up nicely with numbers.

PR #948, opened 2023-10-04
fix schema mutation shenanigans

ryanatkn:
update: these are the main changes, the root problem was both `json-schema-to-typescript` and a Gro function mutate input schemas, so depending on the order in which we ran tasks, we could see schemas in different states, leading to confusing bugs. I made a fix in Gro that also removed some phantom schema data that we'd discussed before.

- https://github.com/feltjs/felt/pull/948/commits/015251efa8844f2a3bd98c8337ad0e43dc21a414
- https://github.com/grogarden/gro/pull/418

Fixes the error on `gro check`. Minimal repro:

```
 gro check --no-typecheck -- gro test serviceHttpEndpoints.test
```

`gro test` works, which is strange because it's tests that are failing, but not in CI (??)

```
Error: reference "/schemas/EntityId" resolves to more than one schema
```

this is a logged value in `ajv/dist/compile/resolve.js` in the passing `gro test` -

```
traversing sch {
  '$id': '/schemas/ClearFreshnessParams',
  type: 'object',
  properties: { directory_id: { '$ref': '/schemas/EntityId' } },
  required: [ 'directory_id' ],
  additionalProperties: false
}
```

in the failing version, it has `EntityId` resolved -

```
traversing sch {
  '$id': '/schemas/ClearFreshnessParams',
  type: 'object',
  properties: {
    directory_id: {
      tsType: 'EntityId',
      tsImport: "import type {EntityId} from '$lib/vocab/entity/entity';",
      '$id': '/schemas/EntityId',
      type: 'number'
    }
  },
  required: [ 'directory_id' ],
  additionalProperties: false,
  id: 'ClearFreshnessParams'
}
```

narrowed down further, it happens with `gro test` if I run `sync` during it, I think it's the metadata gen or another gen file that's resolving the schemas in memory.

It seems the culprit was some data being mutated during the actionData gen file by the external library `json-schema-to-typescript`. Cloning before the call seems to fix the problem.

But CI still fails locally for me, this time to a different problem with gen during `check`, but `gro gen --check` passes.. It has to do with the mutations.


src/lib/server/httpServiceMiddleware.ts
```diff
@@ -11,7 +11,7 @@ import {
 	type AfterResponseCallback,
 	flushAfterResponseCallbacks,
 } from '$lib/server/service';
-import {validateSchema, toValidationErrorMessage} from '$lib/util/ajv';
+import {validateSchema, toValidationErrorMessage} from '$lib/server/ajv';
```
ryanatkn, line 14: I moved this module because we don't want clients to import it -- it also may be better named `validate.ts`

I could have added `.server.` to the file too if you prefer.

src/lib/vocab/action/actionTypes.gen.ts
```diff
@@ -118,7 +118,7 @@ export interface ServiceByName {
 	}, '')}
 }
 
-${await actionDatas.reduce(
+${await structuredClone(actionDatas).reduce(
```
ryanatkn, line 121: this is the fix, the library we call into mutates the params

src/lib/server/httpServiceMiddleware.ts
greatbacon: I think just moving it is fine

greatbacon (approved):

PR #949, opened 2023-10-06
change to the default static sveltekit directory

ryanatkn:
We previously put the `static/` SvelteKit directory in `src/static`, but SvelteKit's default is `static/` at the root. This changes to use their default. Gro no longer has any restrictions on the `src/` directory for builds that made this choice for me a while back.


greatbacon (approved):
Nice! 🧹

PR #951, opened 2023-10-07
update data model diagram

ryanatkn:
Updates the data model diagram. One is the `entities->items` rename, the other I'm not sure about.

I don't love the old pngs being in git history, but it doesn't seem that simple to purge them. I may do that in a followup soon. This one is 27kb but it could add up, and the whole repo's zip is 433kb.

That mermaid dependency is just so heavy. #943


src/lib/docs/data_model_diagram.ts
```diff
@@ -7,9 +7,10 @@ export const data_model_diagram = `classDiagram
   Actors <|-- Hubs
   Hubs <|-- Actors
   Hubs <|-- Roles
-  Roles <|-- Assignments
   Actors <|-- Assignments
+  Hubs <|-- Assignments
+  Roles <|-- Assignments
```
ryanatkn, line 12: moving `Roles` improved the layout

src/lib/docs/data_model_diagram.ts
```diff
@@ -7,9 +7,10 @@ export const data_model_diagram = `classDiagram
   Actors <|-- Hubs
   Hubs <|-- Actors
   Hubs <|-- Roles
-  Roles <|-- Assignments
   Actors <|-- Assignments
+  Hubs <|-- Assignments
```
ryanatkn, line 11: I'm not sure about this, it's technically in the schema but is it an implementation detail implied by the role?

greatbacon: I think this is right, since it's supposed to be an easy visualization of the underlying data schemas!

greatbacon (approved):
Yeah, I suppose this a potential problem waiting to happen, but I think for now it's fine. I think we can always rewrite history for items if we need to, right?

PR #952, opened 2023-10-07
add `FeltWindowHost` and `FeltWindowTenant`

ryanatkn:
This adds `FeltWindowHost` and `FeltWindowTenant` as they were removed from Fuz. It's probable we want these to be part of `felt-ui` or extracted into a standalone library, because `felt` is a heavy dependency if all you need is the tenant component.

`felt-ui` is no longer published -- one possibility is we find a more specialized name for these window communication components, and then rename `felt-ui` to `sketches`. The sketches depend on the tenant component but not anything else in `felt`, so it's a case where taking on the whole `felt` dependency seems wrong. (you shouldn't be asking as a user, "why am I installing a server for this?", it'll keep `felt` well-scoped and well-factored to do this I think)

Maybe we should republish `@feltjs/felt-ui` with these two components as the first things in it? Or `@feltjs/felt-tenant` or `@feltjs/felt-client` or something. It's a good time to consider the longer term view of how repos get laid out -- it may be that `felt-ui` shouldn't exist, so we shouldn't republish it without knowing.


greatbacon (approved):
I think maybe as far as rolling these out into their own package, would just `felt-window` be too narrow? 

If I remember correctly these are components that wouldn't just be in felt-client work or any felt specific plugins, but the Tenant component is something that would need to be pulled in by something outside the Felt ecosystem wanting to get the epherma behavior inside Felt via iframe right?

PR #955, opened 2023-11-10
delete the `gro deploy` override for `gro infra/deploy`

ryanatkn:


package.json
```diff
@@ -22,7 +22,7 @@
     "dev": "gro dev",
     "build": "gro build",
     "test": "gro test",
-    "deploy": "gro deploy",
+    "deploy": "gro infra/deploy",
```
ryanatkn, line 25: `npm run deploy` continues to behave the same, maybe that's confusing, I don't know what to think about npm scripts in general, we may want to 1:1 them to Gro tasks as a rule, so this would become `"infra/deploy": "gro infra/deploy",`

greatbacon (approved):

package.json
ryanatkn: ok cool, 2 scripts remain, both for production - `npm start` and `npm run after_deploy`

PR #957, opened 2023-11-26
upgrade deps

ryanatkn:


greatbacon (approved):
I really dig the new UX for mural btw!

